PR #12, opened 2017-12-20
Add command line options to extraction

Ganonmaster:
My experience with `argparse` is very limited. A lot of this code could probably be cleaned up later. Mainly splitting the actual detection/extraction logic from the argument parsing and output.


Clorr:
Thanks for the feature!

PR #17, opened 2017-12-22
Implemented options to change the input directories for the pictures.

Yutsa:
I changed the train.py script to be able to change the source image folder. So now it is not hardcoded you can use `python train.py-a imgA/ -b imgB/` or `python train.py--image-a=imgA/ --image-b=imgB/`

I don't know if you accept this kind of PR but I find it useful.

### DIFF
--- a/train.py
+++ b/train.py
@@ -1,72 +1,86 @@
 import cv2
 import numpy
+import sys
 
-from lib.utils import get_image_paths, load_images, stack_images
-from lib.training_data import get_training_data
+from utils import get_image_paths, load_images, stack_images
+from training_data import get_training_data
 
-from lib.model import autoencoder_A
-from lib.model import autoencoder_B
-from lib.model import encoder, decoder_A, decoder_B
+from model import autoencoder_A
+from model import autoencoder_B
+from model import encoder, decoder_A, decoder_B
+from getopt import getopt
+
 
 try:
-    encoder.load_weights('models/encoder.h5')
-    decoder_A.load_weights('models/decoder_A.h5')
-    decoder_B.load_weights('models/decoder_B.h5')
+    opts, args = getopt(sys.argv[1:], "a:b:", ["image-a=", "image-b="])
+except GetoptError as err:
+    print(err)
+    sys.exit(2)
+for option, value in opts:
+    if option in ("-a", "--image-a"):
+        imageAFolder = value
+    elif option in ("-b", "--image-b"):
+        imageBFolder = value
+    else:
+        assert False, "Unhandled option"
+
+print("Image A path is {}".format(imageAFolder))
+print("Image B path is {}".format(imageBFolder))
+
+try:
+    encoder.load_weights( "models/encoder.h5"   )
+    decoder_A.load_weights( "models/decoder_A.h5" )
+    decoder_B.load_weights( "models/decoder_B.h5" )
 except:
     pass
 
-
 def save_model_weights():
-    encoder.save_weights('models/encoder.h5')
-    decoder_A.save_weights('models/decoder_A.h5')
-    decoder_B.save_weights('models/decoder_B.h5')
-    print('save model weights')
+    encoder.save_weights( "models/encoder.h5"   )
+    decoder_A.save_weights( "models/decoder_A.h5" )
+    decoder_B.save_weights( "models/decoder_B.h5" )
+    print( "save model weights" )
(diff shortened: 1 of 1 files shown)


Ganonmaster:
Hi, thanks for your pull request. I will be reviewing it shortly.

train.py
```diff
@@ -1,72 +1,86 @@
...
+from utils import get_image_paths, load_images, stack_images
+from training_data import get_training_data
 
-from lib.model import autoencoder_A
-from lib.model import autoencoder_B
-from lib.model import encoder, decoder_A, decoder_B
+from model import autoencoder_A
+from model import autoencoder_B
+from model import encoder, decoder_A, decoder_B
+from getopt import getopt
```
Ganonmaster: I would prefer the usage of `argparse` over `getopt` as we don't really need the lower level parsing that the `getopt` library provides. It would also make it consistent with other scripts that I rewrote yesterday. Have a look at `extract.py`, `convert.py` and `lib/cli.py` if you want examples.

train.py
```diff
@@ -1,72 +1,86 @@
 import cv2
 import numpy
+import sys
 
-from lib.utils import get_image_paths, load_images, stack_images
-from lib.training_data import get_training_data
+from utils import get_image_paths, load_images, stack_images
+from training_data import get_training_data
```
Ganonmaster: `utils.py` is in the `lib` package, so I'm not sure if this would even work?

train.py
```diff
@@ -1,72 +1,86 @@
...
+    decoder_B.save_weights( "models/decoder_B.h5" )
+    print( "save model weights" )
 
-    figure = numpy.clip(figure * 255, 0, 255).astype('uint8')
+images_A = get_image_paths( imageAFolder )
+images_B = get_image_paths( imageBFolder )
+images_A = load_images( images_A ) / 255.0
+images_B = load_images( images_B ) / 255.0
 
-    cv2.imwrite('_sample.jpg', figure)
```
Ganonmaster: Removing this line will remove any sort of preview for those who are using the Docker version.

train.py
```diff
@@ -1,72 +1,86 @@
...
+        autoencoder_A.predict( test_B ),
+        ], axis=1 )
+
+    figure = numpy.concatenate( [ figure_A, figure_B ], axis=0 )
+    figure = figure.reshape( (4,7) + figure.shape[1:] )
+    figure = stack_images( figure )
+
+    figure = numpy.clip( figure * 255, 0, 255 ).astype('uint8')
 
+    cv2.imshow( "", figure )
```
Ganonmaster: Breaks the Docker image.

train.py
```diff
@@ -1,72 +1,86 @@
...
+        assert False, "Unhandled option"
+
+print("Image A path is {}".format(imageAFolder))
+print("Image B path is {}".format(imageBFolder))
 
 try:
-    encoder.load_weights('models/encoder.h5')
-    decoder_A.load_weights('models/decoder_A.h5')
-    decoder_B.load_weights('models/decoder_B.h5')
+    encoder.load_weights( "models/encoder.h5"   )
```
Ganonmaster: Codestyle: Try to avoid placing spaces immediately inside parentheses, brackets or braces. https://www.python.org/dev/peps/pep-0008/#pet-peeves

Ganonmaster:
Other than that, this is indeed useful functionality. Once we get the command line functionality working, we have a really good basis that is already more friendly to use and might actually start distributing as a package.

train.py
Yutsa: Actually I merged from another version of this script and it made a lot of unwanted changes like that.

I guess it'd be better to start from scratch if you're already working on the command line arguments.

Yutsa:
Several changes were because I modified a older/modified version of the script.

Since you want to use `argparse`, I could either implement it using the right version of the script or you could reject the PR and do it like you did for the other ones.

Thanks for the review

train.py
Clorr: +1 on that, as argparse is really conveient to also build the help at the same time

Ganonmaster:
I've pushed changes that allow for these arguments to be passed to the script. Please verify their correctness, I was very drunk while refactoring, so I'll probably hate it tomorrow morning.

Yutsa:
I read you commits and it looks great to me. Removing duplicate scripts was a great idea too.

PR #22, opened 2017-12-24
Unified CLI

Yutsa:
I created a single script that calls the other ones so we have a single command.

It's used like that : `./faceswap.py train` or `./faceswap.py extract` or `./faceswap.py convert` with of course the same options as before.

I'm discovering argparse while doing it so feel free to fix my beginner mistakes.


Ganonmaster:
Looks excellent so far. However, you seem to have forgotten the `scripts` package. 😅

Yutsa:
Ah I didn't modify the gitignore, that's why ! That's fixed now.

faceswap.py
```diff
@@ -0,0 +1,13 @@
...
+from scripts.train import TrainingProcessor
+from scripts.convert import ConvertImage
+
+parser = argparse.ArgumentParser()
+subparser = parser.add_subparsers()
+extract = ExtractTrainingData(subparser, "extract", "Extract the faces from a pictures.")
+convert = ConvertImage(subparser, "convert", "Convert a source image to a new one with the face swapped.")
+train = TrainingProcessor(subparser, "train", "This command trains the model for the two faces A and B.")
+arguments = parser.parse_args()
+arguments.func(arguments)
```
lukaville: I think it will be better to wrap everything into:
```
if __name__ == "__main__":
    ...
```

In order to not execute this by accidental importing faceswap as a module

Yutsa: Oh yeah right. It's been quite some time since I've done some Python programming.

faceswap.py
```diff
@@ -0,0 +1,14 @@
...
+import argparse
+from scripts.extract import ExtractTrainingData
+from scripts.train import TrainingProcessor
+from scripts.convert import ConvertImage
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    subparser = parser.add_subparsers()
+    extract = ExtractTrainingData(subparser, "extract", "Extract the faces from a pictures.")
+    convert = ConvertImage(subparser, "convert", "Convert a source image to a new one with the face swapped.")
```
lukaville: Probably it will be more clear to save the order `extract` -> `train` -> `convert` here to avoid confusion (because as far as I understand this order also affects the order in auto-generated help output) :)

faceswap.py
```diff
@@ -0,0 +1,14 @@
...
+    train = TrainingProcessor(subparser, "train", "This command trains the model for the two faces A and B.")
+    arguments = parser.parse_args()
+    arguments.func(arguments)
```
Ganonmaster: Maybe print the default help if this script is called without arguments. Right now it results in a less than helpful argument error when none are given..

lib/cli.py
```diff
@@ -22,10 +22,12 @@ def __call__(self, parser, namespace, values, option_string=None):
 class TrainingProcessor(object):
     arguments = None
 
-    def __init__(self, description='default'):
+    def __init__(self, subparser, command, description='default'):
         print('Initializing')
```
Ganonmaster: All three actions/scripts/objects are initialized when `faceswap.py` is run. Maybe alter or remove these print statements, or only initialize only object that belongs to the subcommand we are actually running.

```
> python .\faceswap.py
Using TensorFlow backend.
Initializing
Initializing
Initializing
```

scripts/train.py
```diff
@@ -0,0 +1,2 @@
+from lib.cli import TrainingProcessor
+
```
Ganonmaster: This file seems kind of useless now.

Yutsa:
The issues you raised should now be fixed.

lukaville:
Please also update the README.md, there are instructions in old style (like `python train.py`)

PR #29, opened 2017-12-26
Allow users to swap the A/B models

Ganonmaster:
Also made the aligner optional/toggleable. I have yet to test these changes.


Yutsa:
It looks great. Were we using the aligner before ?

Ganonmaster:
The aligner was not ready, that's why the default setting is off. Having it as an option should make testing improvements for it easier though.

Yutsa:
What does it align exactly ? How does it improve the results ? Sorry for my ignorance

Ganonmaster:
IF you look at the script in its current form, it simply crops a straight rectangle and applies the transformation to that + pastes it back. However, sometimes a head is tilted sideways and we could get a better crop if we rotated the cropped image for example. That's what the aligner is for. It's kind of rough in its current state, that's why it was commeted out.

Ganonmaster:
I just tested this latest pull request, and it seems to be functioning accordingly.

PR #37, opened 2017-12-28
Improvement of the train action usability.

Yutsa:
* Prints the current iteration number in verbose mode.
* The number of iterations before saving the data can be changed by a command line option.
* Added an option to write result to file even in preview mode.
* Prints time elapsed for each iteration in verbose mode when training.


Ganonmaster:
LGTM 👍

PR #46, opened 2018-01-01
First draft of a plugin approach

Clorr:
WARNING: This is NOT to be merged, it is a draft!

Hi guys, I have a bit more time now, but I'm still not in good conditions to work on the project. I just wanted to submit you a Work In Progress for a more pluggable approach for the project.

As you know, we have a couple variants to the script that is not easy to maintain or try when we have new proposals. A plugin approach would let us have at the same time old legacy code with current official code and with new proposals without having to mess with branches and patches.

The PR here contains just 2 basic plugins, one for `extract`, and one for `convert`. I want to have one for models also, but it is much more complicated for now. They are just simple classes, and adding new plugin will just be adding new classes with the same interface. What's not decided yet is how we switch between "plugin", it can be done dynamically with parameterized file load but it also can be done statically if prefered.

Note that I did some changes that lead, IMHO, to a cleaner code:
- `TrainingProcessor` has been moved to `scripts/train.py` along with `extract.py` and `convert.py`
- I moved `image_augmentation.py` and 2 methods of `utils.py` in `training_data.py` 
- I removed the loop over images in `DirectoryProcessor` as the code is override in `convert.py` and the original code used only in `extract.py`. I think it is better to let a single responsibilty to the `DirectoryProcessor` and not use it to crop faces (At some point the crop faces may be also pluggable)
- I renamed `pixel_shuffler.py` into `PixelShuffler.py` as it is a one class file

Other consideration, but not done in this PR, is:
- The `cli.py` maybe is no more useful in `lib/` maybe the `scripts/` folder should be named `cli/` and the `DirectoryProcessor` could go there

Feel free to integrate part of these changes in the next commits (it will be less modifications in my own commit), and also feel free to give me your feedback on all this.


Clorr:
I updated my code so it can run ;-)

Clorr:
I updated my code so it can run 😃 

Please make me some feedback when you can so I can move on...

Clorr:
Hi guys!

Little awesome news: I added a face align plugin that does not rely on pytorch!

I still need to do the switch between plugins, but otherwise, I think I'm on a good way!

Don't hesitate to give me feedback!

Thx

Yutsa:
I love the plugin idea since there are lots of versions of the scripts on the deepfake subreddit, it'd be a good way to use several versions.

Yutsa:
On reddit a user also patched the falce_align and merge_face scripts to use masks and it worked great apparently, should we be able to include those two scripts now that you deleted the pytorch dependency ?

It's a bit sad that all those people on reddit use a collection of scripts scattered everywhere and are not using and contributing directly to this project

Clorr:
Yes, my goal is to gather/integrate the different improvments, and that's why a plugin approach comes handy.

I was trying since a while to understand/integrate the face alignment. Now that it is done on `extract`, I'll move on the face merge.

Clorr:
But I don't like to spend too much time on debugging, so if you can try and give me feedback, feel free

Ganonmaster:
I'll have some time tonight to review and provide feedback.

Yutsa:
I ran the extract with align and it worked flawlessly ! Great job !

Should we add a way for the user to choose which extractor plugin to use ?

Clorr:
Thanks for the kind feedback! Did you also try on side faces? I tried on a group photo, but not much more...

I'm working currently on a factory to return a plugin according to a string. I'll let you add the argparser part.

I have a working branch (https://github.com/Clorr/faceswap/tree/dev/masked_plugin) where I'm adding the masked convert. As you may see, the convert plugin will have to handle specific params depending on which plugin is active. This should be passed as command line args, but I don't know how to tackle that. But that's for later I think...

Yutsa:
I tried on a face that was leaning on the side and it correctly put it straight.

For instance :

Clorr:
Ah, I was thinking about some face you see from the side, profile view ;-)

Yutsa:
Oh sorry, english is not my native language and I misunderstood. I've got that one from the side but I don't have the original one anymore.

Clorr:
No problem. It seems fine ;-)

Yutsa:
Weirdly enough, I tried to train the model with these new pictures added to my dataset and the training script exits before even starting the training.

I tried with the old pictures, without the alignement, and it works. I don't know what causes this for my pictures, I'm looking into it.

It may have to do with the size being 352x352 instead of 256x256

Yutsa:
I resized and indeed it was the issue. You wrote :

```Python
new_size = int( size + padding * 2 )
```
 I don't really get why but the extra 96 pixels in width and height are because of that since you pass a padding of 48 here 

```Python
return self.transform( image, alignment, size, 48 ) # original size was 160. Crop was on 256, is that ok?
```

Clorr:
Nice catch, original script was using 160 instead of `size`. I replaced without really understanding how things work....

Clorr:
Also a good catch is the silent failure. If you can add some warning there, that would be handy to some people...

Clorr:
I pushed a working version of the enhanced face merge (in https://github.com/Clorr/faceswap/tree/dev/masked_plugin). It should be working now, but I did not test it thoroughly. I'll do that tomorrow, and update the master. Also I will add the PluginSelector

Yutsa:
Yeah I'll try, I think it's from the `load_images(image_paths, convert=None)` function, I'll look into it

Clorr:
Doing a new PR with my final version

Issue #44, opened 2017-12-30
Allow a user to gracefully cancel training when there is no preview


Ganonmaster:
In its current state, the tool does not allow users to hit `q` to quit the tool unless the preview window is active and has focus. The tool will only save after a certain amount of iterations, so it would be great if we could enable "non-previewers" or Docker users to quit out this way also and preserve their work..

Clorr:
Done in PR #59

PR #47, opened 2018-01-01
Print help information only if arguments parsing error occurred

### LINKED ISSUE #43: Do not print help text after cancelling
At this time, all the scripts will print help text after exiting in an unexpected fashion. For example, when cancelling a command, it will still print the help text before quitting.

lukaville:
Fixes #43

### DIFF
--- a/lib/cli.py
+++ b/lib/cli.py
@@ -161,7 +161,7 @@ class TrainingProcessor(object):
 
         for epoch in range(1, 1000000):
             if self.arguments.verbose:
-                print("Iteration number {}".format(1))
+                print("Iteration number {}".format(epoch))
                 start_time = time.time()
             warped_A, target_A = get_training_data(images_A, BATCH_SIZE)
             warped_B, target_B = get_training_data(images_B, BATCH_SIZE)
--- a/lib/cli.py
+++ b/lib/cli.py
@@ -2,6 +2,7 @@ import argparse
 import os
 import cv2
 import numpy
+import sys
 import time
 
 from lib.utils import get_image_paths, get_folder, load_images, stack_images
@@ -41,7 +42,14 @@ class TrainingProcessor(object):
             print('Not loading existing training data.')
             print(e)
 
-        self.process()
+        try:
+            self.process()
+        except KeyboardInterrupt:
+            try:
+                self.save_model_weights()
+            except KeyboardInterrupt:
+                print('Saving model weights is cancelled')
+            sys.exit(0)
 
     def parse_arguments(self, description, subparser, command):
         parser = subparser.add_parser(
--- a/lib/cli.py
+++ b/lib/cli.py
@@ -27,6 +27,17 @@ class TrainingProcessor(object):
     def __init__(self, subparser, command, description='default'):
         self.parse_arguments(description, subparser, command)
 
+        self.models = [{
+            'file': 'encoder.h5',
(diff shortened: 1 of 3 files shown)


lib/cli.py
```diff
@@ -26,22 +27,40 @@ class TrainingProcessor(object):
...
-        except Exception as e:
-            print('Not loading existing training data.')
-            print(e)
+        for model in self.models:
+            file_path = os.path.join(self.arguments.model_dir, model['file'])
+            if os.path.isfile(file_path):
+                model['container'].load_weights(file_path)
+            else:
+                print("Model file {} is not exists, creating...".format(model['file']))
+                model['container'].save_weights(file_path)
```
lukaville: Create model files before training. So script will fail almost immediately if for example there are problems with writing permissions #failfast

lib/cli.py
```diff
@@ -26,22 +27,40 @@ class TrainingProcessor(object):
...
 
-        self.process()
+        try:
+            self.process()
+        except KeyboardInterrupt:
+            try:
+                self.save_model_weights()
```
lukaville: Saves weights after Ctrl+C (possibly fixes #44)

lib/utils.py
```diff
@@ -44,3 +47,13 @@ def stack_images(images):
...
         axes=numpy.concatenate(new_axes)
         ).reshape(new_shape)
+
+
+class FullHelpArgumentParser(argparse.ArgumentParser):
+    """Identical to the built-in argument parser, but on error
+    it prints full help message instead of just usage information
+    """
+    def error(self, message):
+        self.print_help(sys.stderr)
```
lukaville: kinda hacky solution, maybe we don't need to print full help in case of argument parsing error?

lib/cli.py
Yutsa: This is a nice feature !

lib/cli.py
Yutsa: This is cleaner indeed, it looks good to me

lib/cli.py
```diff
@@ -136,13 +155,13 @@ def process(self):
...
-        print('press "q" to stop training and save model')
+        print('Press "q" to stop training and save model')
 
         BATCH_SIZE = 64
 
-        for epoch in range(1000000):
+        for epoch in range(1, 1000000):
             if self.arguments.verbose:
-                print("Iteration number {}".format(epoch + 1))
+                print("Iteration number {}".format(1))
```
Yutsa: I don't know why you wrote `print("Iteration number {}".format(1))` ? It will be a hardcoded 1 instead of the actual epoch, which is not helpful. Or am I mistaken ?

lukaville: Oh, yeah. It's a mistake. I wanted to write just "epoch" because now it starts from 1.

Yutsa: Oh right, it's cleaner indeed then.

Clorr:
Hi @lukaville ,

I'm closing the PR as I did reintegrate the modifications in my plugin branch.

You will see the modifications here : https://github.com/Clorr/faceswap/commit/0eb85275f15511fdb9ca9e9cffb823ba2431c621

Feel free to review, test and provide feedback.

Note that it is still a work in progress branch so maybe you may encounter problems

Also Note that the default model there is the GAN network, if you want to change this, I can help you

PR #50, opened 2018-01-03
Adding plugins + Integration of Face Alignment & Masked script

Clorr:
Hi guys,

Here is a new PR for the plugin approach. It is still rough, but should already be as good as the previous version.

It features:
- A plugin folder where we can put different versions of algorithm
- New plugins with the `align` extraction and the `masked` merge (which are the latest AFAIK)

There is still room for improvments:
- General syntax may be improved, but please provide me specific feedback if you see something (or even do a PR on my own repo if you prefer)
- We can add easily the plugin choice through arguments, but its out of my scope for now
- We should add arguments parsing for plugins, but this part is not clear for me
- I want to add a specific support for `Model` plugins as well as a cleaner model loading, but I'm not sure of the consequences yet, so I let that for later


Apollo122:
Great! When it's merged a usage guide would be nice so we can try out

Clorr:
Usage did not change. The command line plugin selection is still to be done. The people who will add arg parsing will certainly do the doc updates, but its not done yet...

Yutsa:
I tested it and it works great ! For me the only missing thing will be the ability to choose which plugin to use and pass arguments to that plugin as you said.

I could look into it, unfortunately I have exams during the whole next week so I won't be available to code before somewhere around the 15th of January.

lib/training_data.py
```diff
@@ -27,3 +27,61 @@ def get_training_data(images, batch_size):
...
+    result = cv2.warpAffine(
+        image, mat, (w, h), borderMode=cv2.BORDER_REPLICATE)
+    if numpy.random.random() < random_flip:
+        result = result[:, ::-1]
+    return result
+
+# get pair of random warped images from aligened face image
+
+def random_warp(image):
+    assert image.shape == (256, 256, 3)
```
Ganonmaster: Probably want to make this assertion more explicit, so the user knows what's wrong with his image. It seems to fail silently due to us catching all exceptions at the beginning of `faceswap.py` and any attempt to parse it comes up without error message.

plugins/Convert_Masked.py
```diff
@@ -0,0 +1,84 @@
...
+        self.autoencoder = autoencoder
+
+        # if args.erosionKernelSize > 0:
+        #     self.erosion_kernel = cv2.getStructuringElement(cv2.MORPH_ELLIPSE,(args.erosionKernelSize,args.erosionKernelSize))
+        # else:
+        #     self.erosion_kernel = None
+
+        self.erosion_kernel = None
+        self.blurSize = 2
+        self.seamlessClone = False
```
Ganonmaster: I would opt for a default of `True`

Ganonmaster:
Other than that, seems legit. I was unable to test this on GPU at this time, but I doubt it would yield different results. Just faster.

Clorr:
Ok, many thanks! I merge this because we need to move on.

The 2 points are from the original code, so we can say it's legacy ;-) I can change it at some point anyway...

PR #51, opened 2018-01-03
Proposal for pluggable Models (NOT TO BE MERGED (YET))

Clorr:
Hi guys,

Here is a next step in the plugin approach. I separated models as plugins. The example I provide is the original one, and the one for low end cards.

Feel free to have a look and give me feedback. If it is ok, it will be pushed with other plugins...


Clorr:
I think I have to rethink some things...

Issue #39, opened 2017-12-28
Don't reload models everytime `convert_one_image` is called


Ganonmaster:
## Expected behavior

Use the convert command to convert a directory. `convert_one_image` loads the model once.

## Actual behavior

Use the convert command to convert a directory. `convert_one_image` loads the model every time that it is called.

Clorr:
I'm adding a Convert class which will load models on __init__

It will be in my branch until I do a PR, but feel free to have a look

Clorr:
The class is in my PR. However, I did not solve the fact that the model is reloaded for each image.

@Ganonmaster , let me know if you want to tackle this separately, or if you are ok with my plugins folder. If so I'll make it run so it can be integrated quickly

Clorr:
This is handled in PR #52 in a dedicated commit

PR #52, opened 2018-01-04
[PREVIEW] Adding models as plugins + GAN + #39 + #43 + # 44(?) + #53

Clorr:
Hi, 

I submit a new preview for my "models as plugins" branch.

It's been reworked so we can more easily provide different code bases. This one integrates the (awesome) GAN version of @shaoanlu. 

It is still very rough, so do not forget it is a beta more than a ready-to-use version

Any feedback is appreciated of course!


Apollo122:
Man so excited about GAN. I saw the gifs and I was blown away with masking it uses which doesn't blur foreign objects in front of faces such as hands.
I have one question though, does it help with resolution of generated face? 64x64 is really low and because of that close up scenes suck with original code.
I can test it if anyone give me a walk-through about how to use it.

Clorr:
Yes, it is indeed very interesting. the code though is a bit hard to split and structure so I'll move on slowly on this. I did a new PR that is more dedicated to GAN

The usage is exactly the same as the previous one.

On the face resoltuiion, it doesn't help, but somehow I think we can train the encoder with low res images and the decoder with higher res one. I should give it a try but I won't have much time soon

Apollo122:
I hope we can get high resolution generated images. So we can have better close-ups. With GAN, objects blocking faces are no longer being an issue doing more close-up scenes is very tempting.

Issue #10, opened 2017-12-20
Add video support


Clorr:
I'll try to tackle that one with the simple [cv2 support](http://opencv-python-tutroals.readthedocs.io/en/latest/py_tutorials/py_gui/py_video_display/py_video_display.html)

Ganonmaster:
Does OpenCV make it possible to account to interpolate face positions between frames? I've had several instances where I've tried to convert a series of frames, but it fails to detect faces in a few frames in between; causing[ a glitching effect where the target face turns on and off](https://gfycat.com/GrandVengefulGalapagossealion).

Clorr:
Indeed this a common issue. Face detection is often blocked by compression artifacts. It is the case in still pics and is even more present in videos.

I don't think there is a built in interpolation, but you can try doing it manually. While parsing frames, just keep a track of 1 or 2 previous frames and if current frame has nothing detected, do a 'mean' with previous data.

Other possibility is to keep the result of the last detection, and keep frames without faces until you detect another face. If the new face is close to the last, apply the new detection (or the old, or an interpolation) to the buffered frames. (Don't forget to give a max size to your buffer and flush it if no new face has been detected for x frames)

I'll try to tackle that later on...

Clorr:
For now, I have an issue with input video reading. It seems python-opencv has no video decoder so I had to compile the full version...

Clorr:
As I see on your video, there are also little glitches due to the face generator that may generate different faces for 2 consecutive frames. I don't see a clear solution here for now.

Clorr:
I'm giving up for now, opencv-python seems not to have video support, and the full opencv lib is a pain to build....

I'll try to handle sequences of photos instead (video will just be a sequence of frames so it will adapt easily to videos)

Clorr:
Remark: ‘python-opencv‘ seems not to have video support built-in as default. Is ‘moviepy‘ a better alternative?

Ganonmaster:
In the end, all these libraries boil down to is basic interaction with ffmpeg. I'm pretty sure python-OpenCV has this, but only if configured to use it. I would opt for the library with the most straightforward/pythonic interface. I would guess that is most likely moviepy, but I will look further into this after I've recovered from the new years hangover.

IcyTv:
How about taking existing code from somewhere that extracts all frames from a video, running the face swap over those and then taking the output and combine it to a video?

fetchlister:
[...]

PR #54, opened 2018-01-05
[DRAFT] Video conversion #10

Clorr:
This is a draft of a direct video conversion command.

It is here for test/debug purpose.

Note that the conversion is done by default with `Convert_GAN` which is also a rough draft

You can try this with other plugins by setting manually the plugin in the `scripts/convert_video.py`

Have fun!

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,2 +1,3 @@
 *
 !requirements*
+!install_ffmpeg.py
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,4 +26,7 @@ RUN apt-get update -qq \
 COPY ./requirements-python35.txt .
 RUN pip3 --no-cache-dir install -r ./requirements-python35.txt
 
+COPY ./install_ffmpeg.py
+RUN python3 ./install_ffmpeg.py
+
 WORKDIR /srv/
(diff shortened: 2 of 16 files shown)


torzdf:
Quick question, what's the intended workflow here? From a quick scan it looks like it assumes that you have already trained your model, and you are pointing a video at it and it is performing the swap on the frames within the video, to output a swapped video. Would this be correct?

Clorr:
It is part of the GAN port to this repo. You will see that there is a plugin GANBoundingBox. The improvment here is that the convert uses a context between frames to lower flicker. See GAN repo if you want more details (although I am not sure how much details there is there)

torzdf:
Ok, I see. So this is intended to be GAN specific? Would the context between frames also be applicable if frames have already been extracted.

The reason I ask is that I envisage a workflow similar to following (and I accept that I may be way off the mark here, just tossing some ideas around ;) )

Convert: 
- Input dir is provided (as currently) 
- if input is a movie file (e.g. mp4, mkv etc) move on, otherwise continue as normal
- use ffmpy to read the movie file to gather fps, number of frames, resolution
- pipe output from ffmpy, converted to pngs (resized down if necessary) through the standard convert process and pipe back in to an output movie file with the audio from the original video into your output.

This would mean that video would still follow the same process as currently exists, the only change would be reading the image from a pipe and passing it back to a pipe rather than from the disk. It would also mean that any bounding box improvements would benefit the standard process (it indeed it works that way).
- pipe input

Issue #49, opened 2018-01-02
Integrate enhanced train.py


Clorr:
A guy posted here a better [train.py](https://pastebin.com/UuGgjd3N) with thread support. If it is interesting, I can add it to the main repo

Source: https://www.reddit.com/r/deepfakes/comments/7nlql5/optimized_trainpy/

Clorr:
Added in my master, and in PR #59

PR #62, opened 2018-01-11
Implemented multicore directory processor; increased speed of extracting faces

iSevenDays:
I hope you are tired as I - waiting half an hour to extract faces in 1 thread.

### DIFF
--- a/lib/cli.py
+++ b/lib/cli.py
@@ -132,6 +132,9 @@ class MultiProcessDirectoryProcessor(DirectoryProcessor):
                             help="Jobs number. Should be between 1 and number of cpu cores.")
 
     def process_directory(self):
+        if self.maximum_jobs_count == 1:
+            super().process_directory()
+            return
         jobs = []
         for filename in self.input_dir:
             if self.arguments.verbose:
@@ -139,7 +142,7 @@ class MultiProcessDirectoryProcessor(DirectoryProcessor):
             p = multiprocessing.Process(target=self.process_image, args=(filename,))
             jobs.append(p)
             p.start()
-            if len(jobs) > self.maximum_jobs_count:
+            if len(jobs) >= self.maximum_jobs_count:
                 for job in jobs:
                     job.join()
                 jobs = []
--- a/lib/cli.py
+++ b/lib/cli.py
@@ -147,5 +147,7 @@ class MultiProcessDirectoryProcessor(DirectoryProcessor):
                     job.join()
                 jobs = []
             self.images_processed = self.images_processed + 1
-
+        # Process last jobs left in queue
+        for job in jobs:
+            job.join()
         self.finalize()
--- a/lib/cli.py
+++ b/lib/cli.py
@@ -116,6 +116,21 @@ class DirectoryProcessor(object):
 
 class MultiProcessDirectoryProcessor(DirectoryProcessor):
 
+    maximum_jobs_count = 1
+
+    def process_arguments(self, arguments):
+        self.maximum_jobs_count = arguments.jobs
+        print("Parallel jobs: {}".format(self.maximum_jobs_count))
+        super().process_arguments(arguments)
+
+    def parse_arguments(self, description, subparser, command):
+        super().parse_arguments(description, subparser, command)
+        self.parser.add_argument('-j', '--jobs',
+                            dest="jobs",
+                            type=int,
+                            default=1,
+                            help="Jobs number. Should be between 1 and number of cpu cores.")
+
     def process_directory(self):
         jobs = []
         for filename in self.input_dir:
@@ -124,7 +139,7 @@ class MultiProcessDirectoryProcessor(DirectoryProcessor):
             p = multiprocessing.Process(target=self.process_image, args=(filename,))
             jobs.append(p)
             p.start()
-            if len(jobs) > multiprocessing.cpu_count():
+            if len(jobs) > self.maximum_jobs_count:
                 for job in jobs:
                     job.join()
                 jobs = []
--- a/lib/cli.py
+++ b/lib/cli.py
@@ -1,8 +1,10 @@
 import argparse
+import multiprocessing
 import os
 import time
 
 from lib.utils import get_image_paths, get_folder, load_images
+
 
 class FullPaths(argparse.Action):
     """Expand user- and relative-paths"""
@@ -10,6 +12,7 @@ class FullPaths(argparse.Action):
     def __call__(self, parser, namespace, values, option_string=None):
         setattr(namespace, self.dest, os.path.abspath(
             os.path.expanduser(values)))
+
 
 class DirectoryProcessor(object):
     '''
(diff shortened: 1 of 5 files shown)


Clorr:
Nice PR, I think it will be accepted.

Would it be possible to have it also on convert (If that brings any improvment)?

iSevenDays:
@Clorr  I will try this also on convert, but I think it will not be an option - convert usually requires a lot of video memory, I saw it used around ~3-4 GB from my 6 GB.

Clorr:
Yes, maybe parallelizing the work of convert is not a good idea. Or maybe I could handle it as for the training, where the photo are read separately from the main training loop (see #49)

iSevenDays:
@Clorr convert will not work properly in multi-threaded mode. https://github.com/tensorflow/tensorflow/issues/8220
Firstly, I got an error that a second process doesn't have enough memory.
Then, when I set to "allow growth" of video memory in TensorFlow - I got errors CUDA NOT INITIALIZED.

I added command line argument --job -j , you can specify jobs number.

Clorr:
Hi,

Thanks for the update. If convert may have problem with the multiprocessing, I think it is better not to propose the option.

On my side, I was not proposing to make the convert multithreaded, I was proposing to do the photo preprocessing(face recognition + landmarks detection) in a separate thread, like I did for training. It shouldn't have impact on tensorflow itself

iSevenDays:
@Clorr  I have just tried to merge conflicts and it seems that your PR doesn't work for me.

Class DirectoryProcessor has a method 
def process(self, reader):
..

But above, in process_arguments, you call self.process() without arguments.

Given the above, class ExtractTrainingData(DirectoryProcessor) has a method without arguments at all
def process(self):

And this is not a correct override as I think. I didn't get what does 'reader' stands for.

I couldn't make a fast merge, so it seems that you know more details than me on how to merge it.

Dagur:
This PR doesn't work for me. I get this error

E:\opt\faceswap>python faceswap.py extract -i E:\vids\equi\orig -o
E:\proj\first\data\equi -j 2
Using TensorFlow backend.
Parallel jobs: 2
Input Directory: E:\vids\equi\orig
Output Directory: E:\proj\first\data\equi
Starting, this may take a while...
  File "faceswap.py", line 18, in <module>
    arguments.func(arguments)
  File "E:\opt\faceswap\lib\cli.py", line 124, in process_arguments
    super().process_arguments(arguments)
  File "E:\opt\faceswap\lib\cli.py", line 52, in process_arguments
    self.process_directory()
  File "E:\opt\faceswap\lib\cli.py", line 144, in process_directory
    p.start()
  File
"C:\Users\dagur\AppData\Local\Programs\Python\Python36\lib\multiprocessing\process.py",
line 105, in start
    self._popen = self._Popen(self)
  File
"C:\Users\dagur\AppData\Local\Programs\Python\Python36\lib\multiprocessing\context.py",
line 223, in _Popen
    return _default_context.get_context().Process._Popen(process_obj)
  File
"C:\Users\dagur\AppData\Local\Programs\Python\Python36\lib\multiprocessing\context.py",
line 322, in _Popen
    return Popen(process_obj)
  File
"C:\Users\dagur\AppData\Local\Programs\Python\Python36\lib\multiprocessing\popen_spawn_win32.py",
line 65, in __init__
    reduction.dump(process_obj, to_child)
  File
"C:\Users\dagur\AppData\Local\Programs\Python\Python36\lib\multiprocessing\reduction.py",
line 60, in dump
    ForkingPickler(file, protocol).dump(obj)
AttributeError: Can't pickle local object
'ArgumentParser.__init__.<locals>.identity'

E:\opt\faceswap>Using TensorFlow backend.
  File "<string>", line 1, in <module>
  File
"C:\Users\dagur\AppData\Local\Programs\Python\Python36\lib\multiprocessing\spawn.py",
line 105, in spawn_main
    exitcode = _main(fd)
  File
"C:\Users\dagur\AppData\Local\Programs\Python\Python36\lib\multiprocessing\spawn.py",
line 115, in _main
    self = reduction.pickle.load(from_parent)
EOFError: Ran out of input

E:\opt\faceswap>

iSevenDays:
@Dagur, unfortunately, I don't have a Windows set up to test, I use macOS. It seems that multiprocessing module works incorrectly in Windows.
As an option, you can always use -j 1, and this will fall back to default implementation that will work.

Apollo122:
I'm not sure about merging a PR which is not available for Windows. Maybe we can branch it as macOS version but that would create maintenance issues for future.

Clorr:
@iSevenDays I pushed an update. The `def process(self, reader)` was a miss from an older version of my code. I was passing `process_directory` as argument to the function, but now I call it directly from the `process` functions.

I will have a look to integrate your changes on my branch.

Do you have an idea of the windows related problem?

Dagur:
I found this about windows and multiprocess
https://stackoverflow.com/questions/9670926/multiprocessing-on-windows-breaks#9671030

Sounds like all that needs to be done is to move process_image so that it's a top level function. I'll give it a go

Dagur:
I can confirm that this fixes the problem. I added this top the bottom of cli.py 

```
from plugins.PluginLoader import PluginLoader
import cv2
from lib.faces_detect import detect_faces
from pathlib import Path

def process_image(filename, output_dir):
    extractor = PluginLoader.get_extractor("Align")()
    faces_detected = 0
    try:
        image = cv2.imread(filename)
        for (idx, face) in enumerate(detect_faces(image)):
            # if idx > 0 and self.arguments.verbose:
            #     print('- Found more than one face!')
            #     self.verify_output = True

            resized_image = extractor.extract(image, face, 256)
            output_file = output_dir / Path(filename).stem
            cv2.imwrite(str(output_file) + str(idx) +
                        Path(filename).suffix, resized_image)
            faces_detected = faces_detected + 1
    except Exception as e:
        print('Failed to extract from image: {}. Reason: {}'.format(filename, e))
    finally:
        return faces_detected
```

and called it like this
`p = multiprocessing.Process(target=process_image, args=(filename, self.output_dir))`

I'll let you refactor this. I'm not used to object oriented python :)

iSevenDays:
@Dagur 

I think the reason why does it work is that args now contain self.output_dir, could you please check it?

Dagur:
@iSevenDays I'm not sure I understand the question. According to the stackoverflow thread, the problem is that self.process_images is a bound function and can't be pickled. I don't think that adding output_dir as a parameter fixes anything

Clorr:
@iSevenDays I added in my branch the same preload for the convert as the one I had on train. Feel free to review it

Clorr:
Hi @iSevenDays !

Sorry I could not work on the project these last weeks, but I'm now back with a bit of time. I'll probably take your changes now that I finished my own modifications. If you want to review your changes, feel free, otherwise, I can do it on my side.

iSevenDays:
@Clorr  thanks, I wanted to make a contribution, but, currently, I just don't have time. Feel free to use my code

Clorr:
I did not follow your code, but I did a `lib.multithread.py` to have the code separated and easily rewritable. We still have problems on windows, but I hope someone can sort this out ;-)

But anyway thanks for the contrib, I would have been pleased to see your name in the commits ;-)

Issue #85, opened 2018-01-31
Suggestion: Change default values of variables in Convert_Masked


Irastris:
```
self.erosion_kernel = None
self.blurSize = 8 # Default: 2
self.seamlessClone = True # Default: False
self.maskType = 'FaceHullAndRect' # Default: Rect
```
In my testing, this produces far better results than the default values in almost every case, and would help prevent confusion for less technically inclined users were they made the default. Thoughts?

Irastris:
Or atleast add optional command line parameters to tweak these

Clorr:
Thanks for your feedback! I'd be happy if we had some command line params for that but, unfortunately, no one seems to be available for that now. If you have skills you can do a pull request, it will certainly be accepted

Irastris:
I'll give it a shot :)

ruah1984:
@Irastris @Clorr ,i think seamlessclone and masktype but not sure the erosion kernel can be adjust or not.
default will be "none" , if want to change, what will be the command?

i suggest we can put at USAGE.md , the command to change. refer to this 

https://www.reddit.com/r/deepfakes/comments/7mbzu2/patched_face_alignment_and_merge_scripts_with/

[–]depfakacc[S] 2 points 1 month ago 
These need to be used together as the new align_images_masked.py also saves the face points detected to the json file along with the filenames and transforms.

A few interresting options to pass merge_faces_masked.py:

--seamlessClone=yes will attempt to use opencv seamless clone on the faces, otherwise alpha blending is used.

--blurSize=2 will blur the borders of the mask to give a smooth transition between faces and head.

--erosionKernelSize=2 will nibble at the borders of the face mask, making the face area used smaller before the merge is performed.

--maskType=FaceHullAndRect takes three possible options:

Rect - The standard transformed square edges. FaceHull - Create an outline from the detected face points and mask with these. FaceHullAndRect - a AND of the two keeping portions of the face that are both within the detected face points and the face rect.

align_images_masked.py also has new options for --startFrame and --maxFrames which control the image number to start conversion at, and the maximum number of images to convert.

Irastris:
Welp, I actually have no idea how, my Python knowledge is very lacking. I can't figure out how to pass the args from the `ConvertImage` class in `convert.py` to the `Convert` class in `Convert_Masked.py`.

Clorr:
If you feel more comfortable with it, try to do it in a config file

gdunstone:
[...]

PR #89, opened 2018-02-01
extended arguments for convert.py

gdunstone:
re https://github.com/deepfakes/faceswap/issues/85

Removed some print statements because they mess with tqdm.


gdunstone:
I changed the default mask type to facehullandrect.
I also made Convert_Adjust and Convert_Masked able to take extra parameters (and ignore ones that do not apply).

PR #90, opened 2018-02-01
frame-range option for convert

gdunstone:
this option allows you to define multiple frame ranges to process.

to convert the frames 0 to 50 and 100 to end and still write out unconverted frames:

`./faceswap.py convert -s -fr min-50 100-max`

to convert frames 0 to 50 and 100 to 200 and discard any unconverted frames:

`./faceswap.py convert -s -fr 0-50 100-200 --discard-frames`

### DIFF
--- a/plugins/Convert_Adjust.py
+++ b/plugins/Convert_Adjust.py
@@ -6,11 +6,11 @@ import numpy
 import os
 
 class Convert(object):
-    def __init__(self, encoder):
+    def __init__(self, encoder, smooth_mask=True, avg_color_adjust=True, **kwargs):
         self.encoder = encoder
-         
-        self.use_smooth_mask=True
-        self.use_avg_color_adjust=True
+        
+        self.use_smooth_mask = smooth_mask
+        self.use_avg_color_adjust = avg_color_adjust
 
     def patch_image( self, original, face_detected ):
         #assert image.shape == (256, 256, 3)
--- a/plugins/Convert_Masked.py
+++ b/plugins/Convert_Masked.py
@@ -6,18 +6,16 @@ import numpy
 from lib.aligner import get_align_mat
 
 class Convert():
-    def __init__(self, encoder):
+    def __init__(self, encoder, blur_size=2, seamless_clone=False, mask_type="facehullandrect", erosion_kernel_size=None, **kwargs):
         self.encoder = encoder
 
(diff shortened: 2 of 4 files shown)


Clorr:
Hi, I just read your code as I'm continuing on the GAN plugin. I just noticed something that would be a better practice in order to keep the code clean.

In the code below:
```

    def convert(self, converter, item):
        try:
            (filename, image, faces) = item
            skip = False
            try:
                if self.frame_ranges is not None:
                    # grab the index with last number regex
                    idx = int(self.imageidxre.findall(filename)[0])
                    # only skip if the current index is not between any of the frame ranges.
                    skip = not any(map(lambda b: b[0]<=idx<=b[1], self.frame_ranges))
            except:
                # if we error, dont skip
                skip = False

            if not skip: # process as normal
                for idx, face in faces:
                    image = converter.patch_image(image, face)

            output_file = self.output_dir / Path(filename).name
            if self.arguments.discard_frames and skip:
                return
            cv2.imwrite(str(output_file), image)
        except Exception as e:
            print('Failed to convert image: {}. Reason: {}'.format(filename, e))
```

it would be better to avoid having the logic inside the main function, because if add more and more logic here, it will end up with a big messy function. It would be better to have something like:

```
    def convert(self, converter, item):
        try:
            (filename, image, faces) = item
            if not checkSkip(filename): # process as normal
                for idx, face in faces:
                    image = converter.patch_image(image, face)

            output_file = self.output_dir / Path(filename).name
            if self.arguments.discard_frames and skip:
                return
            cv2.imwrite(str(output_file), image)
        except Exception as e:
            print('Failed to convert image: {}. Reason: {}'.format(filename, e))
```
and define `checkSkip` in an outside function ;-)

gdunstone:
You are totally right about checkSkip needing to be a separate func.

gdunstone:
I think this is ok?

I feel like checking a variable is less taxing than calling a func here.

Clorr:
Yes, it's ok like that ;-)

Adding an intermediate var has no impact on the speed, if you do `if not func()`, there will always be an internal var created. Also calling a function has a cost, but it is very small (typically around 0.1 microsecond).

The only cases where you would prefer inlining the code (ie. not creating a function) is in places where the code is called millions of times or more.

For a loop like this, that is called once per image, the performance impact is totally unnoticeable and it is much better to keep code clean

PR #92, opened 2018-02-01
frame-range changes

gdunstone:
feedback from Clorr

https://github.com/deepfakes/faceswap/pull/90#issuecomment-362309166


gdunstone:
I think this is allright.

gdunstone:
@Clorr sorry if I rely on you too much. I normally work on private repos.

Clorr:
No problem, I'm not that used to public repos either. And I prefer a living repo than a too strict one where nothing moves on ;-)

gdunstone:
lol yep. Also I'm super keen for GAN.
Haven't looked much at it yet but I'm excited. XD

PR #99, opened 2018-02-02
python version check

gdunstone:
will now raise exception if using python version below 3.2


gdunstone:
why is git including all my previous commmits wth...

PR #102, opened 2018-02-02
[Preview] GAN plugin

Clorr:
Hi guys!

Here a new version for the GAN plugin. The model has not changed since last time, but now the convert is working (without masking however).

It is still based on the model done [here](https://github.com/shaoanlu/faceswap-GAN/blob/master/temp/faceswap_GAN_keras.ipynb), so it is not the absolute latest, but now that I can move on, I'll try to be closer from the latest versions of @shaoanlu

### DIFF
--- a/plugins/Convert_GAN.py
+++ b/plugins/Convert_GAN.py
@@ -1,0 +1,18 @@
+# Based on the https://github.com/shaoanlu/faceswap-GAN repo (master/FaceSwap_GAN_v2_train.ipynb)
+
+import cv2
+import numpy
+
+class Convert(object):
+    def __init__(self, encoder, **kwargs):
+        self.encoder = encoder
+
+    def patch_image( self, original, face_detected ):
+        face = cv2.resize(face_detected.image, (64, 64))
+        face = numpy.expand_dims(face, 0) / 255.0 * 2 - 1
+        mask, new_face = self.encoder(face)
+        new_face = mask * new_face + (1 - mask) * face
+        new_face = numpy.clip((new_face[0] + 1) * 255 / 2, 0, 255).astype('uint8')
+
+        original[face_detected.y: face_detected.y + face_detected.h, face_detected.x: face_detected.x + face_detected.w] = cv2.resize(new_face, (face_detected.w, face_detected.h))
+        return original
--- a/plugins/Model_GAN/Model.py
+++ b/plugins/Model_GAN/Model.py
@@ -1,0 +1,179 @@
+# Based on the https://github.com/shaoanlu/faceswap-GAN repo (master/temp/faceswap_GAN_keras.ipynb)
+
+from keras.models import Model
+from keras.layers import *
+from keras.layers.advanced_activations import LeakyReLU
+from keras.activations import relu
+from keras.initializers import RandomNormal
+from keras.applications import *
+from keras.optimizers import Adam
+
+from lib.PixelShuffler import PixelShuffler
+
+netGAH5 = '/netGA_GAN.h5'
+netGBH5 = '/netGB_GAN.h5'
+netDAH5 = '/netDA_GAN.h5'
+netDBH5 = '/netDB_GAN.h5'
+
+class GANModel():
+    img_size = 64 
+    channels = 3
+    img_shape = (img_size, img_size, channels)
+    encoded_dim = 1024
+    
+    def __init__(self, model_dir):
+        self.model_dir = model_dir
+
+        optimizer = Adam(1e-4, 0.5)
+
+        # Build and compile the discriminator
+        self.netDA, self.netDB = self.build_discriminator()
+
+        # For the adversarial_autoencoder model we will only train the generator
+        self.netDA.trainable = False
+        self.netDB.trainable = False
+        
+        self.netDA.compile(loss='mse', optimizer=optimizer, metrics=['accuracy'])
+        self.netDB.compile(loss='mse', optimizer=optimizer, metrics=['accuracy'])
+
+        # Build and compile the generator
+        self.netGA, self.netGB = self.build_generator()
+        self.netGA.compile(loss=['mae', 'mse'], optimizer=optimizer)
+        self.netGB.compile(loss=['mae', 'mse'], optimizer=optimizer)
+
+        img = Input(shape=self.img_shape)
+        alphaA, reconstructed_imgA = self.netGA(img)
+        alphaB, reconstructed_imgB = self.netGB(img)
+
+        def one_minus(x): return 1 - x
+        # masked_img = alpha * reconstructed_img + (1 - alpha) * img
+        masked_imgA = add([multiply([alphaA, reconstructed_imgA]), multiply([Lambda(one_minus)(alphaA), img])])
+        masked_imgB = add([multiply([alphaB, reconstructed_imgB]), multiply([Lambda(one_minus)(alphaB), img])])
+        out_discriminatorA = self.netDA(concatenate([masked_imgA, img], axis=-1))
+        out_discriminatorB = self.netDB(concatenate([masked_imgB, img], axis=-1))
+
+        # The adversarial_autoencoder model  (stacked generator and discriminator) takes
+        # img as input => generates encoded represenation and reconstructed image => determines validity 
+        self.adversarial_autoencoderA = Model(img, [reconstructed_imgA, out_discriminatorA])
+        self.adversarial_autoencoderB = Model(img, [reconstructed_imgB, out_discriminatorB])
+        self.adversarial_autoencoderA.compile(loss=['mae', 'mse'],
+                                              loss_weights=[1, 0.5],
+                                              optimizer=optimizer)
+        self.adversarial_autoencoderB.compile(loss=['mae', 'mse'],
+                                              loss_weights=[1, 0.5],
+                                              optimizer=optimizer)
+
+    def converter(self, swap):
+        predictor = self.netGB if not swap else self.netGA
+        return lambda img: predictor.predict(img)
+
+    def build_generator(self):
+        def conv_block(input_tensor, f):
+            x = input_tensor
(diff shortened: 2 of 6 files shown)


Clorr:
Now with masking!

Clorr:
@gdunstone I modified the way we pass arguments to convert plugin. I now pass the full arguments object instead of separate fields; I think it is easier in the long term because the parameters list will grow the more we add plugins...

Feel free to make a feedback if you prefer your version

gdunstone:
I think its better to use explicit keyword arguments with defaults because its self documenting and it provides for the case of the converter being used as a lib, but what you wrote seems fine.

Ill give this version a go at some point soon.

AngryLoki:
When launching with `-p`:
```[12:55:41] [0/num_epochs][0] Loss_DA: 0.246727 Loss_DB: 0.248788 Loss_GA: 0.523137 Loss_GB: 0.475922
  File "faceswap.py", line 24, in <module>
    arguments.func(arguments)
  File "/projects/face-work/faceswap/scripts/train.py", line 21, in process_arguments
    self.process()
  File "/projects/face-work/faceswap/scripts/train.py", line 99, in process
    for name, image in self.preview_buffer.items():
RuntimeError: dictionary changed size during iteration
```

gdunstone:
I changed some code that fixes your issue @AngryLoki :
```
diff --git a/scripts/train.py b/scripts/train.py
index aa536d8..e097e31 100644
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -6,11 +6,14 @@ from lib.utils import get_image_paths
 from lib.cli import FullPaths
 from plugins.PluginLoader import PluginLoader
 
+from threading import Lock
+
 class TrainingProcessor(object):
     arguments = None
 
     def __init__(self, subparser, command, description='default'):
         self.parse_arguments(description, subparser, command)
+        self.lock = Lock()
 
     def process_arguments(self, arguments):
         self.arguments = arguments
@@ -96,8 +99,9 @@ class TrainingProcessor(object):
             print('Using live preview')
             while True:
                 try:
-                    for name, image in self.preview_buffer.items():
-                        cv2.imshow(name, image)
+                    with self.lock:
+                        for name, image in self.preview_buffer.items():
+                            cv2.imshow(name, image)
 
                     key = cv2.waitKey(1000)
                     if key == ord('\n') or key == ord('\r'):
@@ -156,15 +160,16 @@ class TrainingProcessor(object):
         except Exception as e:
             print(e)
             exit(1)
-    
+
     preview_buffer = {}
 
     def show(self, image, name=''):
         try:
-            if self.arguments.preview:
-                self.preview_buffer[name] = image
-            elif self.arguments.write_image:
-                cv2.imwrite('_sample_{}.jpg'.format(name), image)
+            with self.lock:
+                if self.arguments.preview:
+                    self.preview_buffer[name] = image
+                elif self.arguments.write_image:
+                    cv2.imwrite('_sample_{}.jpg'.format(name), image)
         except Exception as e:
             print("could not preview sample")
             print(e)

```

subzerofun:
@Clorr 

Thanks for adding this Plugin! Have two questions about it, hope you can help me out:

1) After training, when i start the conversion script and select the `--detector` option `cnn` i get this error:
```Failed to convert image: (...) Reason: Error while calling cudaMalloc(&data, n) in file /private/var/folders/1j/r_nj68nn3tl7sqfc93_18t_80000gp/T/pip-build-o7algsyj/dlib/dlib/dnn/cuda_data_ptr.cpp:28. code: 2, reason: out of memory```

The output images are 1280 x 720px and i use a GTX 1080 Ti with 11 GB VRAM. I have two cards installed, one for handling the display and OS stuff and the 1080 solely for Machine Learning. So i always have the full 11 GB available for CUDA processing. Was just wondering – **aren't 11 GB enough for this output size**? Can i tweak some parameters – where should i look? I know the alternative option is selecting `hog` and that works, but i guess the `cnn` option yields better results.

2) Is there an option when using the GAN converter to get seamless faces on the output images? I've trained quite a while and still have hard edges around the faces.

Clorr:
Thanks for your feedback. I see you are using cuda with dlib. With loaded model for convert, maybe it starts being too much for the gpu. You should try to profile to see how much memory is used, maybe try nvidia-smi. The cnn option is not better for all, just for rotated and side faces.

The convert with GAN is limited for now, but with small changes, I can make all convert work with GAN. I'll try to add it this week

subzerofun:
Thanks for the info. I've monitored the memory usage, with the `cnn` option it quickly goes up from 2% to 99% (11149 MB) and then the error appears.

Looking forward to the additional options! Have you compared the output quality from the GAN method with the standard one? Is there a significant difference? I can't tell yet, will have to start training my current dataset with the "old" method.

Clorr:
I'm also merging this, because thankfully to trainer arguments, GAN is no more the default value, and the plugin can now live without having impact on the rest of the codebase

Clorr:
Now, I can have a look at the famous GAN 128 ;-)

subzerofun:
@Clorr Do you mean a resolution of 128 x 128 px for the applied faces?

Clorr:
I just mean transforming [this version of @shaoanlu script](https://github.com/shaoanlu/faceswap-GAN/blob/master/FaceSwap_GAN_v2_sz128_train.ipynb) to a plugin. It has 128px large applied face

petergerten:
That's great. It would be cool to have it configurable. So 256px would be possible as well. I guess that is largely a constraint on GPU memory and training time?

subzerofun:
@petergerten I've edited the [Model_Original.py ]( faceswap/plugins/Model_Original.py )(Autoencoder, not Clorrs GAN plugin) to allow the generation of 128x128px faces here: #157. Had to change the [random_warp](https://github.com/deepfakes/faceswap/blob/20753a64b76a156aea17724348269d60dd525f87/lib/training_data.py#L64) function and the [TrainingDataGenerator](https://github.com/deepfakes/faceswap/blob/20753a64b76a156aea17724348269d60dd525f87/lib/training_data.py#L8) with some lines from Clorrs new [GAN128 plugin](https://github.com/Clorr/faceswap/commit/8a22415af7c908683d785e83117aea47a4068a7d) (he has created his [own fork](https://github.com/Clorr/faceswap/) to experiment with) and have trained the 128px model for around 100000 iterations. It seems to work 🎉! My loss values are at `loss_A: 0.0164708, loss_B: 0.0155299` now and the sample output is looking good! It wouldn't take much to edit the `Model_Original.py` to allow for 256x256px input images.

#### Some results of the 128x128px test (Info: i use a GTX 1080 Ti with 11GB):
With 64x64 input images and `ENCODER_DIM` set to 1024 (this variable is also important to keep in mind) i could use a batch size of 220 and one iteration took ~0.5s. After editing the second [Dense](https://github.com/deepfakes/faceswap/blob/20753a64b76a156aea17724348269d60dd525f87/plugins/Model_Original.py#L53) Layer the tensor size changed from 16384 to 65536 (4 * 4 * 1024 -> 8 * 8* 1024). And this 4x data increase of course needs to fit in the GPU memory. If i understand it correctly you have to multiply this layers tensor size with the batch number to get a rough understanding of how much GPU memory it consumes (to be precise you have to do it for all [changed layers](https://github.com/deepfakes/faceswap/blob/20753a64b76a156aea17724348269d60dd525f87/plugins/Model_Original.py#L45-L65) and that quickly adds up to ... _a lot_ 😁). So a `batch size` of 220 would definitely not work with the new memory requirements 😅. I had to reduce it to 20 to not get an out of memory error. First i've tried to also increase the `ENCODER_DIM` from `1024` to `2048`, but that produced an OOM error. I could have lowered the batch size but instead chose to set the encoder dim to `1536`. And now the 20 batches barely fit in the memory. One iteration now takes ~0.6 s (reminder: 220 64x64 batches took 0.5s). Disclaimer: I'm no expert on neural net architecture, so my reasoning behind doubling the values of the last Dense layer was just that it needs to be 2x that of the 64 model. _There is certainly a more efficient way to change the layers_, but i'm relying on the knowledge of others to explain how to find a better solution. Maybe someone with more experience (which is probably a lot of people here 😀) knows if this increase was even necessary. Increasing the `ENCODER_DIM` (and leaving the 2nd Dense layer as it is) probably yields better results. Hope someone can look at that and explain it (in simple terms for me, please :-) ).

**TLDR:** I've changed the original autoencoder model to allow 128x128px input images, but my edit of certain layer dimensions is probably not the best solution as it needs tons of GPU memory and is decreasing the possible batch sizes as well as increasing training time. Which is to expected if the input images are larger, but i'm certain it's possible to reduce the required memory with a better model (by someone who knows what they are actually doing, which is not me).

### Looking at the 256x256px option: 
I will try to edit the model again and look at how much memory this needs. But i see one (small) problem with the 256 input images in general: All faces get rotated, transformed, cropped according to the `coverage` variable (which was 160 in the 64px, and now 220 in Clorrs 128px plugin) and then scaled to the defined internal size (64 or 128 e.g.). So if i take 256x256px image as input, do the transform and warping, then trim it to 220x220 and scale it **down** to 128x128 for the neural net the image would be sharp. But if you want 256x256px internal images you have to scale the 220x220 in the last step **up** again, which gets you a slightly interpolated and blurred image. Don't think this is a big problem, but if you want the best internal image quality you would need to extract images bigger than 256x256 to have to some leeway to crop it. Or maybe i'm just overthinking this.


Then there is the bigger problem of memory consumption, you could only use small batch sizes on a card with low memory and it would take significantly longer to get to acceptable loss values. I will stop the speculation now – this message is already too long. Will modify the model for 256x256 images now and test how long the training takes.

EXJUSTICE:
Is it possible to extract the decoder component of the GAN to allow for a simple decoder exchange and swap generation via calling .predict?

Currently having trouble shortening the code to create a simpler call to predict - all I'm getting is A-to-A autoencoding with the GAN

PR #109, opened 2018-02-03
Misc updates on master before GAN. Added multithreading + mmod face detector

Clorr:
In order to progress on GAN release, I'm pushing a couple of non GAN related changes.

This release brings multithreading, like what @iSevenDays suggested in #62

The code is mostly clean but could be improved, so feel free to test and give a feedback

### DIFF
--- a/lib/FaceFilter.py
+++ b/lib/FaceFilter.py
@@ -6,15 +6,18 @@ import face_recognition
 class FaceFilter():
     def __init__(self, reference_file_path, threshold = 0.6):
         image = face_recognition.load_image_file(reference_file_path)
-        self.encoding = face_recognition.face_encodings(image)[0] # Note: we take only first face, so the reference file should only contain one face. We could also keep all faces found and filter against multiple faces
+        self.encoding = face_recognition.face_encodings(image)[0] # Note: we take only first face, so the reference file should only contain one face.
         self.threshold = threshold
     
     def check(self, detected_face):
-        encodings = face_recognition.face_encodings(detected_face.image)[0] # we could use detected landmarks, but I did not manage to do so
-        score = face_recognition.face_distance([self.encoding], encodings)
-        print(score)
-        return score <= self.threshold
-
+        encodings = face_recognition.face_encodings(detected_face.image) # we could use detected landmarks, but I did not manage to do so. TODO The copy/paste below should help
+        if encodings is not None and len(encodings) > 0:
+            score = face_recognition.face_distance([self.encoding], encodings[0])
+            print(score)
+            return score <= self.threshold
+        else:
+            print("No face encodings found")
+            return False
 
 # # Copy/Paste (mostly) from private method in face_recognition
 # face_recognition_model = face_recognition_models.face_recognition_model_location()
--- a/lib/ModelAE.py
+++ b/lib/ModelAE.py
@@ -2,7 +2,7 @@
 
 import time
 import numpy
-from lib.training_data import minibatchAB, stack_images
+from lib.training_data import TrainingDataGenerator, stack_images
 
 encoderH5 = '/encoder.h5'
 decoder_AH5 = '/decoder_A.h5'
@@ -40,11 +40,20 @@ class ModelAE:
         print('saved model weights')
 
 class TrainerAE():
+    random_transform_args = {
+        'rotation_range': 10,
+        'zoom_range': 0.05,
+        'shift_range': 0.05,
+        'random_flip': 0.4,
+    }
+
     def __init__(self, model, fn_A, fn_B, batch_size=64):
         self.batch_size = batch_size
         self.model = model
-        self.images_A = minibatchAB(fn_A, self.batch_size)
-        self.images_B = minibatchAB(fn_B, self.batch_size)
+
+        generator = TrainingDataGenerator(self.random_transform_args, 160)
+        self.images_A = generator.minibatchAB(fn_A, self.batch_size)
+        self.images_B = generator.minibatchAB(fn_B, self.batch_size)
 
     def train_one_step(self, iter, viewer):
         epoch, warped_A, target_A = next(self.images_A)
--- a/lib/cli.py
+++ b/lib/cli.py
@@ -1,7 +1,6 @@
 import argparse
 import os
 import time
-from tqdm import tqdm
 
 from pathlib import Path
 from lib.FaceFilter import FaceFilter
@@ -28,7 +27,6 @@ class DirectoryProcessor(object):
 
     verify_output = False
     images_found = 0
-    images_processed = 0
     faces_detected = 0
 
     def __init__(self, subparser, command, description='default'):
@@ -48,18 +46,13 @@ class DirectoryProcessor(object):
             print('Input directory not found. Please ensure it exists.')
             exit(1)
 
-        self.images_found = len(self.input_dir)
         self.filter = self.load_filter()
         self.process()
         self.finalize()
         
     def read_directory(self):
-        for filename in tqdm(self.input_dir):
-            if self.arguments.verbose:
-                print('Processing: {}'.format(os.path.basename(filename)))
-
-            yield filename
-            self.images_processed = self.images_processed + 1
+        self.images_found = len(self.input_dir)
+        return self.input_dir
     
     def get_faces(self, image):
         faces_count = 0
@@ -125,7 +118,6 @@ class DirectoryProcessor(object):
     def finalize(self):
         print('-------------------------')
         print('Images found:        {}'.format(self.images_found))
-        print('Images processed:    {}'.format(self.images_processed))
         print('Faces detected:      {}'.format(self.faces_detected))
         print('-------------------------')
 
--- a/lib/cli.py
+++ b/lib/cli.py
@@ -63,7 +63,7 @@ class DirectoryProcessor(object):
     
     def get_faces(self, image):
         faces_count = 0
-        for face in detect_faces(image):
+        for face in detect_faces(image, self.arguments.detector):
             if self.filter is not None and not self.filter.check(face):
                 print('Skipping not recognized face!')
                 continue
--- a/lib/cli.py
+++ b/lib/cli.py
@@ -78,7 +78,7 @@ class DirectoryProcessor(object):
             self.verify_output = True
     
     def load_filter(self):
-        filter_file = "filter.jpg" # TODO Pass as argument
+        filter_file = self.arguments.filter
         if Path(filter_file).exists():
             print('Loading reference image for filtering')
             return FaceFilter(filter_file)
--- a/lib/faces_detect.py
+++ b/lib/faces_detect.py
@@ -2,8 +2,8 @@ import dlib
 import face_recognition
 import face_recognition_models
 
-def detect_faces(frame):
-    face_locations = face_recognition.face_locations(frame)
+def detect_faces(frame, model="hog"):
+    face_locations = face_recognition.face_locations(frame, model=model)
     landmarks = _raw_face_landmarks(frame, face_locations)
 
     for ((y, right, bottom, x), landmarks) in zip(face_locations, landmarks):
--- a/lib/multithreading.py
+++ b/lib/multithreading.py
@@ -2,12 +2,15 @@ import multiprocessing as mp
 
 method = None
 
-def pool_process(method_to_run, data):
+def pool_process(method_to_run, data, processes=None):
     global method
+    if processes is None:
+        processes = mp.cpu_count()
     method = method_to_run
-    pool = mp.Pool()
+    pool = mp.Pool(processes=processes)
+
     for i in pool.imap_unordered(runner, data):
         yield(i)
-    
+        
 def runner(item):
     return method(item)
--- a/lib/multithreading.py
+++ b/lib/multithreading.py
@@ -10,7 +10,7 @@ def pool_process(method_to_run, data, processes=None):
     pool = mp.Pool(processes=processes)
 
     for i in pool.imap_unordered(runner, data):
-        yield(i)
-        
+        yield i if i is not None else 0
+    
 def runner(item):
     return method(item)
--- a/lib/multithreading.py
+++ b/lib/multithreading.py
@@ -1,0 +1,13 @@
+import multiprocessing as mp
+
+method = None
+
+def pool_process(method_to_run, data):
+    global method
+    method = method_to_run
+    pool = mp.Pool()
+    for i in pool.imap_unordered(runner, data):
+        yield(i)
+    
+def runner(item):
+    return method(item)
--- a/lib/training_data.py
+++ b/lib/training_data.py
@@ -5,97 +5,90 @@ from random import shuffle
 from .utils import BackgroundGenerator
 from .umeyama import umeyama
 
-coverage = 220 # Coverage of the face for training. Larger value will cover more features. @shaoanlu recommends 220. Original is 160
+class TrainingDataGenerator():
+    def __init__(self, random_transform_args, coverage):
+        self.random_transform_args = random_transform_args
+        self.coverage = coverage
 
-random_transform_args = {
-    'rotation_range': 10,
-    'zoom_range': 0.05,
-    'shift_range': 0.05,
-    'random_flip': 0.4,
-}
+    def minibatchAB(self, images, batchsize):
+        batch = BackgroundGenerator(self.minibatch(images, batchsize), 1)
+        for ep1, warped_img, target_img in batch.iterator():
+            yield ep1, warped_img, target_img
 
-# GAN
-# random_transform_args = {
-#     'rotation_range': 20,
-#     'zoom_range': 0.1,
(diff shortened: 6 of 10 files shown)


Clorr:
Now with mmod_human_face_detector! Add the -D cnn argument to activate it. But be careful, it is really slow and time consuming, but it detects faces from many more angles

gdunstone:
Is the accuracy higher for this one? And can we specify this for the convert as well? I have had some issues with inaccurate scaling with the current detector for convert but I have processing power up the wazoo.

facepainter:
tl;dr - I get a  "NoneType is not callable" when extracting

My python is not the strongest, need to RTFM myself but to me the global use in #109 lib/multithreading.py seems odd.

Can runner call (processFiles) via the 'method' global? I thought it was not possible to use a global between two separate processes like this... If I do a quick test of this... indeed runner throws "NoneType is not callable".

Again - don't quote me - but pretty sure you need Queue or Pipe, to share picklable stuff between processes... and as this is a bound method I'm not sure that would work.

Maybe I'm too tired or not high enough...totally disregard if this makes no sense - I could well be doing something odd :)

Clorr:
@gdunstone the mmod face detector is in extract & convert, it wouldn't make sense otherwise ;-)

@seattleg yes, the mmod detector is ultra slow on large images, but it remains an alterntive, it is not used by default

@facepainter I'm not really satisfied of how I handled `lib/multithreading.py` but I got it working like that.. The `global method` shouldn't be a big deal because it is written once, and read multiple times, which should be thread-safe. If anyone has a better solution, I'm fine, it also why I did a separate file so that it is simple to modify it

gdunstone:
What's the benefit of using Process over Pool in this instance? It seems like it adds some unnecessary complexity.

Clorr:
Hmm, can you point out which line it is? In `lib/multithreading.py`, I'm using `Pool` because it was almost a one liner, but I had to use an intermediate var because the `map/imap` functions where not ok with a class method

gdunstone:
whoops my bad, I was going off the old version proposed here:

https://github.com/deepfakes/faceswap/pull/62/commits/c4a65ebc147d39105b1acee5345ec756a403b9b7

facepainter:
@Clorr - Hmm yes I still get the error, I get 0 results and the following in the console.

```
  File "U:\faceswap\faceswap.py", line 24, in <module>
    arguments.func(arguments)
  File "U:\faceswap\lib\cli.py", line 53, in process_arguments
    self.process()
  File "U:\faceswap\scripts\extract.py", line 30, in process
    self.faces_detected = sum(pool_process(self.processFiles, list(self.read_directory())))
  File "U:\faceswap\lib\multithreading.py", line 9, in pool_process
    for i in pool.imap_unordered(runner, data):
  File "U:\Python36\Lib\multiprocessing\pool.py", line 735, in next
    raise value
TypeError: 'NoneType' object is not callable
```

> The global method shouldn't be a big deal because it is written once, and read multiple times, which should be thread-safe.

Not pretending to know loads about this or anything, like I said my python is well rusty ( that sounds wrong) ...anyway, basically multi-threading!=multi-processing. Threads share the same memory, processes don't - they have....different process ID, etc. It is confusing because 

```
def runner(item):
    return method(item) 
```

is right there....but "method" here is not "global method" above - this is because when runner is called ```pool.imap_unordered(runner, data)``` it starts a whole new process that has its own memory. At least as I understand it. Or else am I totally missing the point!?! Was pretty sure that is why it was failing for me.... Tell me to shut up if I am missing something silly :) Just really want it to be fassssst :)

Clorr:
@facepainter I see from your stack that you are using python 3.6, while I'm using 3.5, maybe the behavior have change in between. I took reference [here](https://stackoverflow.com/questions/423379/using-global-variables-in-a-function-other-than-the-one-that-created-them) for the `global`.

Also on the thread/process questions, I took what I found in Python, as I'm still beginner in Python, so maybe it is not the right way to go. If you have better samples to show me, feel free.

Clorr:
@facepainter if you find a patch for your platform, let me know. I'm checking on my side, but I don't want to upgrade python for now...

Clorr:
I tried with

```
from concurrent.futures import ThreadPoolExecutor

executor = ThreadPoolExecutor()

def pool_process(method_to_run, data):
    for i in executor.map(method_to_run, data):
        yield i if i is not None else 0
```

And it seems to perform as well as the previous solution. Can you tell me if it is ok for you @facepainter ?

facepainter:
@Clorr

>   I took reference here for the global.

Ah yeah - that wont work across process at all. 
If you imagine, you can't spawn a process - then have that process do a bunch of stuff in memory on another process (i.e. the one that spawned it), you have to use IPC rather than shared memory when doing such things. 

multi threading/processing has a bunch of caveats to it - mainly that anything you pass across processes has to be picklable (serialisable). 

(As an aside this is actually where electron/nodejs excels - it is totally built on the asynchronous IPC model - almost enforces it throughout...more my background too).

Will test now with the thread pool but at a glance that looks good :)

Clorr:
> If it worked for you and you are linux? Then it must be linux/windows os.fork BS. Wouldn't be surprised if global behaves differently here somehow across platform :(
Yes it is annoying

For the multi processing, the sharing of data is a problem, but here, we don't really have to share data, because each task handles one image and saves it, so it simplifies things.

facepainter:
@Clorr  yeah threading is the way to do it, much simpler and avoids the memory isolation / pickling issues :) - that is really good in the profiler too, much faster :)

OOI what are you using to profile/benchmark - i've been using the Python Profiling in Visual Studio 2017  and I'm really impressed.

Clorr:
Ok, let's go for that.

I'm just benchmarking by running an extract with 400 images

facepainter:
lol just RTFM now - https://docs.python.org/3/library/multiprocessing.html#module-multiprocessing.dummy

there is actually a ThreadPool in multiprocessing.  Just not well doc'd at all - not this but something very like...

Using imap_unordered (like you had with the processes) is going to be hella faster than ThreadPoolExecutor.map - plus this was there  is no dependence of futures :) Still, it could be a lot prettier sans globals here....

Really don't want to tread on toes or offend here :) I'm happy to play with this for a couple of hours to get is solid?

```
from multiprocessing.pool import ThreadPool

method = None

def pool_process(method_to_run, data):
    global method
    method = method_to_run
    pool = ThreadPool()
    for i in pool.imap_unordered(runner, data):
        yield(i)
    
def runner(item):
    return method(item)
```

gdunstone:
I profiled these two as ive never seen ThreadPool before. Its hella slow in comparison to a normal Pool
I think its because pythons threading isnt real threading (thanks GIL), and multiprocessing actually has the ability to do real multiprocessing.

```python3

import multiprocessing as mp
from multiprocessing.pool import ThreadPool

method = None

def pool_process(method_to_run, data, j=None):
    global method
    if j is None:
        j = mp.cpu_count() * 2 # I though *2 for HT if you have intel, made no difference
    method = method_to_run
    pool = ThreadPool(threads)
    for i in pool.imap_unordered(runner, data):
        yield(i)

def runner(item):
    return method(item)
```

facepainter:
@gdunstone  - Ah I didn't figure on that overhead - was just thinking imap_unordered vs map :( ...There isn't much info on  ThreadPool at all :(

gdunstone:
Its not that there is an overhead, but that Thread is not **full parallelism**, just concurrent.

https://code.tutsplus.com/articles/introduction-to-parallel-and-concurrent-programming-in-python--cms-28612

facepainter:
@gdunstone - Ah interesting link - thanks :) I really just meant the cost difference of the alternate approaches really - again not claiming to be an expert or anything.

My take away is whilst multiprocessing is not affected by GIL, you cant use non picklable objects...and that there are pretty huge differences between Linux and Windows when it comes to processes, shared memory, and concurrency mechanisms in python (and in general really).

Clorr:
Unless other complaints, I will merge this tomorrow so we can move on

facepainter:
@Clorr  hey totally wasn't a complaint at all - just an observation that using globals across processes won't work on windows :) didn't mean for it to get long winded or to offend :)

Clorr:
No problem, it is just a way to say "feedback" ;-)

gdunstone:
@Clorr Great work on https://github.com/deepfakes/faceswap/pull/109/commits/c83f800c11983147d8289479c2cb06e45ec50f80

Apollo122:
Hey guys,
Have anyone tested this PR on Windows? I can take a look if not.

Clorr:
It would be nice, because even if someone has, maybe he hasn't the same config as you

Apollo122:
Alright then.
Is there any specific config should I use? Or just regular extract, train and convert?

Clorr:
Just regular commands. You can also try different plugins, but that's not mandatory; more specific cases may be raised as issues

Apollo122:
Extract doesnt work. 
`multiprocessing.pool.RemoteTraceback:
"""
  File "C:\IntelPython3\lib\multiprocessing\pool.py", line 119, in worker
    result = (True, func(*args, **kwds))
  File "K:\faceswap-master\lib\multithreading.py", line 16, in runner
    return method(item)
TypeError: 'NoneType' object is not callable
"""

The above exception was the direct cause of the following exception:

  File "faceswap.py", line 29, in <module>
    arguments.func(arguments)
  File "K:\faceswap-master\lib\cli.py", line 50, in process_arguments
    self.process()
  File "K:\faceswap-master\scripts\extract.py", line 46, in process
    for _ in tqdm(pool_process(self.processFiles, files, processes=processes), total = len(files)):
  File "C:\IntelPython3\lib\site-packages\tqdm\_tqdm.py", line 955, in __iter__
    for obj in iterable:
  File "K:\faceswap-master\lib\multithreading.py", line 12, in pool_process
    for i in pool.imap_unordered(runner, data):
  File "C:\IntelPython3\lib\multiprocessing\pool.py", line 735, in next
    raise value
TypeError: 'NoneType' object is not callable`

Clorr:
Ah yes, this has been discussed above. You can replace the code given by @gdunstone in it's comment https://github.com/deepfakes/faceswap/pull/109#issuecomment-362954373 The code goes in `lib/multithreading.py`. If it doesn't work let us know, or try another of the versions above

Apollo122:
I got the same error for both codes above.

Loading Extract from Extract_Align plugin...
  File "faceswap.py", line 29, in <module>
    arguments.func(arguments)
  File "K:\faceswap-master\lib\cli.py", line 50, in process_arguments
    self.process()
  File "K:\faceswap-master\scripts\extract.py", line 46, in process
    for _ in tqdm(pool_process(self.processFiles, files, processes=processes), total = len(files)):
TypeError: pool_process() got an unexpected keyword argument 'processes'

Clorr:
Oh, it is because of the signature of the copy/paste'd code does not have the argument `processes`. It is named `j` in the copy/paste'd code I showed you, it should be renamed. Test if it is still ok for you, and maybe push a correction, if not I will finalize on my side tomorrow

Apollo122:
So this one didnt work
python faceswap.py extract -i photos/A -o data/A --processes 2

but this worked: python faceswap.py extract -i photos/A -o data/A --processes 1

but its not utilizing multi-threading anymore

Clorr:
Yes, because with 1 process, it does not call the 'bad' part with the `processes` argument. But if other parts are ok, let me know

Apollo122:
Ok i changed both "j" and "threads" with "processes" and its working. 
i tried with --processes 8 but it still takes same time. I didn't see any improvement with remaining time.

Clorr:
hmm strange. Have you tried opening the resource monitor to see if all cores are used? also, how big is your data set ? If it is too small, the startup overhead may compensate the parallelization

Apollo122:
yeah i see the splikes. 
on task manager, it shows under CPU python.exe uses 13.
Im extracting 640 images

Clorr:
13% ? If you have 4 cores + hyperthreading, it is likely that it uses only one core (100% / 8 = 12.5%)

640 images is largely enough for the test.

I'll have more in-depth look about it tomorrow...

Apollo122:
i have 4 cores yes. 
btw i tried train got this error and is stuck.
Unable to create file (unable to open file: name = 'models/encoder.h5', errno = 2, error message = 'No such file or directory', flags = 13, o_flags = 302)

created the folder manually and it worked.

Clorr:
Ah thanks for this bug. maybe I should add a check for the output folder

Apollo122:
So i tested the convert as well seems fine.
Although I have a suggestion about it:

Right now convert takes a full frame. First it detects the face then crops it, runs model and finally merges it back.
But we already done half the job during extracting, we detected the face and cropped it to train on it. So it doesn't make sense to do it all over again(especially if we want to use cnn). Also algorithm may miss some faces or output some garbage images. With this approach we have absolutely no control what images it uses. We should use the extracted images to process and merge those on the frame. For example 1adrianb's face-detector outputs alignment.json after extraction. With using something like that we would know where the face is on the frame and where to merge it without processing again.

facepainter:
@Clorr have been playing with this myself - the face_detection lib actually allows you to batch process upto 128 frames at a time using a CNN model. You just have to make sure that you compile dlib with CUDA extensions and then you get to use the GPU too :) totally removes the ffmpeg splitting to frames step :)  (oh - all the frames do have to be the exact same dimensions!).

I have just been testing it - the idea is you give it a video, a small sample of images of a person you want (say a directory with 5-10 good shots). an output folder and a batch size i.e.

`faceswap.py extract -i videoOfX.mp4 -r ./ImagesOfX -o ./grabedFramesX -b 32`

It is actually working pretty well - it is faster - and as it negates first extracting and loading the frames as a step outside the application in ffmpeg or whatever. 

I even put together a version with youtube-dl included so you can do...

`faceswap.py extract -i 2z3JUS1aXdY  -r ./ImagesOfX -o ./grabedFramesX -b 32`

Where **2z3JUS1aXdY**  is the id of a youtube video (it just gets the best version, then does as above).

Really just all just based off this example: https://github.com/ageitgey/face_recognition/blob/master/examples/find_faces_in_batches.py

Anyhow thought you might be interested - apologies if this is something that has already been looked at :)

Clorr:
@Apollo122 supporting an `alignment.json` file is totally feasible because it is just a matter of saving the landmarks that are generated on extract. Just this is not related to this PR so maybe discuss it in a specific issue.

@facepainter if you have some code to propose, we are totally open to it ;-) Feel free to make a PR. Just be careful to make it after this one has been merged, so we don't get too much conflicts

Apollo122:
I classified that as misc update also :)
But sure can be opened as different issue

Clorr:
I'm merging this now because we should move on, don't hesitate to open issues if you find bugs...

modelsex:
cnn mode work perfect on Extract, but I get follow problem when use cnn to convert:

Command example: python faceswap.py convert -i D:/project4/data_A/ -o D:/face/final/ -m D:/face/project4/models/ -D cnn 

My Env. CUDA-8.0 | cudnn-8.0 | dlib-19.9

**Log output**
name: GeForce GTX 1080 major: 6 minor: 1 memoryClockRate(GHz): 1.7715
pciBusID: 0000:01:00.0
totalMemory: 8.00GiB freeMemory: 6.62GiB
2018-02-08 19:24:41.501735: I C:\tf_jenkins\home\workspace\rel-win\M\windows-gpu\PY\36\tensorflow\core\common_runtime\gpu\gpu_device.cc:1120] Creating TensorFlow device (/device:GPU:0) -> (device: 0, name: GeForce GTX 1080, pci bus id: 0000:01:00.0, compute capability: 6.1)
loaded model weights
Loading Convert from Convert_Masked plugin...
  0%|                                                                                          | 0/800 [00:00<?, ?it/s]Failed to convert image: D:\project4\data_A\kar0001.png. **Reason: Error while calling cudaMalloc(&data, n) in file D:\dlib-master\dlib\dnn\cuda_data_ptr.cpp:28. code: 2, reason: out of memory**
  0%|▏                                                                                 | 2/800 [00:00<04:24,  3.02it/s]Failed to convert image: D:\project4\data_A\kar0002.png. **Reason: Error while calling cudaGetLastError() in file D:\dlib-master\dlib\dnn\gpu_data.cpp:91. code: 2, reason: out of memory**
Failed to convert image: D:\project4\data_A\kar0003.png. Reason: Error while calling cudaMalloc(&data, n) in file D:\dlib-master\dlib\dnn\cuda_data_ptr.cpp:28. code: 2, reason: out of memory
  0%|▍                                                                                 | 4/800 [00:00<02:34,  5.16it/s]Failed to convert image: D:\project4\data_A\kar0004.png. Reason: Error while calling cudaGetLastError() in file D:\dlib-master\dlib\dnn\gpu_data.cpp:91. code: 2, reason: out of memory
......

Clorr:
It is because convert loads cnn **+** model, and for some reason they take both much memory. I have added a patch in #140 that tells tensorflow to allocate memory on-demand and not all at once. I don't gaurantee it works, but if you can try and give feedback it would be nice....

PR #121, opened 2018-02-05
Fix breakage on some versions of python3 (on Ubuntu 17.10 for sure)

fat-tire:
Also gives better info when you don't include any arguments (or bad ones).  Didn't test on a non-Linux
machine or other version of python than 3.6.3, so someone should do that and see if it hurts anything.


gdunstone:
Thanks!

gdunstone:
tested on python3.6.4 as well.
this is unrelated to the python version though.

PR #129, opened 2018-02-06
adviser from advuser

geebrown43:
Changed typo from advuser to adviser


Clorr:
"advuser" was a keyword for advanced-user, but I agree it is not that clear ^^

PR #151, opened 2018-02-08
[Draft] Adding GAN128

Clorr:
This creates a plugin from the [GAN 128 version](https://github.com/shaoanlu/faceswap-GAN/blob/master/FaceSwap_GAN_v2_sz128_train.ipynb) from @shaoanlu

This is a draft to get your feedback and no convert is possible at the moment


Apollo122:
Does it include perception loss? According to @shaoanlu it's pretty important.

https ://github.com/shaoanlu/faceswap-GAN/issues/50#issuecomment-366720454

shaoanlu:
From my experiments, introducing perceptual loss can:

1. Make eyeball direction be consistent with input face.

2. Inject some prior knowledge to mask prediction, making output mask more smooth.

Otherwise there is little difference. I got similar masked faces with/without PL as shown in the issue.

Apollo122:
Smooth is better :)
What makes you think that PL will cause more issue post @shaoanlu? I thought procedure is automatic. What could cause problems for users? And more realistic eyeball movement is pretty important because right now they look like dolls :) Also there are lots of issue posts already about gan. It's quality, time it needs and masked output. With PL we would at least stop masked post because users will see the results prior as far as I understand.

petergerten:
Is anything missing here until it can be merged?
I have some GAN128 models trained with shaoanlu's notebook I would like to use :)

plugins/Model_GAN128/Model.py
```diff
@@ -0,0 +1,177 @@
...
+            out64 = Conv2D(64, kernel_size=3, padding='same')(x)
+            out64 = LeakyReLU(alpha=0.1)(out64)
+            out64 = Conv2D(3, kernel_size=5, padding='same', activation="tanh")(out64)
+            
+            x = upscale_ps(32)(x)
+            x = res_block(x, 32)
+            x = res_block(x, 32)
+            alpha = Conv2D(1, kernel_size=5, padding='same', activation="sigmoid")(x)
+            rgb = Conv2D(3, kernel_size=5, padding='same', activation="tanh")(x)
+            out = concatenate([alpha, rgb])
```
oatssss: Why do you concatenate here?

oatssss:
I think this change also suffers from the conversion issue I was having with v2 port I did. See commit https://github.com/deepfakes/faceswap/pull/217/commits/ca2af7128a79cb4bfcdc552c8955fc21ff466def for how I fixed.

petergerten:
Looks like the GAN128 converter is missing!?

petergerten:
I created "Convert_GAN128.py" copying the original GAN converter and changed
`        face = cv2.resize(face_detected.image, (128, 128))`
But I get the following error:

>   File "/data/plugins/Convert_GAN128.py", line 14, in patch_image
>     new_face = mask * new_face + (1 - mask) * face
> ValueError: operands could not be broadcast together with shapes (1,128,128,4) (1,64,64,3) 

Although I selected the GAN128 model, for some reason there seem to be still 64px faces

oatssss:
I had the same problem when updating gan64 to v2. See commit https://github.com/deepfakes/faceswap/pull/217/commits/ca2af7128a79cb4bfcdc552c8955fc21ff466def
 for how I fixed it.

Note, using the same fix won't work because the gan128 model [returns a list of 2 items](https://github.com/deepfakes/faceswap/pull/151/files#diff-9a22c7ad248bbd42edb82b0285838a44R119): `[out, out64]`. So the resulting array from `self.encoder(face)` needs to be sliced differently to get just the `[out]` we're interested in. I dunno the shape of the array because I don't normally code in Python, but my first guess would be something like

``` python
fake_output = self.encoder(face)
mask = fake_output[0,:,:, :1]
new_face = fake_output[0,:,:, 1:]
```

oatssss:
try:

``` python
mask = fake_output[0][:,:,:, :1]
new_face = fake_output[0][:,:,:, 1:]
```

TooMuchFun:
@oatssss 

That fix works for me just great, thanks for fixing what I spent all last Sunday trying to figure out. Have you tried out the results of converting out of GAN? My models are really well trained and look great in the preview trainer but quality really drops in the convert step. The bounding box around the face remains very clear and apparent. My model has over 20k iterations with a batch size 16.

Are you seeing this too?

plugins/Model_GAN128/Trainer.py
```diff
@@ -0,0 +1,222 @@
...
+            loss_G += pl_params[1] * K.mean(K.abs(fake_feat28 - real_feat28))
+            loss_G += pl_params[2] * K.mean(K.abs(fake_feat55 - real_feat55))
+        
+        return loss_D, loss_G
+
+    def show_sample(self, display_fn):
+        _, wA, tA = next(self.train_batchA)
+        _, wB, tB = next(self.train_batchB)
+        display_fn(self.showG(tA, tB, self.path_A, self.path_B), "raw")
+        display_fn(self.showG(wA, wB, self.path_bgr_A, self.path_bgr_B), "masked")
```
TooMuchFun: The masked image preview is quite warped. Should this be passing `tA, tB` instead of `wA, wB`? Commit https://github.com/deepfakes/faceswap/commit/aa063975021f18fb5a7f2e7b824dca21d5276008 from @oatssss fixes this on his GAN64 v2 branch.

Related, https://github.com/deepfakes/faceswap-playground/issues/65

plugins/Model_GAN128/Trainer.py
```diff
@@ -0,0 +1,222 @@
...
+        fake_output = netG.outputs[0]
+        fake_output64 = netG.outputs[1]
+        alpha = Lambda(lambda x: x[:,:,:, :1])(fake_output)
+        rgb = Lambda(lambda x: x[:,:,:, 1:])(fake_output)
+        
+        masked_fake_output = alpha * rgb + (1-alpha) * distorted_input 
+
+        fn_generate = K.function([distorted_input], [masked_fake_output])
+        fn_mask = K.function([distorted_input], [concatenate([alpha, alpha, alpha])])
+        fn_abgr = K.function([distorted_input], [concatenate([alpha, rgb])])
```
oatssss: @shaoanlu what would we use `fn_abgr` for?

plugins/Model_GAN128/Trainer.py
oatssss: @shaoanlu what information does the "raw" preview give us? "raw" and "masked" both seem to give a masked output, but I believe "raw" is manually computed using the output rgb and alpha channels.

plugins/Model_GAN128/Trainer.py
```diff
@@ -0,0 +1,222 @@
...
+        figure = np.concatenate([figure_A, figure_B], axis=0 )
+        figure = figure.reshape((4,self.batch_size // 2) + figure.shape[1:])
+        figure = stack_images(figure)
+        figure = np.clip((figure + 1) * 255 / 2, 0, 255).astype('uint8')
+        return figure
+
+    def showG_mask(self, test_A, test_B, path_A, path_B):
+        figure_A = np.stack([
+            test_A,
+            (np.squeeze(np.array([path_A([test_A[i:i+1]]) for i in range(test_A.shape[0])])))*2-1,
```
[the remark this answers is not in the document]
shaoanlu: The bgr output has range [-1,+1] and the mask [0, 1]. Here we scale the mask to [-1, +1] for display purpose.

plugins/Model_GAN128/Trainer.py
shaoanlu: I used it for debugging and forgot to delete it (not sure if it is called somewhere in the script).

shaoanlu:
Sorry for the late reply, my computer was broken and I have very little access to github.

I think line 184 should be the masked result and 185 for raw result (which means before masking). `tA` and `tB` are non-warped faces  and `wA` and `wB` warped ones.

oatssss:
Actually, what is the purpose of warping the faces during training?

shaoanlu:
It is inherited from the original deepfakes' script. He said its inspired by [this paper](https://arxiv.org/abs/1706.02932v2). My understanding is that the warping is introduced to make the model perform batter on this specific task (face swapping). For example, in test phase, we feed face B in to generator A, generator A will treat face B as if it is a warped face A and reconstruct it to a face-A-look-alike face B. Also the warping can function as regularization to prevent overfit.

But recently I becomes suspicious about this warping because I can got good result either when I reduce the strength of warping, and wonder if anyone would like to run some experiments w/o warping.

Clorr:
I also wondered about the warping thing. I came to the conclusion it is made so that the model does not take too much parameter from the original face like eye distance, chin shape, and so on. 

What we want in faceswap is mostly get high level features like face orientation, eyes and mouth closed or not but we don't want the output to resemble the original face too much. (I hope I'm clear ;-) )

But as @shaoanlu said, the fact the encoder is shared, lowers the probability for it to embed too much of the original face. At some point I also suggested we train an encoder with many different faces, just to get the common high level features, and then we could train only decoders. But maybe I'm still missing a point. For now I'm trying to find a dataset that would help me try this solution....

kcimit:
Any prospects on when it is going to be merged with repo?

Clorr:
This one is going to be replaced by #217 which is close to the end AFAIK

PR #155, opened 2018-02-08
Added serializers

gdunstone:
This speeds up convert speed x4 on my machine.

added multiple types of serializers.
I had to change the way that `faces_detect.DetectedFace` works a little bit to get serialization working correctly. Should still be backwards compatible.

added cli arguments for serializer and path to alignments file. providing no values will write alignments.json to the output dir (maybe it should write it to the input dir?)

The structure I'm using for the alignments is:
```python
{
    filename: [
            {
                "x": face.x,
                "w": face.w,
                "y": face.y,
                "h": face.h,
                "landmarksXY": face.landmarksAsXY()
            },
        ...
    ]
    ...
}

### DIFF
--- a/lib/Serializer.py
+++ b/lib/Serializer.py
@@ -1,0 +1,83 @@
+import json, pickle
+class Serializer(object):
+    ext = "wtf"
+    woptions = "????"
+    roptions = "????"
+
+    @classmethod
+    def marshal(cls, input_data):
+        raise NotImplementedError()
+
+    @classmethod
+    def unmarshal(cls, input_string):
+        raise NotImplementedError()
+
+
+class YAMLSerializer(Serializer):
+    ext = "yml"
+    woptions = "w"
+    roptions = "r"
+
+    @classmethod
+    def marshal(cls, input_data):
+        try:
+            import yaml
+        except ImportError:
+            print("You must have PyYAML installed to use YAMLSerializer")
+        return yaml.dump(input_data, default_flow_style=False)
+
+    @classmethod
+    def unmarshal(cls, input_string):
+        try:
+            import yaml
+        except ImportError:
+            print("You must have PyYAML installed to use YAMLSerializer")
+        return yaml.load(input_string)
+
+
+
+class JSONSerializer(Serializer):
+    ext = "json"
+    woptions = "w"
+    roptions = "r"
+
+    @classmethod
+    def marshal(cls, input_data):
+        return json.dumps(input_data, indent=2)
+
+    @classmethod
+    def unmarshal(cls, input_string):
+        return json.loads(input_string)
+
+
+class PickleSerializer(Serializer):
+    ext = "p"
+    woptions = "wb"
+    roptions = "rb"
+
+    @classmethod
+    def marshal(cls, input_data):
+        return pickle.dumps(input_data)
+
+    @classmethod
+    def unmarshal(cls, input_bytes):
+        return pickle.loads(input_bytes)
+
+def get_serializer(serializer):
+    if serializer == "yaml":
+        return YAMLSerializer
+    if serializer == "json":
+        return JSONSerializer
+    if serializer == "pickle":
+        return PickleSerializer
+    raise NotImplementedError()
+
+
+def get_serializer_fromext(ext):
+    if ext in (".yaml", ".yml"):
+        return YAMLSerializer
+    if ext == ".json":
+        return JSONSerializer
+    if ext == ".p":
+        return PickleSerializer
+    raise NotImplementedError("No serializer matching that file type.")
--- a/lib/cli.py
+++ b/lib/cli.py
@@ -64,7 +64,7 @@ class DirectoryProcessor(object):
 
     def read_alignments(self):
 
-        fn = os.path.join(self.arguments.input_dir,"alignments.{}".format(self.serializer.ext))
+        fn = os.path.join(str(self.arguments.input_dir),"alignments.{}".format(self.serializer.ext))
         if self.arguments.alignments_path is not None:
             fn = self.arguments.alignments_path
 
@@ -79,7 +79,7 @@ class DirectoryProcessor(object):
 
     def write_alignments(self):
 
-        fn = os.path.join(self.output_dir,"alignments.{}".format(self.serializer.ext))
+        fn = os.path.join(str(self.input_dir), "alignments.{}".format(self.serializer.ext))
         if self.arguments.alignments_path is not None:
             fn = self.arguments.alignments_path
         try:
@@ -100,7 +100,7 @@ class DirectoryProcessor(object):
 
     def get_faces_alignments(self, filename, image):
         faces_count = 0
-        faces = self.faces_detected[filename]
+        faces = self.faces_detected[os.path.basename(filename)]
         for rawface in faces:
             face = DetectedFace(**rawface)
             face.image = image[face.y : face.y + face.h, face.x : face.x + face.w]
--- a/lib/cli.py
+++ b/lib/cli.py
@@ -79,7 +79,7 @@ class DirectoryProcessor(object):
 
     def write_alignments(self):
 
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by Apollo122** [comment]
Doesn't really matter where its written but original code writes it to input folder.

I'm willing to test this, so I'll get the file from extract.py and how should I use it for swap and merge? Should I write any parameters, will it only convert the files in A dataset?

**Comment by Apollo122** [comment]
So i checked the convert code and it seems that it doesn't use whats in the alignment.json only. 
I think if user passes alignment file, code should only care whats in it and discard rest of the images and never call detect face because they have been detected before.

`    def prepare_images(self):
        self.read_alignments()
        for filename in tqdm(self.read_directory()):
            image = cv2.imread(filename)

            if self.have_face(filename):
                faces = self.get_faces_alignments(filename)
            else:
                faces = self.get_faces(image)
            yield filename, image, faces`

**Comment by iperov** [comment]
Windows 7, python 3.5.4
```
Using json serializer
Starting, this may take a while...
Loading Extract from Extract_Align plugin...
100%|############################################| 6/6 [00:04<00:00,  1.38it/s]
  File "_internal\bin\faceswap\faceswap.py", line 29, in <module>
    arguments.func(arguments)
  File "D:\FaceSwap\_internal\bin\faceswap\lib\cli.py", line 62, in process_argu
ments
    self.process()
  File "D:\FaceSwap\_internal\bin\faceswap\scripts\extract.py", line 58, in proc
ess
    self.write_alignments()
  File "D:\FaceSwap\_internal\bin\faceswap\lib\cli.py", line 82, in write_alignm
ents
    fn = os.path.join(self.output_dir,"alignments.{}".format(self.serializer.ext
))
  File "ntpath.py", line 113, in join
  File "genericpath.py", line 143, in _check_arg_types
TypeError: join() argument must be str or bytes, not 'WindowsPath'
```

**Comment by iperov** [comment]
also problem with your commit

problem with convert
_--converter Adjust_
or
_--converter GAN_

it produce this error FLOOD every frame:

```
OpenCV Error: Assertion failed (ssize.width > 0 && ssize.height > 0) in cv::resi
ze, file C:\projects\opencv-python\opencv\modules\imgproc\src\imgwarp.cpp, line
3483
Failed to convert image: D:\FaceSwap\data_dst\out00057.png. Reason: C:\projects\
opencv-python\opencv\modules\imgproc\src\imgwarp.cpp:3483: error: (-215) ssize.w
idth > 0 && ssize.height > 0 in function cv::resize
```

so when I remove alignments.json it works ok
explorer_2018-02-10_00-04-07

alignments.json **convert** works **only** with _--converter Masked_

can you please fix it ?

**Comment by iperov** [comment]
also you missed 
`face.image` field 
that required for ADJ and GAN

> def patch_image( self, original, face_detected ):
> #assert image.shape == (256, 256, 3)
> image = cv2.resize(**face_detected.image**, (256, 256))

**Comment by iperov** [comment]
so I fixed with

```
    def get_faces_alignments(self, filename, image):
        faces_count = 0
        faces = self.faces_detected[filename]
        for rawface in faces:
            face = DetectedFace(**rawface)
            face.image = image[face.y : face.y + face.h, face.x : face.x + face.w]
```

and

```
    def prepare_images(self):
        self.read_alignments()
        for filename in tqdm(self.read_directory()):
            image = cv2.imread(filename)
            
            if self.have_face(filename):
                faces = self.get_faces_alignments(filename, image )
```

**Comment by gdunstone** [comment]
Thanks for the input. I'll fix that stuff when I get home.

**Comment by iperov** [comment]
@gdunstone also I need to copy **alignments.json** from **extract output dir** to **convert input dir**

`for /r "data_dst\aligned\" %%x in (alignments.*) do move "%%x" "data_dst\"`

**Comment by iperov** [comment]
so with this commit I finally able to launch convert without memory error on my 6GB card.

**Comment by iperov** [comment]
found one more problem

serialized path forcibly assigned to drive and folders
**"D:\\FaceSwap\\data_src\\out00183.png": [**

so if I move whole dir, then cant convert.
it's necessary leave only filename.ext

**Comment by iperov** [comment]
@gdunstone also fix

```
    def have_face(self, filename):
        return os.path.basename(filename) in self.faces_detected
```

**Comment by iperov** [comment]
I also suggest this:

cli.py
```
 def have_alignments(self):
        fn = os.path.join(str(self.arguments.input_dir), "alignments.{}".format(self.serializer.ext))
        return os.path.exists(fn)
```

convert.py:
```
import os

. . .

            if self.have_alignments():
                if self.have_face(filename):
                    faces = self.get_faces_alignments(filename, image)
                else:
                    print ('no alignments found {}'.format(os.path.basename(filename)))
                    continue
            else:
                faces = self.get_faces(image)
```

because if alignments exist and no specific record found its cause memory error on my 6GB videocard due to  `faces = self.get_faces(image)` call

gdunstone:
are you using the cnn face detector on that, because I always get an oom error with paralellism and that cnn detector?

gdunstone:
however it is correct behaviour to skip an image that isnt in the alignments.json

iperov:
yes I always use cnn

gdunstone:
dont use the `-j` option with the cnn detector. It oom errors on my 12gb titan

iperov:
I have memory error in convert, not extract, without alignments :)

gdunstone:
Thats odd. I dont think convert is gpu accellerated. Perhaps the cnn detector isnt freeing memory correctly.
Either way I dont think that it is related to this pull request more than this pull request preventing the oom error... very strange.
Do you have an issue open?

iperov:
`Using Tensorflow` - convert not accelerated?

also accelerated due to call `faces = self.get_faces(image)` when dlib built with `DLIB_USE_CUDA`
this call produce memory error relate to cnn detector only.

gdunstone:
Tensorflow isnt by default GPU accelerated. you need to install tensorflow-gpu.

The machine I'm on now isn't GPU accelerated.

The cnn detector does have a huge memory footprint.

iperov:
I using tensorflow-gpu

gdunstone:
I still think that oom error is due to the cnn detector not freeing memory for some reason, which is outside the scope of this pull request. 

Thanks for your help with this. I'm going to merge now.

iperov:
now all fine with windows, thx

Apollo122:
I have a question; let's say during extraction we had some false positives. Before training we deleted those manually but they're still in alignment.json. So in convert won't the algorithm use those false positives because they are still in json? @gdunstone

gdunstone:
@Apollo it can because it scans the directory and checks if the file names are in alignment.json

It will do an intersection of files and entries in alignment.json

I also highly recommend using the yaml serialiser, because you can append other yamls to alignment.yaml

gdunstone:
Can't* damn autocorrect

Apollo122:
So it reads file names from data/A right?

gdunstone:
Yes

Apollo122:
@gdunstone Hmm i checked again and I'm pretty sure its not like that. Maybe I'm wrong but lets be sure. I'll try to describe it with better words:
 
1.Think a video where there is a person called Bob and a painting of Carl on the wall. We want to swap Bob's face with another and leave Carl there.
2. We used some tool to extract the frames and put those in photos/A
3. We used extract, it output Bob's images and the Carl's.
4. We deleted the Carl's images and put Bob's images in data/A. Note that Carl's data are still in alignment.json. And start training using data/A
5. We are done with the training and we want to convert the whole scene. So we using this command: convert -i ./photo/A/ -o ./output/
6. At this point photo/A contains the whole frame with Bob and Carl in it, it also has alignment.json which contains Carl's data which we don't want to swap. But it will get swapped because all data are there.

This is the what i have been asking. To use convert with original deepfakes you provide a folder called aligned. This folder contains images of swapped faces and that algorithm uses those images to merge with the original. To have the same functionally with our json we should maybe write an utility function to check filenames in data/A and delete the entries from alignment.json those that don't match.

PR #160, opened 2018-02-08
configurable loss functions + SSIM, en/decoder filename now include trainer name

IUsedToBeAPygmy:
add configurable loss functions including SSIM (Structural Image Similarity)
written filenames now include trainer name

### DIFF
--- a/lib/ModelAE.py
+++ b/lib/ModelAE.py
@@ -3,28 +3,36 @@
 import time
 import numpy
 from lib.training_data import TrainingDataGenerator, stack_images
+from keras.losses import *
 
-encoderH5 = 'encoder.h5'
-decoder_AH5 = 'decoder_A.h5'
-decoder_BH5 = 'decoder_B.h5'
 
 class ModelAE:
-    def __init__(self, model_dir):
+    def __init__(self, model_dir, filename_modifier):
+
         self.model_dir = model_dir
+        self.encoder_filename = self.model_dir + '/encoder_' + filename_modifier + '.h5'
+        self.decoder_A_filename = self.model_dir + '/decoder_A_' + filename_modifier + '.h5'
+        self.decoder_B_filename = self.model_dir + '/decoder_B_' + filename_modifier + '.h5'
 
         self.encoder = self.Encoder()
         self.decoder_A = self.Decoder()
         self.decoder_B = self.Decoder()
-
+        self.loss_function = mean_absolute_error
         self.initModel()
+        
+    #
+    # set_loss_function has to be called *before* the model is initialized!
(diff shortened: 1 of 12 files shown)


Dev-iL:
Adding a link to [the relevant issue](https://github.com/deepfakes/faceswap/issues/142)...

IUsedToBeAPygmy:
@gdunstone please review, I don't see any option to add/assign you as reviewer

gdunstone:
I'll have a review once I'm back at my pc. (havent forgotten about ya)

scripts/train.py
```diff
@@ -79,6 +79,16 @@ def parse_arguments(self, description, subparser, command):
                             type=int,
                             default=64,
                             help="Batch size, as a power of 2 (64, 128, 256, etc)")
+        parser.add_argument('-l', '--loss-function',
```
gdunstone: should the selection of loss function change the default  loss_mix?

plugins/LossFunction_l1.py
```diff
@@ -0,0 +1,11 @@
+from keras import losses
+
+# L1 loss function = Mean of Absolute Error
+
```
gdunstone: Is there any benefit here to having this over just assigning LossFunction to the loss function signature from keras?

LossFunction = losses.mean_absolute_error

lib/ModelAE.py
```diff
@@ -3,28 +3,36 @@
...
         self.decoder_A = self.Decoder()
         self.decoder_B = self.Decoder()
-
+        self.loss_function = mean_absolute_error
         self.initModel()
+        
+    #
+    # set_loss_function has to be called *before* the model is initialized!
+    #
+    def set_loss_function(self, loss_function):
```
gdunstone: whats the reasoning behind this function over passing the loss function to __init__?

scripts/train.py
IUsedToBeAPygmy: Currently it doesn't need to. The loss mix is currently only used with (both) the SSIM loss functions, and is set by default to a very low value hardly influencing results - experimentation with this value is needed to see whether SSIM actually improves results or is completely unnecessary.
This was purely added to make experimentation with the Keras.Contrib SSIM implementation possible - I have not had the time to actually test the results.

plugins/LossFunction_l1.py
IUsedToBeAPygmy: Wow. After reviewing this code I found it actually doesn't work - it will always use the default loss function as it's not being overwritten in time.

Hold on while I try to figure out how to undo this pull request and fix my code :)

PR #163, opened 2018-02-09
Update Docker file using ustc mirror for cn users

riclava:


Clorr:
i'm not sure about what this does, can you explain a bit more? Can someone review this?

zhujunsan:
@Clorr ustc mirror is within china's firewall thus faster than the deb original mirror for CN users
@riclava but you are slowing down for other users : (

Clorr:
Ok, thanks @zhujunsan 

I close this as it is too specific, but hanks for the PR

Issue #153, opened 2018-02-08
Outputting training graphs


IUsedToBeAPygmy:
Once I'm done with my current task I'd like to add a feature to output a graph of the training performance.

- Add graph depicting epoch vs loss
- add max-epoch parameter to be able to run for a specified amount of epochs

Please assign this to me.
I'd like to hear more ideas on this subject.

GSonderling:
Both seem simple enough. The epochs are already limited in the script. It would only be necessary to change constant to a variable. 

As for the graph, that would require a graphics output, possibly in separate window. Again, not exactly hard by itself. But since there is still no proper GUI we should be careful not to reinvent the wheel and make features that will be quickly obsolete.

Of course there are ways around that issue. For example text output to a csv file. That can be later used for creation of graph. Another option might be bare bones graphics in form of a bitmap.

But the real question is if the graph should be real time or made after the run of program.

IUsedToBeAPygmy:
I was just thinking of writing the graph to a jpg at the same time the preview is shown/saved.
Nice and easy.

GSonderling:
Well that wouldn't be too hard.

IUsedToBeAPygmy:
I went with outputting SVG using Pygal as that has the least amount of dependencies and SVG is a universal format (can be opened with any webbrowser).

Ill create a pull for it soon.

gdunstone:
I am expressing interest in another preview window with a graph.

dfaker:
@subzerofun If you want to use Tensorboard it'd might make more sense to use it directly on a merged model.

    i1 = Input( shape=IMAGE_SHAPE )
    i2 = Input( shape=IMAGE_SHAPE )
    tbcallback = TensorBoard(log_path)
    mergedModel = Model( [x1,x2] decoder_A( encoder(x1) ), decoder_B( encoder(x2) ) )
    model.fit(x=[warped_A,warped_B], y=[target_A,target_B], callbacks=[tbcallback])

fat-tire:
How about logging periodically so tensorboard can show all kinds of info beyond loss?

torzdf:
I'm going to add a fairly basic loss graph to the GUI. Of course tensorboard will be able to do a lot more than I'll be able to do, but at least it will be something.

torzdf:
#345

torzdf:
Graphing added with PR #352 . Please open a new issue if looking to extend functionality.

PR #177, opened 2018-02-11
New epoch limiting arg.

GSonderling:
In response to second point in #153.

The epoch argument allows user to limit runtime in a simple and user-friendly fashion.

Name and description of arg are subject to adjustment, if you think they should be different I will change them.

### DIFF
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -79,6 +79,10 @@ class TrainingProcessor(object):
                             type=int,
                             default=64,
                             help="Batch size, as a power of 2 (64, 128, 256, etc)")
+        parser.add_argument('-ep', '--epochs',
+                            type=int,
+                            default=1000000,
+                            help="Length of training in epochs.")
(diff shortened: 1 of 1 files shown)


iperov:
where is save after loop end ?

GSonderling:
I didn't alter anything after the loop. This is the original version:

https://github.com/deepfakes/faceswap/blob/20753a64b76a156aea17724348269d60dd525f87/scripts/train.py#L136-L151


As you can see that part is unchanged. The save behavior is same as in original. I want to change that too, but it should have its own PR.

iperov:
so why you change num of epochs if you cant understand behaviour after loop end ?

GSonderling:
I understand that behavior. I just don't like it. If you don't like it either, fine, I want to change it later. But standard approach is one PR per feature. I can put more in this one, but it's going to add more complexity.

PR #178, opened 2018-02-11
Update INSTALL.md

GSonderling:
INSTALL.md now aligns with requirements specified in the script.

It should hopefully prevent some confusion among less adept users and simplyfy installation process.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -39,7 +39,7 @@ The developers are also not responsible for any damage you might cause to your o
 
 ## Installing dependencies
 
-### Python 3.6
+### Python >= 3.2
 
 Note that you will need the 64bit version of Python, especially to setup the GPU version!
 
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -49,9 +49,9 @@ Download the latest version of Python 3 from Python.org: https://www.python.org/
 
 #### macOS
 
-By default, macOS comes with Python 2.7. For best usage, need Python 3.6. The easiest way to do so is to install it through `Homebrew`. If you are not familiar with `homebrew`, read more about it here: https://brew.sh/
+By default, macOS comes with Python 2.7. For best usage, need at least Python 3.2.  The easiest way to do so is to install it through `Homebrew`. If you are not familiar with `homebrew`, read more about it here: https://brew.sh/
 
-To install Python 3.6:
+To install Python 3.2:
```
Clorr: I'm ok with the 3.2 above, but I find this line confusing, because one would think it is recommended to install 3.2 and may be confused if brew installs 3.6, don't you tink?

GSonderling: How about:

> To install Python >= 3.2:


That would fit for all compatible versions of Python.

GSonderling: I decided for slightly different wording, but to a same effect.

Clorr:
Ok, let's go like this

PR #181, opened 2018-02-11
lowercase mask_type

dfaker:
The mask_type passed in to this function is lowercase, changing string literals to match.


Clorr:
Hi, and thank you for this pull request ;-)

May I suggest that you make this even more robust by using something like `self.mask_type.lower()` where relevant? This way, we don't have any more problem with that ;-)

Clorr:
Oh sorry, it's done already ;-)

Issue #161, opened 2018-02-08
Cuda 9, CuDNN 7, TensorFlow 1.5 seem to work fine. any reason not to use the latest versions?


IUsedToBeAPygmy:
I'm working on Windows 10.
I've installed Cuda 9, CuDNN 7 and TensorFlow 1.5, and everything has been working fine for a week now.
I think it's a bit faster, but I have not measured it.
Is there any specific reason to linger with Cuda 8 and TF 1.4?

gdunstone:
I think we should use something like `tensorflow-gpu>=1.4.0`

facepainter:
@IUsedToBeAPygmy 

I too have a set up with CUDA 9, CuDNN 7 and TF 1.5 using Python 3.6 on Win10 64bit and it works great...execpt I can't build dlib with CUDA support against version 9 :( 

Have you had joy with this? This is all that is keeping me on v8/6 and TF 1.4...

LordVulkan:
I have been using CUDA 9.1, cuDNN 7 and a custom build from TensorFlow master branch without trouble.

Clorr:
I have a commit ready for that:

https://github.com/deepfakes/faceswap/commit/c1ce49ec6277ff20b8e85c30ff3a7fa3d2a90e4a

You can merge it if it's ok for you

shadowfolder:
@facepainter i was unable install dlib via pip too. But i sucessfully installed dlib via github https://github.com/davisking/dlib via python setup.py install

dr-benway:
works great on > CUDA9 CuDNN7 TF 1.6.0rc0 Python 3.6.2  Win10 x64

gdunstone:
Not everyone runs Archlinux.

dr-benway:
works great on > CUDA9 CuDNN7 TF 1.6.0rc0 Python 3.6.2  Win10 x64
training with GAN128 (increased kernel sizes to 4 in order for them to be divisible by the strides)

dr-benway:
@facepainter @shadowfolder 
check this out for a compiled wheel > http://www.xavierdupre.fr/app/pymyinstall/helpsphinx/blog/2017/2017-07-10_dlib.html

facepainter:
@dr-benway @shadowfolder - hmm yeah I can download/install and compile it fine...just not with CUDA support.

LordVulkan:
> facepainter commented 18 minutes ago
[...]

PR #183, opened 2018-02-11
Clearer requirements for each platform

Clorr:
As it has been upvoted in #161 , I'm PRing this now. Please make me some feedback before I merge it


LordVulkan:
Something to have in mind. There is a bug in current versions of CUDA 9.0 and 9.1.

[See TensorFlow 1.6.0-rc0 release](https://github.com/tensorflow/tensorflow/releases/tag/v1.6.0-rc0)

> # Known Bugs
> Using XLA:GPU with CUDA 9 and CUDA 9.1 results in garbage results and/or
> CUDA_ILLEGAL_ADDRESS failures.
> 
> Google discovered in mid-December 2017 that the PTX-to-SASS compiler in CUDA 9
> and CUDA 9.1 sometimes does not properly compute the carry bit when
> decomposing 64-bit address calculations with large offsets (e.g. load [x + large_constant]) into 32-bit arithmetic in SASS.
> 
> As a result, these versions of ptxas miscompile most XLA programs which use
> more than 4GB of temp memory. This results in garbage results and/or
> CUDA_ERROR_ILLEGAL_ADDRESS failures.
> 
> A fix in CUDA 9.1.121 is expected in late February 2018. We do not expect a
> fix for CUDA 9.0.x. Until the fix is available, the only workaround is to
> downgrade to CUDA 8.0.x
> or disable XLA:GPU.
> 
> TensorFlow will print a warning if you use XLA:GPU with a known-bad version of
> CUDA; see e00ba24.

As they say a fix for CUDA 9.0 is not expected, maybe the official requirements should stay at CUDA 8.0 until the fix for CUDA 9.1 and TensorFlow 1.6.0 releases later this month, and skip CUDA 9.0. I don't know if the bug affects this project.

PR #194, opened 2018-02-14
measure and output min,max,avg, last train timing

avilv:


Clorr:
Brings too few improvments. However if you have  solution for plotting and making code cleaner, it would be welcome

PR #195, opened 2018-02-14
color match A and B set images + performance improvement

avilv:
improve performance by pre-loading images before iteration.

i noticed FakeApp does this better by matching the colors between the sets so i added that back in
also avoiding cv2.readim on each iteration is a much better idea than doing that repeatedly
 (HDD's are slow)
i'm seeing a nice performance improvement from this as-well

### DIFF
--- a/lib/ModelAE.py
+++ b/lib/ModelAE.py
@@ -51,12 +51,10 @@ class TrainerAE():
         self.model = model
 
         generator = TrainingDataGenerator(self.random_transform_args, 160)
-        self.images_A = generator.minibatchAB(fn_A, self.batch_size)
-        self.images_B = generator.minibatchAB(fn_B, self.batch_size)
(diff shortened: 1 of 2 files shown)


LordVulkan:
What if you don't have enough memory to allocate all images?

subzerofun:
@avilv  With how many images in folder A and B combined did you test this?
Do the images now get stored in RAM or on the GPU? I currently have a project where folder A has ~4000 face images, folder B also ~4000. I'm a little reluctant to test this without knowing if my hardware can handle pre-loading 8000 images :-)

subzerofun:
OK 8000 images _are_ too much. Memory (24 GB) full, swap files piling up. Terminal freezes, several Apps crash – but the Python process with tensorflow is still running, allocating space 😆.

Clorr:
Has pros and cons, tha fact it loads all files at the beginning is a problem, but you are right, we still can do better for that part...

PR #199, opened 2018-02-15
Add debugging option for drawing landmarks on face extraction.

LordVulkan:


iperov:
more logical place debug argument at end of list
`def extract(self, image, face, size, debug_landmarks = False):`

LordVulkan:
False as default value is redundant as it is already specified in the arguments parser.

iperov:
Think as modular programming, is_debug_landmarks actually optional value and doesnt affect to `extract` major behaviour

LordVulkan:
I have updated the base code from master and moved the drawing code to extract script, but I don't know if I did the merge correctly. Please checkout.

PR #214, opened 2018-02-20
Skip already extracted frames when using extract.py

ppmdo:
Added a function to verify if extracted faces already exists prior to generating the input list. This ensures no images get reprocessed, and reduces the size of the input for processing.


oatssss:
It doesn't look like you added logic to merge the previous alignments file with the new incoming one. Does this clobber the alignments for the images that were already extracted?

ppmdo:
You're right. Let me fix that.

ppmdo:
It's done bro. @oatssss

Clorr:
@oatssss are you ok for merging this?

oatssss:
I haven't had a chance to test it yet.

lib/cli.py
```diff
@@ -82,6 +95,12 @@ def write_alignments(self):
         fn = os.path.join(str(self.arguments.input_dir), "alignments.{}".format(self.serializer.ext))
         if self.arguments.alignments_path is not None:
             fn = self.arguments.alignments_path
+        print("Alignments filepath: %s" % fn)
+        if os.path.exists(fn):
+            with open(fn, self.serializer.roptions) as inf:
+                data = self.serializer.unmarshal(inf.read())
+                for k, v in data.items():
+                    self.faces_detected[k] = v
```
oatssss: Thanks for your work :) I had a chance to test out the PR, this is what I found:

If I don't switch on the `--skip-existing` flag, and there's an old alignments file, it still gets merged with the new one. I suggest not merging with the old alignments file if `--skip-existing` isn't enabled, but if you wanna don't wanna go down that path and instead want to always merge there's a problem here:

The order you have it is:
1. Extract alignments into `self.faces_detected`
2. Use previous alignments file to overwrite any entries already in `self.faces_detected`
3. Write out merged alignments

I believe it should be the other way around:
1. Use previous alignments file to fill `self.faces_detected` during initialization
2. Extract alignments into `self.faces_detected`, overwriting any old alignments with the new extracted ones.
3. Write out merged alignments

This will ensure that the new alignments are kept.

lib/utils.py
```diff
@@ -11,9 +12,20 @@ def get_folder(path):
...
 
-def get_image_paths(directory):
-    return [x.path for x in scandir(directory) if
-     any(map(lambda ext: x.name.lower().endswith(ext), image_extensions))]
+def get_image_paths(directory, exclude=[], debug=False):
+    exclude_names = [basename(Path(x).stem[:-1] + Path(x).suffix) for x in exclude]
+    dir_contents = []
+    dir_scanned = list(scandir(directory))
+    for x in dir_scanned:
+        if any(map(lambda ext: x.name.lower().endswith(ext), image_extensions)):
```
oatssss: Personally, I think this is easier to read:
``` python
if any([x.name.lower().endswith(ext) for ext in image_extensions]):
```

lib/cli.py
ppmdo: Totally missed deactivating the alignments merging when not using skip-existing. Fixing asap.

lib/utils.py
ppmdo: Me too. That was the original syntax. Changed now.

oatssss:
If I specify `--skip-existing` and also an `--output-dir` that doesn't exist yet, I get:

```
File "D:\machine-learning\faceswap\lib\utils.py", line 18, in get_image_paths
    dir_scanned = list(scandir(directory))
FileNotFoundError: [WinError 3] The system cannot find the path specified: "/the/output/dir/specified"
```

Previous behaviour would just create it. Could you implement that as well?

ppmdo:
Implemented :smile:

oatssss:
Thanks for your work! Works great on my end :)

PR #216, opened 2018-02-21
Refactoring scripts/extract.py

leftler:
Refactored extract to have `handleImage` take in a image instead of a filename. This will be useful for writing future extensions for extract where you may not have a file on disk but you do have a image in memory. For example extracting faces from a video directly.


Clorr:
Seems fine, thanks for the PR

Issue #200, opened 2018-02-16
Encoder weights stored twice in GAN model


leijurv:
In the original model, encoder decoder_A and decoder_B have their weights stored separately.
https://github.com/deepfakes/faceswap/blob/master/lib/ModelAE.py#L24

```
try:
            self.encoder.load_weights(str(self.model_dir / encoderH5))
            self.decoder_A.load_weights(str(self.model_dir / face_A))
            self.decoder_B.load_weights(str(self.model_dir / face_B))
            print('loaded model weights')
            return True
```

In the new GAN model, weights are stored as a combination in two files, one with encoder and decoder_A, and another with the same encoder and decoder_B. 

https://github.com/deepfakes/faceswap/blob/master/plugins/Model_GAN/Model.py#L126

```
        encoder = Encoder(self.img_shape)
        decoder_A = Decoder_ps(self.img_shape)
        decoder_B = Decoder_ps(self.img_shape)    
        x = Input(shape=self.img_shape)
        netGA = Model(x, decoder_A(encoder(x)))
        netGB = Model(x, decoder_B(encoder(x)))           
        try:
            netGA.load_weights(str(self.model_dir / netGAH5))
            netGB.load_weights(str(self.model_dir / netGBH5))
            print ("Generator models loaded.")
        except:
            print ("Generator weights files not found.")
            pass
        return netGA, netGB, 
```

leijurv:
Is there a reason for this change? Since the encoder is the same between the two models, what is the purpose of storing it in two different places on disk?

shaoanlu:
To prevent overwriting encoder weights of non-GAN models.

PR #217, opened 2018-02-22
Update GAN64 to v2

oatssss:
Based from Clorr's fork after the GAN128 draft.

The GAN64 model is still mostly based off [shaoanlu's v1](https://github.com/shaoanlu/faceswap-GAN/blob/0260702077261b77242217d04734426851f7ce42/temp/faceswap_GAN_keras.ipynb).

The mask results I was getting were unusable; I'd have masks that were basically all transparent. I updated GAN64 to [shaoanlu's v2](https://github.com/shaoanlu/faceswap-GAN/blob/0260702077261b77242217d04734426851f7ce42/FaceSwap_GAN_v2_train.ipynb).

With these changes, my mask results are better. It could be that only I was having this problem so please test yourself!


babilio:
Awesome! yeah GAN here was unusable for me too. Will be trying your changes this weekend. Thanks

oatssss:
The results I get with shaoanlu's original v2 are still a lot better than what I'm getting with this. I wonder why.

Apollo122:
Thank you for your contribution @oatssss. Is this the full port of v2 or just a partial one? GAN here still not great yeah, If you could check the existing code for wrong implementation

oatssss:
I compared the Model and Trainer to shaoanlu's v2 and made sure everything's the same. I think it should work better now.

Apollo122:
Awesome! Will give it a go this weekend.Let you know the results.
If you have the motivation you can work on gan128 model. Mind that it's converter doesn't even exist in this repo so would be great if you can do it as well. Thanks again

Apollo122:
If you do it please check the existing gan64 converter here as well and make sure all is in order. There maybe some bug that prevents us to get the same results

oatssss:
I think I know how to get the GAN128 converter working, but I don't think I'll be able to test it because I don't have enough vram.

Apollo122:
Oh if that's the case I can test it. There are other users here as well. just do the convert, then we can provide weights files.

plugins/Model_GAN128/Trainer.py
```diff
@@ -74,6 +75,12 @@ def setup(self):
         loss_GA += 3e-3 * K.mean(K.abs(mask_A))
         loss_GB += 3e-3 * K.mean(K.abs(mask_B))
 
+        w_fo = 0.01
+        loss_GA += w_fo * K.mean(first_order(mask_A, axis=1))
+        loss_GA += w_fo * K.mean(first_order(mask_A, axis=2))
+        loss_GB += w_fo * K.mean(first_order(mask_B, axis=1))
+        loss_GB += w_fo * K.mean(first_order(mask_B, axis=2))
```
oatssss: This is the main difference between Clorr's current gan128 implementation and shaoanlu's latest gan128. shaoanlu added it in the commit https://github.com/shaoanlu/faceswap-GAN/commit/ea6b27b5e665748285de34a52aed767459740674

Apollo122:
Alright. I'll start the training then. I think 5-6 hours should be enough. Will share weights here tomorrow.
If you can port 128-converter until tomorrow we can test it together.

Apollo122:
@oatssss 
got this crash

  File "C:\IntelPython3\lib\threading.py", line 916, in _bootstrap_inner
    self.run()
  File "C:\IntelPython3\lib\threading.py", line 864, in run
  File "K:\faceswap-update-GAN-v2\scripts\train.py", line 136, in processThread
    trainer = trainer(model, images_A, images_B, batch_size=self.arguments.batch_size)
  File "K:\faceswap-update-GAN-v2\plugins\Model_GAN128\Trainer.py", line 44, in __init__
    self.setup()
  File "K:\faceswap-update-GAN-v2\plugins\Model_GAN128\Trainer.py", line 79, in setup
    loss_GA += w_fo * K.mean(first_order(mask_A, axis=1))
NameError: name 'first_order' is not defined

oatssss:
Should be fixed now

Apollo122:
crashed again. but i fixed it. you should change these lines as well. "self" was missing.

        loss_GA += w_fo * K.mean(self.first_order(mask_A, axis=1))
        loss_GA += w_fo * K.mean(self.first_order(mask_A, axis=2))
        loss_GB += w_fo * K.mean(self.first_order(mask_B, axis=1))
        loss_GB += w_fo * K.mean(self.first_order(mask_B, axis=2))

Apollo122:
now new crash. a little long crash log :)

Exception in thread Thread-1:
  File "C:\IntelPython3\lib\site-packages\tensorflow\python\client\session.py", line 1350, in _do_call
    return fn(*args)
  File "C:\IntelPython3\lib\site-packages\tensorflow\python\client\session.py", line 1329, in _run_fn
    status, run_metadata)
  File "C:\IntelPython3\lib\site-packages\tensorflow\python\framework\errors_impl.py", line 473, in __exit__
    c_api.TF_GetCode(self.status.status))
tensorflow.python.framework.errors_impl.InvalidArgumentError: You must feed a value for placeholder tensor 'conv1/7x7_s2/bn/keras_learning_phase' with dtype bool
         [[Node: conv1/7x7_s2/bn/keras_learning_phase = Placeholder[dtype=DT_BOOL, shape=<unknown>, _device="/job:localhost/replica:0/task:0/device:CPU:0"]()]]

During handling of the above exception, another exception occurred:

  File "C:\IntelPython3\lib\threading.py", line 916, in _bootstrap_inner
    self.run()
  File "C:\IntelPython3\lib\threading.py", line 864, in run
  File "K:\faceswap-update-GAN-v2\scripts\train.py", line 144, in processThread
    trainer.train_one_step(epoch, self.show if (save_iteration or self.save_now) else None)
  File "K:\faceswap-update-GAN-v2\plugins\Model_GAN128\Trainer.py", line 125, in train_one_step
    errGA = self.netGA_train([warped_A, target_A])
  File "C:\IntelPython3\lib\site-packages\keras\backend\tensorflow_backend.py", line 2357, in __call__
    **self.session_kwargs)
  File "C:\IntelPython3\lib\site-packages\tensorflow\python\client\session.py", line 895, in run
    run_metadata_ptr)
  File "C:\IntelPython3\lib\site-packages\tensorflow\python\client\session.py", line 1128, in _run
    feed_dict_tensor, options, run_metadata)
  File "C:\IntelPython3\lib\site-packages\tensorflow\python\client\session.py", line 1344, in _do_run
    options, run_metadata)
  File "C:\IntelPython3\lib\site-packages\tensorflow\python\client\session.py", line 1363, in _do_call
    raise type(e)(node_def, op, message)
tensorflow.python.framework.errors_impl.InvalidArgumentError: You must feed a value for placeholder tensor 'conv1/7x7_s2/bn/keras_learning_phase' with dtype bool
         [[Node: conv1/7x7_s2/bn/keras_learning_phase = Placeholder[dtype=DT_BOOL, shape=<unknown>, _device="/job:localhost/replica:0/task:0/device:CPU:0"]()]]

Caused by op 'conv1/7x7_s2/bn/keras_learning_phase', defined at:
  File "C:\IntelPython3\lib\threading.py", line 884, in _bootstrap
    self._bootstrap_inner()
  File "C:\IntelPython3\lib\threading.py", line 916, in _bootstrap_inner
    self.run()
  File "C:\IntelPython3\lib\threading.py", line 864, in run
  File "K:\faceswap-update-GAN-v2\scripts\train.py", line 136, in processThread
    trainer = trainer(model, images_A, images_B, batch_size=self.arguments.batch_size)
  File "K:\faceswap-update-GAN-v2\plugins\Model_GAN128\Trainer.py", line 44, in __init__
    self.setup()
  File "K:\faceswap-update-GAN-v2\plugins\Model_GAN128\Trainer.py", line 61, in setup
    vggface = VGGFace(include_top=False, model='resnet50', input_shape=(224, 224, 3))
  File "C:\IntelPython3\lib\site-packages\keras_vggface\vggface.py", line 97, in VGGFace
    classes=classes)
  File "C:\IntelPython3\lib\site-packages\keras_vggface\models.py", line 233, in RESNET50
    x = BatchNormalization(axis=bn_axis, name='conv1/7x7_s2/bn')(x)
  File "C:\IntelPython3\lib\site-packages\keras\engine\topology.py", line 603, in __call__
    output = self.call(inputs, **kwargs)
  File "C:\IntelPython3\lib\site-packages\keras\layers\normalization.py", line 190, in call
    training=training)
  File "C:\IntelPython3\lib\site-packages\keras\backend\tensorflow_backend.py", line 2740, in in_train_phase
    training = learning_phase()
  File "C:\IntelPython3\lib\site-packages\keras\backend\tensorflow_backend.py", line 121, in learning_phase
    name='keras_learning_phase')
  File "C:\IntelPython3\lib\site-packages\tensorflow\python\ops\array_ops.py", line 1680, in placeholder
    return gen_array_ops._placeholder(dtype=dtype, shape=shape, name=name)
  File "C:\IntelPython3\lib\site-packages\tensorflow\python\ops\gen_array_ops.py", line 4105, in _placeholder
    "Placeholder", dtype=dtype, shape=shape, name=name)
  File "C:\IntelPython3\lib\site-packages\tensorflow\python\framework\op_def_library.py", line 787, in _apply_op_helper
    op_def=op_def)
  File "C:\IntelPython3\lib\site-packages\tensorflow\python\framework\ops.py", line 3160, in create_op
    op_def=op_def)
  File "C:\IntelPython3\lib\site-packages\tensorflow\python\framework\ops.py", line 1625, in __init__
    self._traceback = self._graph._extract_stack()  # pylint: disable=protected-access

InvalidArgumentError (see above for traceback): You must feed a value for placeholder tensor 'conv1/7x7_s2/bn/keras_learning_phase' with dtype bool
         [[Node: conv1/7x7_s2/bn/keras_learning_phase = Placeholder[dtype=DT_BOOL, shape=<unknown>, _device="/job:localhost/replica:0/task:0/device:CPU:0"]()]]

oatssss:
Try that. I just disabled perceptual loss.

oatssss:
For your first crash with the `self` missing, where was it missing?

Apollo122:
shared the correct code below.

now its working but PL important. do you have any idea why it doesnt work?

oatssss:
That stack trace makes it seem like it's an error with the `keras_vggface` package. No clue.

The code I'm looking at looks like it already has the `self`.

Apollo122:
nope check the lines 79,80 ...

https://github.com/oatssss/faceswap/blob/update-GAN-v2/plugins/Model_GAN128/Trainer.py

oatssss:
lmao, I was looking at gan64, thanks.

Also, from shaoanlu's jupyter notebook, there's a comment about it:

>If you got error _obtain_input_shape(...) error, this is because your keras version is older than vggface requirement.
>Modify _obtain_input_shape(...) in keras_vggface/models.py will solve the problem. The following is 
    what worked for me:

``` python
input_shape = _obtain_input_shape(input_shape,
                                  default_size=224,
                                  min_size=197,
                                  data_format=K.image_data_format(),
                                  include_top=include_top)
```

oatssss:
I left a comment in the gan128 draft PR about trying to fix the converter. You could give it a shot too.

Apollo122:
pip list says i have Keras (2.1.2) and keras-vggface (0.5) so dont i have old version. and i dont see obtain_input_shape in the crash log hmm

Apollo122:
Hey @shaoanlu we got this crash when we enabled PL. 
this seem to be main issue: "tensorflow.python.framework.errors_impl.InvalidArgumentError: You must feed a value for placeholder tensor 'conv1/7x7_s2/bn/keras_learning_phase' with dtype bool" 

do you have any idea?

Exception in thread Thread-1:
File "C:\IntelPython3\lib\site-packages\tensorflow\python\client\session.py", line 1350, in _do_call
return fn(*args)
File "C:\IntelPython3\lib\site-packages\tensorflow\python\client\session.py", line 1329, in _run_fn
status, run_metadata)
File "C:\IntelPython3\lib\site-packages\tensorflow\python\framework\errors_impl.py", line 473, in exit
c_api.TF_GetCode(self.status.status))
tensorflow.python.framework.errors_impl.InvalidArgumentError: You must feed a value for placeholder tensor 'conv1/7x7_s2/bn/keras_learning_phase' with dtype bool
[[Node: conv1/7x7_s2/bn/keras_learning_phase = Placeholderdtype=DT_BOOL, shape=, _device="/job:localhost/replica:0/task:0/device:CPU:0"]]
During handling of the above exception, another exception occurred:
File "C:\IntelPython3\lib\threading.py", line 916, in _bootstrap_inner
self.run()
File "C:\IntelPython3\lib\threading.py", line 864, in run
File "K:\faceswap-update-GAN-v2\scripts\train.py", line 144, in processThread
trainer.train_one_step(epoch, self.show if (save_iteration or self.save_now) else None)
File "K:\faceswap-update-GAN-v2\plugins\Model_GAN128\Trainer.py", line 125, in train_one_step
errGA = self.netGA_train([warped_A, target_A])
File "C:\IntelPython3\lib\site-packages\keras\backend\tensorflow_backend.py", line 2357, in call
**self.session_kwargs)
File "C:\IntelPython3\lib\site-packages\tensorflow\python\client\session.py", line 895, in run
run_metadata_ptr)
File "C:\IntelPython3\lib\site-packages\tensorflow\python\client\session.py", line 1128, in _run
feed_dict_tensor, options, run_metadata)
File "C:\IntelPython3\lib\site-packages\tensorflow\python\client\session.py", line 1344, in _do_run
options, run_metadata)
File "C:\IntelPython3\lib\site-packages\tensorflow\python\client\session.py", line 1363, in _do_call
raise type(e)(node_def, op, message)
tensorflow.python.framework.errors_impl.InvalidArgumentError: You must feed a value for placeholder tensor 'conv1/7x7_s2/bn/keras_learning_phase' with dtype bool
[[Node: conv1/7x7_s2/bn/keras_learning_phase = Placeholderdtype=DT_BOOL, shape=, _device="/job:localhost/replica:0/task:0/device:CPU:0"]]
Caused by op 'conv1/7x7_s2/bn/keras_learning_phase', defined at:
File "C:\IntelPython3\lib\threading.py", line 884, in _bootstrap
self._bootstrap_inner()
File "C:\IntelPython3\lib\threading.py", line 916, in _bootstrap_inner
self.run()
File "C:\IntelPython3\lib\threading.py", line 864, in run
File "K:\faceswap-update-GAN-v2\scripts\train.py", line 136, in processThread
trainer = trainer(model, images_A, images_B, batch_size=self.arguments.batch_size)
File "K:\faceswap-update-GAN-v2\plugins\Model_GAN128\Trainer.py", line 44, in init
self.setup()
File "K:\faceswap-update-GAN-v2\plugins\Model_GAN128\Trainer.py", line 61, in setup
vggface = VGGFace(include_top=False, model='resnet50', input_shape=(224, 224, 3))
File "C:\IntelPython3\lib\site-packages\keras_vggface\vggface.py", line 97, in VGGFace
classes=classes)
File "C:\IntelPython3\lib\site-packages\keras_vggface\models.py", line 233, in RESNET50
x = BatchNormalization(axis=bn_axis, name='conv1/7x7_s2/bn')(x)
File "C:\IntelPython3\lib\site-packages\keras\engine\topology.py", line 603, in call
output = self.call(inputs, **kwargs)
File "C:\IntelPython3\lib\site-packages\keras\layers\normalization.py", line 190, in call
training=training)
File "C:\IntelPython3\lib\site-packages\keras\backend\tensorflow_backend.py", line 2740, in in_train_phase
training = learning_phase()
File "C:\IntelPython3\lib\site-packages\keras\backend\tensorflow_backend.py", line 121, in learning_phase
name='keras_learning_phase')
File "C:\IntelPython3\lib\site-packages\tensorflow\python\ops\array_ops.py", line 1680, in placeholder
return gen_array_ops._placeholder(dtype=dtype, shape=shape, name=name)
File "C:\IntelPython3\lib\site-packages\tensorflow\python\ops\gen_array_ops.py", line 4105, in _placeholder
"Placeholder", dtype=dtype, shape=shape, name=name)
File "C:\IntelPython3\lib\site-packages\tensorflow\python\framework\op_def_library.py", line 787, in _apply_op_helper
op_def=op_def)
File "C:\IntelPython3\lib\site-packages\tensorflow\python\framework\ops.py", line 3160, in create_op
op_def=op_def)
File "C:\IntelPython3\lib\site-packages\tensorflow\python\framework\ops.py", line 1625, in init
self._traceback = self._graph._extract_stack() # pylint: disable=protected-access
InvalidArgumentError (see above for traceback): You must feed a value for placeholder tensor 'conv1/7x7_s2/bn/keras_learning_phase' with dtype bool
[[Node: conv1/7x7_s2/bn/keras_learning_phase = Placeholderdtype=DT_BOOL, shape=, _device="/job:localhost/replica:0/task:0/device:CPU:0"]]

Apollo122:
btw for gan-128 one epoch takes 44 secs on gtx1080. not sure if it was this slow

oatssss:
So this is after I updated gan128 to shaoanlu's latest? (commit 2f38bf7)

Apollo122:
didnt try his version but it takes 3-4 secs for gan-64 here

Apollo122:
wait i understand. it doesnt use GPU at all

Apollo122:
ok solved it. for whatever reason pip has installed regular tensorflow along with keras and other libraries. uninstalled both tensorflow packages and reinstalled tensorflow-gpu solved the issue.

Apollo122:
What was the importance of 
if iter % 1 == 0:   part?

It seems it got removed as well

oatssss:
The `%` operator calculates the remainder. It's a redundant line since `iter` is always an integer and every integer is divisible by 1. i.e. `iter % 1` is always equal to 0.

I imagine shaoanlu had it as something other than 1 at some point so the output would only display every few iterations instead of with every iteration (like it is now).

Apollo122:
Isn't 100 iterations too much? I think 10 would be ideal

oatssss:
I think the default save interval is 100, so I just used that. Later it can be coded to use whatever was set for the actual save interval.

Apollo122:
Yeah okay. Btw I started the training I will share it here once it's okay quality. 

I have gan64 model as well. do you think that it can be use with your PR? Or should I have to start all over?

oatssss:
Do you know from which commit your gan64 model is from?

oatssss:
My loss values are basically stagnant around ~0.17 for DA/B and ~0.22 for GA/B.

Apollo122:
People who want test this beware: keras-vggface installs cpu based tensorflow as dependency. Won't give you any error but slow your training speed by 40. You should uninstall it and install gpu based tf. Read this comment
https://github.com/shaoanlu/faceswap-GAN/issues/17#issuecomment-367791033

Apollo122:
@oatssss  im still training gan128.  after 14k iterations output is like this. Especially masked output is hilarious :) guess it needs more time.

## **Raw:**
raw

**## Mask:
mask
**
## **Masked:**
masked

oatssss:
Your results seem better than what I'm getting with the gan64. Are your loss values continually decreasing?

Apollo122:
@oatssss 
didnt pay attention. maybe you can add the change for every 100 ite also. btw here is the link for gan128 model. i trained it for 25k iterations still not great but should be enough for you to start convert part. https://mega.nz/#!sp5E1aZD!1pP1Bi3rewmCcU2HvY-FrksXsIdQGFG5p0HaImJr4Ls

Apollo122:
Also I don't remember getting warped masked output with gan64 version. Another user raised this issue as well

oatssss:
What's warped masked output? Which issue is this?

oatssss:
@Apollo122  I haven't tested it, but try this branch https://github.com/oatssss/faceswap/tree/update-GAN-v2-tests for gan128 conversion.

Apollo122:
Will check it. As for the issue you can see it on my previous comment. Check the image under masked. The most obvious one is at right side 3rd position from bottom

Apollo122:
I don't know if it has something to do with v2. But with clorr's gan64 I didn't have this issue

oatssss:
gan128 conversion was working for me so I just pushed the commits to this PR too.

plugins/Model_GAN/Trainer.py
```diff
@@ -24,130 +29,216 @@ def __init__(self, model, fn_A, fn_B, batch_size):
...
-        figure_B = np.stack([
-            test_B,
-            out_test_B_netGB[1],
-            out_test_B_netGA[1],
-            ], axis=1 )
-        
-        display_fig("raw", figure_A, figure_B)       
-
+        display_fn(self.showG(tA, tB, self.path_A, self.path_B), "raw")
+        display_fn(self.showG(wA, wB, self.path_bgr_A, self.path_bgr_B), "masked")
```
oatssss: I believe this is the reason there's warped masked output. Not sure why @Clorr used the warped version here (`wA` instead of `tA`)

plugins/Model_GAN/Trainer.py
```diff
@@ -23,131 +28,230 @@ def __init__(self, model, fn_A, fn_B, batch_size):
...
-            ], axis=1 )
-        figure_B = np.stack([
-            test_B,
-            out_test_B_netGB[1],
-            out_test_B_netGA[1],
-            ], axis=1 )
-        
-        display_fig("raw", figure_A, figure_B)       
+        display_fn(self.showG(tA, tB, self.path_A, self.path_B), "raw")
+        display_fn(self.showG(wA, wB, self.path_bgr_A, self.path_bgr_B), "masked")
```
oatssss: I believe this is the cause for warped masked preview. Should be `tB`/`tB` instead of `wA`/`wB`. Not sure why @Clorr used the warped versions...

Apollo122:
Are you following this conversation @oatssss? Maybe we should add it too. https://github.com/shaoanlu/faceswap-GAN/issues/73#issuecomment-368245627

oatssss:
I haven't tried to use perceptual loss. I do get the best results using shaoanlu's original repo. I tried to make the model here exactly the same, but it's still giving me different results and I have no idea why.

Apollo122:
Convert doesnt seem to be doing any correction you can see the box.

plugins/Model_GAN/Trainer.py
Apollo122: If this is the reason then same change is needed in GAN128 trainer too

Apollo122:
Fixed the PL error. Actually its a bug within keras_vggface.
Add ` K.set_learning_phase(1) ` [Referance](https://stackoverflow.com/questions/42969779/keras-error-you-must-feed-a-value-for-placeholder-tensor-bidirectional-1-keras)  to keras_vggface\models.py before the line 275 meaning above  ` model = Model(inputs, x, name='vggface_resnet50') `

TooMuchFun:
Yes yes, adding `K.set_learning_phase(1)` is essential. I had figured that one out the other day but should have spoken up on this PR too. It uses way way more memory. I had to turn down to batch size 16 to run it on a 1070 w/ 8GB. Training goes great, conversion is still iffy but much more reliable than the facebook-GAN notebook.

Apollo122:
i have 1080 8GB and bs 20 works on GAN128 but not 32. maybe i play it little more to find max value. But yeah before PL it was 40.

oatssss:
@Apollo122 That result looks pretty good! The masking issue at the edges of the bbox might have to do with its size. The bbox in your image extends to way more than just the crucial parts of the face.

Apollo122:
@oatssss Normally this issue is solvable by using erosion-kernel-size and blur-size but in our convert code these are exclusive to Masked converter. Maybe you can add these options to GAN converter as well?

oatssss:
Did you get an error about `K` not being defined?

oatssss:
It's weird, but now even using my local copy of shaoanlu's jupyter notebook, I don't even get the good gan64 results as I was getting before, and afaik I didn't change anything.

Apollo122:
Didn't tried your last commit but I added just to be sure.
That's weird maybe you should delete existing one and train a new.
Also we can check the ports both 64 and 128 to see if they are exactly the same with original.
But I think convert should be the priority. We should add blur, kernel and seamless but also check how shaoanlu smooths the bbox and add that option as well to both convert script.
@oatssss are you up for it? :)

Apollo122:
@oatssss 
check [shaoanlu' convert](https://github.com/shaoanlu/faceswap-GAN/blob/76e14c42efa0c0adbb731d748f866e07793708d1/FaceSwap_GAN_v2_sz128_train.ipynb)
Under section 13, use_smoothed_mask, use_smoothed_bbox and use_color_correction(histogram-matching) i think these can be applied to our convert sciprt

oatssss:
I don't know what `use_smoothed_mask` does, but I believe `use_smoothed_bbox` is better fit to go into the extraction process. If I'm not mistaken, the strategy would be to use the kalman filter to get the smoothed bbox transformations, then save them to `alignments.json`. Potentially we could also use mtcnn for the extract too.

Apollo122:
Yeah and adding blur is important at least it is for the original script.
What we should do is add all there's anything in shaoanlu's convert and experiment on it to get the best results. And yes using alignment.json is important and we should use it for the convert part. So don't waste your time getting face landmarks in convert as they should be provided in alignment.json. There's also a minor improvement you can add. Details in this discussion

TooMuchFun:
Working on MTCNN right now. Hopefully will get staged in a few days...

Apollo122:
@TooMuchFun so MTCNN is for detecting the faces right? If so it should be added in to extract script and save the values in alignment.json which will be used in convert.

Apollo122:
In my opinion mask results are not great for my model either. For how long did you train? I trained that set for 25k iterations with batch_size 40. Also that set only contains 300 photos each. You may try again with PL enabled. Also there is a tip under define_loss function to be used after 15k iteration. You can change those lines too

ruah1984:
Can this be use now? I just get my new setup gpu with 11GB, I can help to try it again

Apollo122:
@oatssss nope.

Apollo122:
@ruah1984 convert needs work. Some essential features are missing. Check the image I posted you can see the bbox. I believe @oatssss is working on it. 
But you can start the training I don't think model would change.

oatssss:
@ruah1984 yea maybe you can confirm that you get good results cuz I don't

Apollo122:
@oatssss 
i added the optional lines and commented them out. This is for the GAN128 trainer script. i believe it would be the same for GAN64 script as well. You can merge it. Preferably you may define a boolean to activate these.

` def define_loss(self, netD, real, fake_argb, fake_sz64, distorted, vggface_feat=None):   
        ...........................
            loss_G = 1 * self.loss_fn(output_fake, K.ones_like(output_fake))  
        # ==========  
        loss_G += K.mean(K.abs(fake_rgb - real))
        **# loss_G += K.mean(K.abs(fake - real)) #OPTIONAL After 15k iteration**
        loss_G += K.mean(K.abs(fake_sz64 - tf.image.resize_images(real, [64, 64])))
        # ==========
        
       ...............................
            real_sz224 = Lambda(preprocess_vggface)(real_sz224)
            # ==========
            fake_sz224 = tf.image.resize_images(fake_rgb, [224, 224])
            **#fake_sz224 = tf.image.resize_images(fake, [224, 224]) #OPTIONAL After 15k iteration**
            fake_sz224 = Lambda(preprocess_vggface)(fake_sz224)
            # ==========   
          .....................................
        
        return loss_D, loss_G `

oatssss:
cool, do you wanna make a PR from your own fork into this branch so you can have the commit, or do you want me to just add it myself?

Apollo122:
Yeah can just add it yourself man.

Apollo122:
I have been training another model with 1300(A) and 3000(B) images. Quality is really good but mask doesnt seem to be working right. Instead of handling the occultation it seems just removing it. And on mask screen I see almost pure white columns(rgb 254,254,254) . For Face_A 2nd column is white and for Face_B 3rd column is white.

@oatssss 
my loss value seems stuck at 0.17 ~ 0.18 but [shaoanlu says that loss value in GAN does not represent output quality](https://github.com/shaoanlu/faceswap-GAN/issues/43#issuecomment-364634620)

ruah1984:
Same as what i have right now, running 8K of face A, 10k of face B (big sample size), and batch size=64, 5k iterations.  

face A 2nd column is white, face B 3rd column is white. 
i have try shaoanlu repo for GAN before, the result for the mask are different.

oatssss:
@ruah1984 can you retry shaoanlu's using the same faceset and see what kind of masks you get?

ruah1984:
@oatssss ,first try 5k, FaceSwap_GAN_v2_sz128_train. (Shaoanlu)
[1/150][5100] Loss_DA: 0.182680 Loss_DB: 0.180278 Loss_GA: 0.717723 Loss_GB: 0.731879 time: 3412.171057
because is NSFW, i only can share you the mask. from what i get from Shaoanlu faceswap-gan 128, still can observe the face shape, but our faceswap here is totally white as what i comment before,  something wrong.I will continue train until 15k  and check if longer hour train, any change for Shaoanlu faceswap-gan 128

face_a
face_a

face_b
face_b

ruah1984:
@TooMuchFun  @oatssss ， if the MTCNN complete i  can try using with the result i have from shaoanlu faceswap GAN128 repo to verify the MTCNN. my new setup 1080TI at least can help for all trial run.

TooMuchFun:
Go ahead without for now. Training shouldn't be affected by it, its just a consideration for better alignment. I'll probably submit the MTCNN work on a few branch into master. Its also not clear to me whether moving to pytorch would make MTCNN unneeded.

Something is off for me in masking steps for this branch recently. Masks degrade for me don't really improve even after 40k iterations. I have not seen this issue in the GAN128 fork on #151 nor faceswap-GAN. I'm wondering if there are unintended differences in the model between the GAN implementations...

Apollo122:
Additional options are always welcomed. So your work on MTCNN would be merged most likely.

Clorr:
@oatssss Thank you for taking time to review/correct/enhance all of this.

My first GAN plugin was based on [the first shaoanlu's attempt to make a class from notebook](https://github.com/shaoanlu/faceswap-GAN/blob/master/temp/faceswap_GAN_keras.ipynb) and I never checked how consistent it was with the notebooks.

On the GAN128 part I tried to make a plugin directly from the notebook. However it is quite hard to modify the code so it fits in classes. Actually, I was thinking reverting all and doing this again from scratch...

If the guys here are ok with your code, I'm happy if it goes into master because on my side, it has been a bit tedious to do all of this ;-)

Apollo122:
Mask output is different from shaoanlu's maybe there are other different things too. @ruah1984 tests for both. I think it should be resolved before merging to master . Also without convert features I think it would be useless for non-programmers.

oatssss:
https://github.com/oatssss/faceswap/commit/44651d7a8dbd1b100c25c391ec93567c29d33259 I'm able to get masks somewhat close to what I was getting before with shaoanlu's if I weaken the generator.

Face A
screen shot 2018-02-27 at 9 51 08 am

Apollo122:
So we can't just use Shaoanlu's code. should we also experiment on it so get the same results? Or there are parts that we are unable to Port?

ruah1984:
@oatssss i try 5 to 10k itr tonight, let's check able to get same result from shaoanlu or not. https://github.com/oatssss/faceswap/commit/44651d7a8dbd1b100c25c391ec93567c29d33259

oatssss:
@ruah1984 pull the latest commit https://github.com/oatssss/faceswap/commit/f3bb173cf1038888e10bc88e6fcc4a1127c6f043 from that `update-GAN-v2-tests` branch. It'll automatically restore the generator strength back to normal past 10k iterations. If you want to do it at 5k, just change [these lines](https://github.com/oatssss/faceswap/commit/f3bb173cf1038888e10bc88e6fcc4a1127c6f043#diff-d40a5cd2e84ef4154ddc11be9a7128acR133)

@Apollo122 as far as I can tell, we are using Shaoanlu's code. At least for the model/trainer, if you can find any differences please let me know. There may be a difference in the generation of training batches, but it seemed to me like there wasn't, but I think this could have an effect on the results we get.

Apollo122:
@oatssss so i started comparing the model code between [yours ](https://github.com/oatssss/faceswap/blob/update-GAN-v2-tests/plugins/Model_GAN128/Model.py)and [original](https://github.com/shaoanlu/faceswap-GAN/blob/0260702077261b77242217d04734426851f7ce42/FaceSwap_GAN_v2_sz128_train.ipynb). I may sometimes ask stupid questions so bear with me :)

First question im gonna ask is that in original code on section 6 load models i see that `netDA, netDB, encoder, decoder_A and decoder_B` get their weights loaded. There are additional weights that we don't create and use such as `encoder.h5, decoder_A.h5, decoder_B.h5` .On your code encoder, decoder_A and decoder_B don't get loaded with weights. Instead netGA and netGB are loaded with weights. Is this intended?

oatssss:
Yes, please ask many questions! Hopefully I can answer them. For this one, yes it's intended but that can change. The difference between shaoanlu's and this port is basically how we choose to save the model. This port opts to save the generator model directly, whereas shaoanlu saves the the encoder and decoders that make up the generator.

In both [the port](https://github.com/oatssss/faceswap/blob/f3bb173cf1038888e10bc88e6fcc4a1127c6f043/plugins/Model_GAN/Model.py#L109-L111) and shaoanlu's we have:
``` python
encoder = Encoder()
decoder_A = Decoder_ps()
decoder_B = Decoder_ps()   
```
You can verify that both functions `Encoder()` and `Decoder_ps()` return a Keras `Model`. We can call `save_weights` and `load_weights` on these model instances to generate weight files for each, and in fact that's what shaoanlu's code does.

Further down in the [code](https://github.com/oatssss/faceswap/blob/f3bb173cf1038888e10bc88e6fcc4a1127c6f043/plugins/Model_GAN/Model.py#L113-L114), we use the encoder and decoder models to construct the generator model:
``` python
netGA = Model(x, decoder_A(encoder(x)))
netGB = Model(x, decoder_B(encoder(x)))
```
Since these are also models, we can also call `save_weights`/`load_weights` on these instances, and in fact that's what our port does.

So that's the difference: we restore the generator as-is and shaoanlu restores the individual pieces and rebuilds the generator.

You'll notice that both generators share the same encoder, so weights for the encoder end up being stored twice. This was already raised in issue https://github.com/deepfakes/faceswap/issues/200.

oatssss:
I generated a summary for the models and compared them to shaoanlu's https://gist.github.com/oatssss/3c1bcae6d00c8ce7f4852ac9e397cee9, they appear to be the same. Whatever's going differently is I think either in the trainer or the generation of batches.

Apollo122:
If we are also sure than gan64 model is the same with original. We can move on check the trainer scripts.

Apollo122:
Thanks for the merge @oatssss .
@iperov can you clarify this: in extract if user selects hog your code doesn't work but if he selects CNN your code activates, first it tries CNN if it fails it tries hog. Or did you remove the options (cnn or hog) and combined them together?

Clorr:
Note that maybe the code from the faceswap-GAN repo has changed since I took the code. If there is a difference in the weights file handling, it is not intentional from my part...

Apollo122:
@Clorr I think we settled the weight handling difference. I will check training code sometime

oatssss:
I enabled the Masked converter for both gan models. This also fixes https://github.com/deepfakes/faceswap/issues/224 - poor results for faces that aren't upright in the original frame.

Apollo122:
@oatssss great! Yeah I don't think adjust converter needs porting. Now only original gan convert remains. I wonder how the results would be with that and can we use it for other trainers.

@babilio alignment.json file's format has been changed. Have you extracted images again to get the new file?

babilio:
@Apollo122 yes, I extracted with a pull made right before that last commit

oatssss:
@Apollo122 I actually moved the gan convert code into masked so right now when converting using a gan model with the masked method, it uses a combination of the gan mask + face hull/rect. I'll change it so the gan mask is only used if gan converter is chosen.

Apollo122:
No no I meant using the original gan convert with other trainers. Like with use_smoothed_mask, use_smoothed_bbox and use_color_correction(histogram-matching). Not smoothed_masked obviously but the other two. Or are they also tied to gan model?

oatssss:
`use_smoothed_bbox` has to do with extracting, the new face-alignment port might have it included already as I was getting pretty smooth tracking with it

`smoothed_masked` in shaoanlu's code just applies a Guassian blur over the whole mask to smooth out the transitions between fake/real face

`use_color_correction` can be added as an option probably and shouldn't be that hard to do

Apollo122:
Alright thanks for the clarification. so there is only color correction missing and that can be added to Masked converter as well as Gan converter I presume.

oatssss:
Yup, maybe some other cool ideas to add to conversion will pop up later.

babilio:
@oatssss when using the Masked with GAN model, I feel that the new face is too small. When using the old GAN converter, it covered a bigger part of the face and that to me looked better for GAN model (obviously there was the problem when the face was not straight). I feel Masked covers too little of the face and always have to use erosion to get rid of double eyebrows. GAN blends so well so the part that gets patched could be bigger.

I can't find where in Convert_Masked there has to be a parameter that crops the face before getting converted, because the images in training shows more of the face for GAN.

babilio:
The Masked convert vs GAN convert covers a much smaller part of the face. To me GAN convert looks better (when the images are straight, otherwise it looks terrible) because it covers more of the face and it blends well. Do you know what parameter determines how much of the face gets patched? I feel like mask crops or covers out a lot of the face. If you compare the images in GAN trainer to the output you'll see that what gets converted is much smaller.

ruah1984:
have it merge to the master??

oatssss:
@babilio the erosion kernel shrinks the face hull. What if for negative values we use [dilation](https://docs.opencv.org/3.0-beta/doc/py_tutorials/py_imgproc/py_morphological_ops/py_morphological_ops.html#dilation) to expand the face hull? You can also use `rect` mode to convert similarly to the original, un-aligned gan converter, the rect bboxes will be aligned to the face though.

babilio:
That makes sense. Thanks for the explanation. I tried rect and you are right it works so much better because it is aligned. Dilation could work really well with GAN

oatssss:
@babilio are you able to get good results with the gan here?

Apollo122:
Mask bug and histogram matching prevents us to merge. But I think they will be resolved soon.

Apollo122:
@oatssss if `rect ` is better maybe it should be default option for GAN converter?

babilio:
@oatssss I was never able to get shaoanlu GAN working to run for me, so I don't really have much to compare. Results I get here are mixed, some frames look really promising others look weird. I havent tried to convert an actual video. Just was training pics of a and b on extracts from multiple sources.

oatssss:
@Apollo122 `rect` will only be good if the masks are good and are completely black at the bbox edges. I think a face hull expanded with dilation would be better than a rectangle, but we'll see.

ruah1984:
Some feedback.Have try to do convert yesterday with -D cnn -t GAN128 -c GAN128 , with my gpu gtx1080ti, it show out of memory, but it not happen if I use -D hog. For the training, I can use batch size max up to 64, result looks pretty good now compare to what previously I share.

ruah1984:
@babilio, can you share some face preview after you convert?

Apollo122:
@oatssss in gan128 convert I see a resize operation: cv2.resize(face_detected.image, (128, 128))
Don't we need to do same thing for Masked Convert?
Also what's the reason keeping gan converters as blur, histogram matching etc. are only available in masked converter?

oatssss:
Yea there's no point in keeping the gan converter so I'll remove it.

These 2 lines in the masked converter give us the same result as that `resize`.

``` python
mat = numpy.array(get_align_mat(face_detected)).reshape(2,3) * size
face = cv2.warpAffine( image, mat, (size,size) )
```

So as long as you pass the right size to `patch_image()`, you're gucci.

TooMuchFun:
A new faceswap-GAN [notebook](https://github.com/shaoanlu/faceswap-GAN/blob/master/FaceSwap_GAN_v2.1_train.ipynb) was posted today!  

Diffing time...

Clorr:
Here is a link to see todays modifications: https://github.com/shaoanlu/faceswap-GAN/compare/76e14c42efa0c0adbb731d748f866e07793708d1...master

However new notebook is not diff-ed with older as it is a new file...

shaoanlu:
The new notebook is still messy and is very likely be overhauled. I'm still exploring the hyper parameters so it might take at least weeks ~ a month to settle down.

oatssss:
@shaoanlu thanks for all your work! You've inspired me to try and learn about model architecture so I can actually attempt to make my own instead of just porting yours :)

oatssss:
I ported shaoanlu's v2.1 in the [`gan-v2.1` branch](https://github.com/deepfakes/faceswap/tree/gan-v2.1).

It uses more memory. I was only able to run gan-128-v2.1 with a batch size of 2 _without_ perceptual loss.

Apollo122:
Hopefully it will get better.
Any news about masks? If port is the same maybe we should ask shaoanlu directly?
After the masks we can merge it finally

oatssss:
Unfortunately, I didn't run it for long enough to see what the mask results were.

ruah1984:
does any one check if the origin model can be upgrade to model 128 like what Dfake/df done before? 
since we already upgrade model GAN to GAN 128, but it will only allow those people with high Vram  to run it. if possible we can upgrade the origin model to 128 or higher to let those without high Vram to try. i have try GAN 128 before, it only can run with small batch size and more longer time.

babilio:
I think dfaker\df is pretty VRAM intensive too. I was getting a ton of OOM errors and couldn't get it to run for 2000 against 3000 images on gtx1080 but maybe there was something wrong with my configuration.

ruah1984:
@babilio ,do you try GAN 128 , any error when training? it looks good from the training preview. 
only convert will be the problem  when i run -D cnn t-GAN128 c-GAN128, it will show out of memory .
and if with only -D cnn / hog , the result is not good.

babilio:
You can try using -c Masked now with GAN it works way better. Also if you used serializer with extract you can use that for convert and that should take care of OOM.

I am extracting again because I think there was a change to the serializers so convert is not doing anything for me know. I haven't done extract with oatsss changes

babilio:
@oatssss the original model is not working for me after the last pull Loading Trainer from Model_Original plugin...
Exception in thread Thread-1:
  File "C:\Users\...\AppData\Local\conda\conda\envs\fs\lib\threading.py", line 916, in _bootstrap_inner
    self.run()
  File "C:\Users\...\AppData\Local\conda\conda\envs\fs\lib\threading.py", line 864, in run
  File "E:\faceswap\scripts\train.py", line 148, in processThread
    trainer = trainer(model, images_A, images_B, self.arguments.batch_size, self.arguments.perceptual_loss)
TypeError: __init__() takes from 4 to 5 positional arguments but 6 were given

babilio:
Also what vggface package did you use, trying to run your new code I installed but it seems to work with tensorflow cpu only for window. So I couldn't even go through an epoch. Thanks

oatssss:
@babilio should be fixed now.

iperov:
I updated faceswap to this PR and Original converter not works. No faces was changed.

oatssss:
It's cuz I added an underscore to the naming scheme of aligned photos. Seems a few people are having this issue so I'll add in a temporary compatibility fix to make it easier for people to just switch between this branch and others.

blvck-mvgic:
I tried running training with GAN128 and after installing `keras_vggface` from repo (which explicitly says it will check for either `tensorflow-gpu` or `tensorflow`) I got some error at some point, this is the last call:
```
File "C:\env\faceswap\lib\site-packages\keras\backend\tensorflow_backend.py", line 2, in <module>
    from tensorflow.python.training import moving_averages
ModuleNotFoundError: No module named 'tensorflow.python.training'
```
Then I tried installing `tensorflow` and training only happens on CPU, which is quite not ideal.

Apollo122:
@blvck-mvgic 
Yeah you should uninstall TF's and reinstall.
Keras_vggface installs regular tf

blvck-mvgic:
I was able to run training with GAN128 only on batch_size=4 on a GTX 1060 6GB, pretty resource intensive isn't it?
With the old trainer I could run it at bs=128

oatssss:
Yea I can't compare to shaoanlu's because there's no 128x128 version of the 2.1 there, I just followed the same pattern to extend 64 to 128, it might not be correct though.
I don't know if the meaning of batch size in this repo is the same as over there, is anyone able to get higher batch sizes for 64x64 using shaoanlu's?

ruah1984:
I am ok up to batch size 64 for shaoanlu and this repo with gtx1080TI.  I still thinking if we can create a repo for all gpu user. for the range of gpu fall within  2 g to 6 g

babilio:
Is perceptual loss for GAN only or can it be used in original model?

Kirin-kun:
Also, not a major issue, but the parameter histogram is misspelled in convert.py : "match-histgoram".

oatssss:
@Kirin-kun I'm not seeing the problem you guys are. I just re-cloned `update-GAN-v2` and it worked.

@Clorr I'm not seeing the ^Ms, neither on GitHub nor my machine. I _have_ written code for this branch on 2 different machines, one windows, one mac so it's very possible the line endings got mucked up around somewhere.

Clorr:
@oatssss sorry if the comments bothered you. Don't hesitate to ask for help at some point. I hope we can merge this soon ;-)

Kirin-kun:
Question: does the GANv2 models support re-training over a different faceset? 

I actually started to train sets of faceA and faceB. After a while, the masks started to show up.

mask

Then I switched the faceA set to another (but kept the faceB set identical), using the same model files, and then, the masks slowly disappeared and the columns became completely black. 

I thought that it needed time to remake the masks, but after hours of training (much longer than when the masks appeared in the first run) the columns stayed black.

I restarted the training with an empty directory and lo, the masks showed up after 1 hour.

oatssss:
I'm happy with the state of this PR tbh, @Clorr if it's cool with you and there are no objections from others, we can merge it?

babilio:
Does keras_vggface needs to be installed separately? Is it just for perception loss? We should add a note to readme or even the requirements? It was kind of tripping me up because the pip install would remove my tensorflow-gpu and just force tensorflow on cpu, which had to be uninstalled and reinstalled gpu version.

cagkanciloglu:
@oatssss how about mask bug? Or has it been resolved?

Apollo122:
@oatssss how about mask bug? Has it been resolved? Last time I tired it just removed the obstacles

ruah1984:
@babilio ,i run the shaoanlu jupyter notebook at the same env,  and it will install keras-vgg once you run it. and it doesn't create any issue for this repo. you can download here 

# https://github.com/rcmalli/keras-vggface

oatssss:
@cagkanciloglu @Apollo122 I think people have reported mixed results with the mask. I haven't been able to find any differences between the model/trainer here and shaoanlu's.

@babilio keras_vggface is only required when you turn on preceptual loss. I will add it to the requirements doc with a note about the tensorflow-cpu/gpu issue.

Apollo122:
@oatssss so you are not getting full white columns anymore on mask preview?

Kirin-kun:
@Apollo122 I'm still training a GAN model, to see where it goes, as we speak and I have white columns on second column for face A and on third column for face B. I think they are mixed up somewhere.

gan

With these numbers:

cmd

oatssss:
@Apollo122 @Kirin-kun I think for face A, the 3rd column is what's actually used as the mask and for face B, the 2nd. @Kirin-kun are your convert results looking good?

Apollo122:
@oatssss yeah man I know you have been busy with faceswap :) it's not about quality though mask feature doesn't work like it should be. It should handle occlusions not remove it. At least that's what I see on preview window. I know that you want to merge this we can do that and open another issue about mask. It wouldn't be ideal but lets do it if you are comfortable

oatssss:
@Apollo122 I don't think you should be training with faces that are partially occluded. If you do, I think you're telling it that partially occluded faces are somewhat acceptable. The more occluded faces you have in the training set, the more it'll think that's what you want. If you only have a few occluded faces, then yea the preview for those images even though they're in the training set should probably have proper occlusion masks since it should understand from all the other un-occluded faces that an occluded face isn't the norm.

Apollo122:
I did that exactly on some NSFW video -you can guess what the partial occlusion is. It didn't work for me. Im not saying it is not working for everyone but didn't see a comment about it. Have you get a proper mask?
You were experimenting with data-generator and said you got the proper mask preview, how did that work out?

oatssss:
@Apollo122 I was experimenting with weakening the generator/discriminator. Kind of like what's done in https://github.com/deepfakes/faceswap/pull/246. I didn't fully explore it because it's hard to gauge a usable ratio between the 2. It was producing masks more similar to what I was getting with shaoanlu's but I can't say for the occlusion cuz I didn't check.

oatssss:
If it's cool with everyone, I'll merge this in a few hours?? Seems like people are waiting on this one to do other stuff.

Clorr:
Go for it. No problem for other PRs you did enough job already. If you merge this, can you "squash and merge" or you prefer otherwise?

Kirin-kun:
@oatssss I didn't succeed to convert. The converter doesn't work (at least in the gan-v2.1 branch). When I tried the GAN converter, it complained about patch_image having too many arguments. It seems the GAN converter doesn't take size as argument. 

In the convert.py, there is 
`image = converter.patch_image(image, face, 64 if "128" not in self.arguments.trainer else 128) `

But in the Convert_GAN.py there is:
`def patch_image( self, original, face_detected ):`

Whereas in Convert_Masked.py there is:
`def patch_image( self, image, face_detected, size ):`

I tried to remove the size, but it then gave another message: 

Failed to convert, Reason: list indices must be integers or slices, not tuple

babilio:
There was never a convert gan working for 128, I think. Masked is the recommended, (maybe that can be added to the param help?)

Kirin-kun:
I think it's just the branch that has problems. I went to master and with the latest commits, I don't have the errors anymore.

But the conversion is just not working : I have just a transparent square over the original face.

And I was not using GAN128, but GAN.

Apollo122:
@Kirin-kun if you use GAN converter it has been depreciated. I believe @oatssss will delete those converters. You should use masked converter. https://github.com/deepfakes/faceswap/pull/217#issuecomment-370146749

oatssss:
Alright, praying for the best. Hopefully this merge won't create too much confusion, I'll help where I can!

iperov:
@oatssss 
hist match glitch
out00002 2
out00002

iperov:
@oatssss but in most cases histogram matching rules!

Issue #207, opened 2018-02-17
train.py loop behavior


GSonderling:
I'm speaking about: 
https://github.com/deepfakes/faceswap/blob/51f1993d93e0ffb581d44416f327f0cf731c34e8/scripts/train.py#L145-L173

As you know pressing enter sends stops the loop after current iteration. Unfortunately the method is rather inflexible:
https://github.com/deepfakes/faceswap/blob/51f1993d93e0ffb581d44416f327f0cf731c34e8/scripts/train.py#L121-L123

This can cause the thread to hang after last iteration.

I did look into the issue, and unfortunately there are only two ways to solve it.

1. Complicated way with additional imports. Sadly, baseline python doesn't include tools we would need to implement desired behavior. We would need to import package with specialized key listeners.
In my opinion that is an overkill. We could also just write our own code from scratch. But the readability would suffer and for very minimal gain. Therefore I would prefer option 2.

2. Drop the `Enter` interrupt and revert to traditional `Ctrl + c` behavior. Everybody who works with cli is familiar with it and python has all the necessary stuff. It is omniplatform code for it is as simple as humanly possible. On chance that somebody doesn't understand it, we can always add some message explaining it. 
Right now the old behavior provides little benefit anyway, since we can already limit the epochs.

I'm asking because I don't want to make mess, submitting useless PR and so on. Granted we could always keep things as they are. But you have to agree, right now it feels sloppy.

Clorr:
I close this as your PR is already sufficient ;-)

Issue #116, opened 2018-02-04
Train output and preview in iPython notebook


flipflopbboi:
Hello and thanks everyone for the amazing work in putting this together!
I am trying to setup a workflow on a Python 3 [Google Colab notebook](https://colab.research.google.com/)

I am currently running the  train command:
```
source_dir = '~/drive/source/trump/extracted'
target_dir = '~/drive/target/cage/extracted'
model_dir = '~/drive/models'
!python ~/faceswap/faceswap.py train -A $source_dir -B $target_dir -m $model_dir -v
```
I have left the script to run for some time and currently getting the following output:

```
> 2018-02-04 12:15:42.099548: I tensorflow/core/platform/cpu_feature_guard.cc:137] Your CPU supports instructions that this TensorFlow binary was not compiled to use: SSE4.1 SSE4.2 AVX AVX2 FMA
> 2018-02-04 12:15:42.228572: I tensorflow/stream_executor/cuda/cuda_gpu_executor.cc:892] successful NUMA node read from SysFS had negative value (-1), but there must be at least one NUMA node, so returning NUMA node zero
> 2018-02-04 12:15:42.229133: I tensorflow/core/common_runtime/gpu/gpu_device.cc:1030] Found device 0 with properties: 
> name: Tesla K80 major: 3 minor: 7 memoryClockRate(GHz): 0.8235
> pciBusID: 0000:00:04.0
> totalMemory: 11.17GiB freeMemory: 11.11GiB
> 2018-02-04 12:15:42.229170: I tensorflow/core/common_runtime/gpu/gpu_device.cc:1120] Creating TensorFlow device (/device:GPU:0) -> (device: 0, name: Tesla K80, pci bus id: 0000:00:04.0, compute capability: 3.7)
> loaded model weights
> Loading Trainer from Model_Original plugin...
> Starting. Press "Enter" to stop training and save model
> saved model weights
> saved model weights
> saved model weights
> saved model weights
> saved model weights
> saved model weights
> saved model weights
> saved model weights
> saved model weights
> saved model weights
```

My questions are:
1. I have disabled the preview argument `-p` because otherwise I get a display X error and the script crashes. Any ideas on how to re-route preview output in an iPython jupyter notebook? Any further details on how the preview works would be useful to try to hack something together.

2. I am seeing only `saved model weights` messages and no printouts on iterations, loss functions etc. is this normal?

3. How can i exit since I cannot provide keyboard interruprs in the notebook shell?
[...]

PR #227, opened 2018-02-27
New control for training.sh loop

GSonderling:
Per the issue #207 submitted by me. I present following change to behavior of main training loop.

In interest of simplicity, I decided to drop the `'Enter'` interrupt entirely, and switch to more traditional `Ctrl + C`.
Additionally, this approach requires no further imports. Thus keeping the code clean and simple. 

It also gets rid of hanging, that sometimes occurs after training is finished.

### DIFF
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -115,15 +115,17 @@ class TrainingProcessor(object):
 
                         key = cv2.waitKey(1000)
                         if key == ord('\n') or key == ord('\r'):
+                            print("Exit requested! The trainer will complete its current cycle, save the models and quit (it can take up a couple of seconds depending on your training speed). If you want to kill it now, press Ctrl + c")
                             break
                         if key == ord('s'):
                             self.save_now = True
                     except KeyboardInterrupt:
+                        print("Exit requested! The trainer will complete its current cycle, save the models and quit (it can take up a couple of seconds depending on your training speed). If you want to kill it now, press Ctrl + c")
                         break
+                thread_event.clear()
         except KeyboardInterrupt:
             print("Exit requested! The trainer will complete its current cycle, save the models and quit (it can take up a couple of seconds depending on your training speed). If you want to kill it now, press Ctrl + c")
             thread_event.clear()
-            print("event cleared")
             thr.join() # waits until thread finishes
 
             
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -99,34 +99,35 @@ class TrainingProcessor(object):
         import threading
         self.stop = False
         self.save_now = False
+        thread_event = threading.Event()
+        thread_event.set()
+        try:
+            thr = threading.Thread(target=self.processThread, args=(thread_event,), kwargs={})
+            thr.start()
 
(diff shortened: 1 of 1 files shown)


babilio:
Enter works fine and reliably for me when I have the preview window and you focus on it before pressing. I am not sure removing it completely is good thing. Couldn't Ctrl+C just stop mid saving weights, potentially corrupting model files?

luiscosio:
I agree that Ctrl + C is better than ENTER and all the difference between keys in OS.

GSonderling:
Oops, I just realized I overlooked indent. 

Anyway, if you look at the code, after I fix it, you can see that the preview window still behaves as it did before. You can still stop it with `Enter`.

GSonderling:
Ok, fix submitted and tested.

Clorr:
Maybe also related to #116

GSonderling:
Any suggestions? Would anyone like something added or changed? Did you find any undesirable behavior?

Clorr:
For me it seems ok, but I haven't tested yet. I could not work on the project lately, but I'm starting to get some more free time so I hope to give you a feedback soon...

iperov:
>  I decided to drop the 'Enter' interrupt entirely.

you decided ? How about existing FaceSwap end-users ? Not developers, users.

3xtr3m3d:
its printing 
> Starting. Press "Ctrl + C" to stop training and save model
so end-users know what to do though

GSonderling:
> you decided ? How about existing FaceSwap end-users ? Not developers, users.

If people don't like it there is no need to accept the PR. But right now the script behaves rather strangely, wouldn't you agree?

It basically just waits for any input followed by `Enter` and if there is none it just hangs. That is a little confusing. 

And for the record, I agree that my fix is not exactly the most elegant or optimal. I would prefer to get rid of those exceptions completely and just use normal key press events.

But all the solutions I tried, and that worked, involved either further imports, which would confuse the users, more threads, which is just an overkill, or worse of all bunch of code that is there just to handle one event.

Now, I don't want to cause problems by making unwanted changes to the code. So how about this, lets go to the playground and put it to vote, so that users decide what kind of behavior they want.

Clorr:
No problem @GSonderling . `Enter` is certainly more natural for very new users, but beside of that is not of a strong interest. `Ctrl + c` adds convenience for usage in special conditions, like notebooks or remote run.

GSonderling:
The enter still works for windowed preview, which is what most new users use, I assume.

3xtr3m3d:
this is good for cloud training like google colaboratory. please merge this

bryanlyon:
I prefer using something a bit more high level than input().  I suggest something like

if cv2.waitKey(1) & 0xFF == 13:
        break

This will break the loop if the 'enter' key is pressed in the window but will not block on ctrl-c.

torzdf:
This has been somewhat mitigated by the latest refactor. I agree that hanging for input is not ideal, but ctrl+c will also terminate the training loop, so there are options, until something better comes along.

This code is now also somewhat behind the current repo, so I'm closing for now. Feel free to re-open against the latest base.

PR #231, opened 2018-03-01
[Patch] Multi GPU model

kellurian:
This is my first pull, so I hope I did this right. The only change should be the Model_original.py file which has worked great for adding multi GPU support. The only problem I have had is with allow growth option. Not really a programmer and new to Github, but trying to learn this stuff.


iperov:
cannot understand what changed

Clorr:
@kellurian there is change in whitespace in your code and it shows like the complete file was changed. Maybe try to copy/paste only changed lines. (If you can't, I can try to do it on my side...)

plugins/Model_Original.py
```diff
@@ -1,68 +1,73 @@
...
+
+class Model(ModelAE):
+    def initModel(self):
+        optimizer = Adam(lr=5e-5, beta_1=0.5, beta_2=0.999)
+        x = Input(shape=IMAGE_SHAPE)
+
+        self.autoencoder_A = KerasModel(x, self.decoder_A(self.encoder(x)))
+        self.autoencoder_B = KerasModel(x, self.decoder_B(self.encoder(x)))
+        self.autoencoder_A = multi_gpu_model( self.autoencoder_A ,2)
+        self.autoencoder_B = multi_gpu_model( self.autoencoder_B ,2)
```
oatssss: According to the [docs](https://keras.io/utils/#multi_gpu_model) the second argument which is a 2 here indicates the number of GPUs to split work between. This shouldn't be hard-coded. What happens if you run this with only 1 GPU? You should add a `--gpus` option to [cli.py](https://github.com/deepfakes/faceswap/blob/232d9313afabc075e28c1a42047b22e0568200e6/lib/cli.py) and use it here.

oatssss: Have you been able to load/save a multi-gpu model? From the Keras docs:
> On model saving:
>
>To save the multi-gpu model, use .save(fname) or .save_weights(fname) with the template model (the argument you passed to multi_gpu_model), rather than the model returned by multi_gpu_model.

So it seems like we still need to maintain a reference to the serial model in order to save.

kellurian: I agree with you, I don't know what happens in a single gpu environment, though I suppose I could disable one of mine to find out. This does work on my system as it splits the load between the two GPU's and uses both the VRAM from the two cards.

bryanlyon: I was able to save one without problem.  I ran one with this model and it works perfectly.  Saving is already handled by train.py in the correct way.  The only problem with this will be if we start using checkpoints, which by default are not multi GPU compatible.

The models are also still compatible with single GPU, as I've taken a multi GPU model and run it on the same model in single GPU mode and then back in multi GPU, so you don't have to wipe it and restart if you change the number of GPUS available.  Since the nodes and layers are the same, it loads without problem and runs fine.

We should be able to simply programatically find out how many GPUS are available and by default use all of them.

The following link shows how to find a number of usable CUDA GPUS in tensorflow, I'm sure some Keras equivalent exists.
https://stackoverflow.com/questions/38559755/how-to-get-current-available-gpus-in-tensorflow

bryanlyon:
I have implemented this patch in pull request #241 along with the -g --gpu option on the command line.  This has been tested and works on all current models.

PR #238, opened 2018-03-03
Allows applying dilation by passing negative erosion kernel values. If value is negative, …

babilio:
…it turns it into a dilation kernel, which allows facehullandrect to cover more space. Can help to cover double eyebrows. Also could be useful with Masked converter for GAN in @oatssss PR.

### DIFF
--- a/plugins/Convert_Masked.py
+++ b/plugins/Convert_Masked.py
@@ -8,11 +8,13 @@ from lib.aligner import get_align_mat
 class Convert():
     def __init__(self, encoder, blur_size=2, seamless_clone=False, mask_type="facehullandrect", erosion_kernel_size=None, **kwargs):
         self.encoder = encoder
-
         self.erosion_kernel = None
+        self.erosion_kernel_size = erosion_kernel_size       
         if erosion_kernel_size is not None:
-            self.erosion_kernel = cv2.getStructuringElement(cv2.MORPH_ELLIPSE,(erosion_kernel_size,erosion_kernel_size))
-
+            if erosion_kernel_size > 0:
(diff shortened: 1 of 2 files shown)


gdunstone:
Can you update the readme with this info as well?

babilio:
I am getting good result with -e -15 and -b 30 with original model.

Clorr:
@gdunstone is the argparse info enough? I don't see where there is an info about this in the README...

Clorr:
@babilio otherwise I'm ok to merge when I get feedback from @gdunstone

gdunstone:
Argparse is enough. Just as long as the knowledge is preserved somewhere.

Clorr:
Ok let's go then ;-)

plugins/Convert_Masked.py
```diff
@@ -8,11 +8,13 @@
...
         self.encoder = encoder
-
         self.erosion_kernel = None
+        self.erosion_kernel_size = erosion_kernel_size       
         if erosion_kernel_size is not None:
-            self.erosion_kernel = cv2.getStructuringElement(cv2.MORPH_ELLIPSE,(erosion_kernel_size,erosion_kernel_size))
-
+            if erosion_kernel_size > 0:
+                self.erosion_kernel = cv2.getStructuringElement(cv2.MORPH_ELLIPSE,(erosion_kernel_size,erosion_kernel_size))
+            elif erosion_kernel_size < 0:
```
leftler: Could there be a potential issue if `erosion_kernel_size = 0`?

babilio: No @leftler the default for the variable is None. Before this commit if you passed -e = 0 it would cause an OpenCV error because it needs a positive number. In this commit erosion_kernel_size 0 just keeps erosion_kernel as None and nothing happens.

PR #241, opened 2018-03-04
MultiGPU plugin

kellurian:
So newby trying this again. I edited the training script to have a command line option for a MultiGPU selection for training. It should still default to the Model_Original.py file and single Gpu option, but you can select -t MultiGpu option and it should then select the Model_MultiGpu.py file and select that model file. I hope I did it right this time. Sorry I took so long to do this, had a conference this weekend and just got home to try to figure this out. Let me know if it works on your systems, it worked on mine; well at least selecting multi gpu worked.


iperov:
separated model not good imho
How about GPU choice option to existing model ? For example [0] or [1] or [all] or [1,2]

kellurian:
I'm afraid I don't follow...I also don't have great skills in this area.  This was an easy fix with what I could do currently. As I am learning to work with keras, I might be able to integrate it into one model file, but for now its what I got. Default option it still selects single GPU, but you can chose to select multiGPU, so it fixes the problem. I'll take a look and see what I can do

iperov:
check here https://github.com/deepfakes/faceswap-playground/issues/90

Clorr:
@kellurian no problem it is nice from you to propose this PR as it will help some people wanting this feature. I also agree with @iperov that we should find a better solution for that question of multi-gpu, and that's why I didn't merge it. But for now your PR is good enough to stay here and help some people try this feature and maybe come with a more advanced solution.

kellurian:
Sure. I’ll look into trying to embed this deeper, but right now it’s above what I can do. Really just trying to help if I can- definitely not on you guys level with this stuff but I have been reading on the keras sites and will post more if can can get a more elegant solution.

Clorr:
no problem @kellurian we don't expect from you more than you can give ;-)

kellurian:
@iperov Yeah, I briefly looked out what you did, looks good from reading.  A lot of what I was reading about using multi GPU stated that we needed to import tensor flow initially in the model file (and I didn't do that), and making the models in parallel data, thought then there is some problem with saving the data in parallel when you are finished...thought the actual data throughput is supposed to be really good- much better than with a single GPU. I keep experimenting and see what I can do...been on call and operating a lot so it may take some time. Any help or suggestions are appreciated- I am an apt pupil to learn from you guys.

bryanlyon:
I have added this feature with pull request #256 .  I have done it safely, added a command line option, and tested it.

PR #242, opened 2018-03-06
Align Eyes Horizontally After Umeyama + Blur Detection

oatssss:
I was getting faces that weren't completely upright even after the umeyama transform, so I used the technique [here](https://www.pyimagesearch.com/2017/05/22/face-alignment-with-opencv-and-python/) to further align the face so both eyes lie horizontally during extraction. I believe this will make the faces more consistent and lead to better training. However, face tracking ends up more jittery so I'm curious to see what final results are like. It might be better to only use this for the creation of face sets and not for finding the alignments of a video for conversion. I added this extra alignment mainly to create better face sets for training.


kvrooman:
On a related note, the hard coded average face values in aligner.py are directly from the dlib toolchain stack -- presumably based off the pre-computed average values from his model https://github.com/davisking/dlib models/blob/master/shape_predictor_68_face_landmarks.dat.bz2

You can see the copy-paste from the lines below
https://github.com/davisking/dlib/blob/b85cb68e79471c819e29081c33d09e7371f41539/dlib/image_transforms/interpolation.h#L1927-L1944

What was the reason to drop the dlib references and alignment tools in favor of these hard coded values and umeyama? There was some follow up discussion on dlib based on how the 5 landmark alignment model was much faster, a more concise model, and a bit more accurate with cnn or hog than the previously used 68 landmark model ( which was used in faceswap prior to the plugin change ) Might start looking into the differences...

Importantly, there is a developer note from dlib that the 68 model is not compatible with cnn face detection due to different bounding boxes during the model training, i.e. only works with hog.  This caused some trouble with some people I can see in the issue history.

oatssss:
Nice, I was wondering where those average landmarks came from. I don't know the umeyama origin, but I'm intrigued by the 5 landmark dlib example. Thank you for this information, I will also look into this :) :)

kvrooman:
Haven't personally tested anything myself (yet) but source is
http://blog.dlib.net/2017/09/fast-multiclass-object-detection-in.html

"Unlike the 68-point landmarking model included with dlib, this model is over 10x smaller at 8.8MB compared to the 68-point model's 96MB.  It also runs faster, and even more importantly, works with the state-of-the-art CNN face detector in dlib as well as the older HOG face detector in dlib.  The central use-case of the 5-point model is to perform 2D face alignment for applications like face recognition.  In any of the dlib code that does face alignment, the new 5-point model is a drop-in replacement for the 68-point model and in fact is the new recommended model to use with dlib's face recognition tooling. "

"The results should in general be the same, but it's faster and smaller. The alignment should actually be slightly more accurate in general, but not by a lot. The real benefit is speed, size, and ability to use it with the CNN face detector in addition to the HOG detector. 

Yes, you can just replace the old shape model with the new model in any face recognition code that used the old one and it will work. I specifically made this new model to be a replacement for the old one. It will create the same kind of alignment as the old model and work with the previously trained face recognition model."

Clorr:
@oatssss sorry I removed the Extract_align folder as it was not relevant anymore with the removal of the aligner.py. However I see that you have modified it so it conflicts. Also I see you have a new align_eyes in lib. If allign_eyes is specific to extract, please recreate the folder and put both files (and more if relevant) in it. Thx

Clorr:
Is that one ready to merge?

oatssss:
@Clorr yea I think this is good now.

PR #246, opened 2018-03-07
add option --balance-loss to train

iperov:


Clorr:
Tahnks @iperov for the PR

I have 2 concerns about it:
- Is that experimental, or does that bring a real advantage that we want to have this as an official feature?
- On the formal part, I'm not fond on how it is written as it brings some duplication of code and is not quite obtrusive. I'd go more for something like below:

```
train_A, train_B = True
if balance_loss:
     train_A = ...
     train_B = ...

if train_A:
    epoch, .... 
    train_on_batch...

if train_B:
    epoch, .... 
    train_on_batch...
```

this way some later options can be added to modify train_A and train_B behavior...

What's your opinion?

iperov:
I dont demand to merge this PR right now.
Let people test difference ?

iperov:
committed code optimize

Clorr:
> I dont demand to merge this PR right now.
Let people test difference ?

Ok for me

i5L4NDOF5T4BiLiTY:
It seems to do exactly what you would think it would do, it trains one encoder twice as fast as two and it concentrates on the higher loss problematic encoder, in my case B.
Seems to be leveling them out nicely.

DLSauron:
Ya I am seeing the same. Looks like it is training A twice as fast since it is the higher one.

iperov:
later...

babilio:
I tried this and left to train overnight and it was producing nice results. I had a smaller set of b that would normally get overfitted but with the balance gave me nice results. Any particular reason you closed this @iperov? Thanks for the work you have been doing by the way, your commits have been super useful.

iperov:
I will reopen later ... waiting gan commit

ruah1984:
cheers !!

ruah1984:
may i know this balance-loss to train option have merge with the master?  the latest repo train not found this option .

Issue #224, opened 2018-02-26
Align rotation of input faces for GAN conversions


oatssss:
Currently, the extractor finds a rotation matrix for each face using umeyama so it can generate a faceset with all the faces mostly upright. Unfortunately this rotation matrix isn't stored in the alignments file, only the bbox (of the un-rotated face) and facial alignments. For the GAN model, when it comes time to convert, the faces aren't rotated upright before being fed through the model so I doubt anyone has been able to get good results for faces that aren't completely upright.

I propose we store the rotation matrix in the alignments file during extract, then at conversion, re-apply it to the cropped face to make it upright before feeding through the model. The swapped output face then needs to be rotated in the inverse direction to match it with the frame again. Hopefully this is possible.

Apollo122:
Definitely worth working on it. Have you checked shaonlu's merge? Maybe there are approaches that we can benefit

oatssss:
I realized that [`Convert_Masked.py`](https://github.com/deepfakes/faceswap/blob/819d10702a8253aec228b8083cdd0857f28e548d/plugins/Convert_Masked.py#L24) already does this, however, it just recalculates the rotation matrix with umeyama on the fly instead of serializing/deserializing in an alignments file.

babilio:
I think serializing it would help implement things like #221 more easily, right? But either option would be great right now because I don't find GAN very useful at the moment because of this problem.

Thank you for all the work you have been doing!

ppmdo:
@oatssss, Yes I've had bad results with rotated faces. I think this deserves attention.

When I read #197, I thought that if the Extract.py script didn't find a face (odd angles), the script could rotate the images (90°) rewrite them to disk, re-try and then store the rotation info in rotation.json or so. Bruteforce, basically. 

Then, when converting, the process would have to invert the rotation and return the frame to its original orientation. However, I don't know if this would also be computationally expensive.

oatssss:
@ppmdo can you see if you get better results with https://github.com/deepfakes/faceswap/pull/217/commits/236112afbc6bb65b0b43e25be37e3308c4f2fd35

torzdf:
@ppmdo 
>  the script could rotate the images (90°) rewrite them to disk, re-try and then store the rotation info in rotation.json or so. Bruteforce, basically.
> 
> Then, when converting, the process would have to invert the rotation and return the frame to its original orientation. However, I don't know if this would also be computationally expensive.
> 
[...]

Issue #197, opened 2018-02-15
Face on angles >180 degrees not recognized on extraction


babilio:
Hi guys thanks for the amazing work here!

I have been following the landmark detection dialogue  #187 and have tried both hog and cnn with both face-alignment and face_recognition. I got face-alignment with cnn working great, with pytorch on win10 now. However, I noticed that none of the above are able to reliably able to identify faces where the face is pointing downwards, for example with the forehead pointing from 6 to 9 o'clock.

I think all these algorithms tend to look for eyes being above the level of the mouth.

For example [nsfw parts covered](https://imgur.com/a/Btk12) this image would not be detected and extracted by hog or cnn in face-alignment or face_recognition.

However by rotating it 90 deg to the right, so that the forehead is pointing up, makes it extracted.

Would it be possible to have an argument set to resend the image for alignment but rotated if it was not caught the first time? 

I am ok with python and novice with git but could maybe even give it a try if someone points me to where the frame is passed for extraction.

Thanks!

Deadfishzzway:
I am pretty newbish with coding and will try anyway, I find this in extract.py

    def handleImage(self, filename):
        count = 0

        image = cv2.imread(filename)
        faces = self.get_faces(image)
        rvals = []
        for idx, face in faces:
            count = idx

            resized_image = self.extractor.extract(image, face, 256)
            output_file = get_folder(self.output_dir) / Path(filename).stem
            cv2.imwrite(str(output_file) + str(idx) + Path(filename).suffix, resized_image)
            f = {
                "x": face.x,
                "w": face.w,
                "y": face.y,
                "h": face.h,
                "landmarksXY": face.landmarksAsXY()
            }
            rvals.append(f)

also rotating from 
https://docs.opencv.org/3.0-beta/doc/py_tutorials/py_imgproc/py_geometric_transformations/py_geometric_transformations.html#geometric-transformations
will report back if I found a solution.

ppmdo:
I'm facing the same issue. Any ideas? I found out that if the image gets rotated so that the forehead is facing upwards, detection works. So maybe if no face gets detected, the script should try again after rotating the image.

torzdf:
I've submitted a PR for this. Feel free to test.
#253

torzdf:
This issue can be closed.

PR #253, opened 2018-03-08
Add image rotation for detecting more faces and dealing with awkward angles

torzdf:
This change allows for the detection of more faces, and often a more satisfactory swap. Dlib has problems recognizing faces tilted over more than 90 degrees from center, so this helps to mitigate that with an optional flag at the extract stage (-r --rotate_images).

If activated, then at extract stage, if a face is not detected the image is rotated 90 degrees and another attempt is made. This pattern is repeated, so in total an image will be scanned at: normal, 90 degrees, 180 degrees, 270 degrees.

If an image is found, then the rotation stops, and the process continues. Obviously, with lots of frames without faces or with faces at awkward angles, this WILL slow down the process, hence why it is an optional flag.

In order to support this, a new item is entered into alignments.json that stores the rotation of the image. At convert stage, this is read back from the file and the correct rotation is done on the output frame to apply the faceswap, before rotating the frame back to the correct orientation.

I needed to amend 5 files for this, but the changes are fairly cosmetic, and *shouldn't* break anything else.

Please let me know thoughts, and merge if possible!

Any questions, let me know.

Cheers


torzdf:
Oh, and I found a bug during testing, hence my revert. You should only focus on the final commit

iperov:
have you example where your fix can detect more faces without flip-glitches by aligner?

Clorr:
@torzdf No problem for the revert, it will be squashed.

torzdf:
@iperov? How do you mean? I have nothing I can share, no, but results have been fine for me. Please feel free to test, any feedback is good.

This doesn't do anything complicated other than rotate the image. The alignments pulled from dlib are correct for the rotated image. The image is then rotated again (by the same amount) for applying the swap, so there shouldn't be any glitches here? I haven't done anything complicated like transposing dlib matrices. That would probably be a lot more efficient, but was beyond me at this stage.

Clorr:
Related to #197

torzdf:
Oh and I removed the count variable from handleImage because it wasn't doing anything.

babilio:
This is working well and #259 that I created it works pretty fast on extraction 8,000 images ~ 40mins and about ~1500 where sideways or upside down and found with this. However the convert seems to be taking much longer. I have a local branch with this and #260 and it is taking 2hrs for convert, I am not sure if it is because of this pull request particularly. I can test more tomorrow. Thanks for creating this!

torzdf:
Hmmm, I did not notice any difference on convert. I found extract was slower if there were lots of images with no faces/at awkward angles, but that is expected.

Convert has to do less work: it checks the rotation from alignments.json, (which it already has loaded) and only rotates the image if a non zero value is found. Obviously if a non zero value is found, it needs to rotate the image to perform the swap and then rotate back to the correct orientation, but the speed loss should be marginal.

If there isn't an alignments.json, it should just default to 0 rotation, so would never rotate the image.

Would be interested if anyone else is having issues, I will test more, but am unable to replicate at the moment.

torzdf:
It isn't automatic. It's run with the command line switch -r (--rotate-images), so is fully optional.

If the switch isn't specified, then no rotation will be attempted, and it will run as it always has

torzdf:
Also indirectly relates to issue: #224 where this was discussed

bryanlyon:
I apologize, my message was incomplete.  I mean rotate explicitly when told to.  As in don't try to check for faces upright first, but when given the -r command only try with rotation.  Since trying upright first may find the same (incorrect) face and then skip the rotate.  Perhaps even with explicit angles listed (-r 90, -r 180, or 270).

torzdf:
Ok, I see what you're saying. That should be a relatively straightforward add, but I think I would leave it as an improvement for after it gets merged, if it does, in its current form.

I may be wrong, but I envisage that most people will activate it when they know they have faces in their target video that may be problematic, so it makes sense to pass all frames in one go, rather than doing upright first and then doing a second pass to attempt to get rotated faces (it would be slower that way).

That said, adding options to the command line argument to check normal and rotated or just rotated may be something I could look to instigate down the line, but I don't see it as a must have. Ultimately it will only perform rotation on frames that it can't find a face on, so if a video is passed in with 99% upright faces, the slowdown will not be high.

bryanlyon:
In my experience, videos with tilted faces will have both an upright face and a tilted face.  The upright face is often not the face that is desired.  This PR solves the problem only if there is not upright face.

My main concern is that if this is added as it is, then future modification to enable that flexibility would harm backward compatibility with command line options.  Obviously that isn't a huge deal, but changing the function of a command line switch is something we should avoid.

bryanlyon:
Apologies for the double post, but I came up with a way to make it work.  This could be adopted and in the future if we enable explicit rotate then "auto rotate check" could be enabled when no number is passed, but if a number is passed, the image will be explicitly rotated the degrees specified and then run through face recognition at that angle.

torzdf:
I'm not sure I follow your second point.

As to the first. the main issue with what your saying is that to detect all faces would mean rotating every single frame the full amount, which would lead to significant (and probably unacceptable) slowdown. It is still possible to use the --filter flag to filter faces with the image-rotator. I haven't used it myself, but if you are just looking to identify a particular face, this would seem to be the way to go, rather than looking to re-invent that functionality here. That way if the face you are looking for is not detected and another one is, then that face would be ignored, the rotation will be performed and it will scan again for the filtered face.

Ultimately I needed to make a trade-off between speed and effectiveness.  The best way to guarantee not to pick out a face you don't want is to mask the video prior to extracting faces.

That being said, I could change the command line flag to take options:
- off: default. Don't perform rotations (current faceswap behaviour)
- on : only rotate the image if a face isn't found (current rotator behaviour)
- force: rotate every frame all iterations (would capture all faces, and probably a lot more false positives, and would slow down the extract significantly)
- only-partial: Don't attempt to extract from upright frame, only perform rotations, stop when face is found
- only-full: Don't attempt to extract from upright frame, only perform rotations, but perform them all (again slow).

To be honest, this seems like overkill to me (I certainly would only ever use the first option).

I'm interested to see what other people's use case would be though, before embarking on a code overhaul. At the minimum, though, I could change the flag to accept arguments (just on or off, default off), so I could easily add options down the line.

babilio:
I think you are right for now @torzdf. The problem right now is that there is nothing that deletes the alignments after extraction so if you do a run through with off, delete bad faces, run again with forced angle, it will have the bad entry in the alignments. So if you tried using skip extracted then it would get skipped anyways.

torzdf:
@bryanlyon 

>  Obviously that isn't a huge deal, but changing the function of a command line switch is something we should avoid.

I've taken this on board, and changed the command line switch to an option (currently accepting on and off) so that I can more easily amend in future if needs be.

babilio:
Yeah I think that's a good solution. Thanks for putting messages only on verbose mode too btw! Also retested this for performance by itself and convert is working great. Barely drop in images per second, even with images upside down.

torzdf:
That's good to hear. 

In my opinion this is ready for merge if enough people are interested. I know @ppmdo and @Deadfishzzway showed interest in Issues. Tagging in case they want to test/feedback.

torzdf:
@babilio How are your results, by the way?

@iperov mentioned:

> have you example where your fix can detect more faces without flip-glitches by aligner?

All my converts have come out fine, and I'm not sure what this potential glitch is, but it would be good to know whether you have run into any unexpected issues on the final output, or whether it's as expected.

babilio:
They were good I had a long video that had even upside down faces. Like 1500 out of 8000 that got picked by the rotation only. They got converted back nicely too. This is great, I have always struggled with the awkward angles and how to spot faces and bulk rotate before faceswapping, this is so much more convinient

ppmdo:
@torzdf Will test this asap and report.

iperov:
@torzdf 
`All my converts have come out fine, and I'm not sure what this potential glitch is, but it would be good to know whether you have run into any unexpected issues on the final output, or whether it's as expected.`

so you dont know why your fix better, because you cannot provide any comparisons, and people have to test your fix ?

babilio:
@iperov this fix is not meant to catch glitches on this extraction, it's to be able to detect sideway faces or upside down faces. Dlib expects eyes to be above nose. So if the person is like laying down on the side or face is upside down it won't get detected without this fix.

iperov:
I just want to see comparison BEFORE and AFTER this subj.

iperov:
for example
@torzdf extracted 100 faces, then he added rotations and extracted 105 faces - I want to see proofs

babilio:
So the results I reported is not enough? Or are you saying that contributors should submit proof with their pull requests?

babilio:
Yeah this I agree, this is a good practice actually

Clorr:
A proof would mean the committer has tested things in various conditions and shows it works in all the cases. This can be expected when you have some critical project with hard consequences.

Here, @torzdf is proposing an improvment, that is only an option. Maybe in some cases it works maybe in other it doesn't. The PR enables people to have a look and say wehter this was useful to them or not or what behavior they encountered. I need absolutely no proof. If someone says "Oh it helped me in some cases", it is already good enough.

The only requirment for a PR is that the advantages overcome the drawbacks.

scripts/convert.py
```diff
@@ -1,235 +1,241 @@
...
-            if is_have_alignments:
-                if self.have_face(filename):
-                    faces = self.get_faces_alignments(filename, image)
-                else:
-                    print ('no alignment found for {}, skipping'.format(os.path.basename(filename)))
-                    continue
-            else:
-                faces = self.get_faces(image)
-            yield filename, image, faces
+import cv2
```
Clorr: There is new line problem in convert.py the diff shows all lines as changed while you just modified some. Can you plz check?

torzdf: @Clorr Yeah, really annoyingly I merged with upstream master just before @oatssss fixed his newline issue, so the issue in that code pulled into my repo. Now my Git says that I'm in sync with upstream, although I'm clearly not :/

This was late last night, so I went to bed. Going to try to fix this morning.

I would ask that people refrain from testing until I get another commit in. I will update

Clorr: No problem, this happens ;-)

torzdf:
@Clorr should be fixed, so this should be good for pulling (I ran a quick test and all is fine and up to date).

@iperov I totally get where you are coming from, and if this was a more substantial change I would provide proof. However, I made this change mostly for myself. I have a deadline to hit so my dev box is currently running 24/7 performing converts (with the image rotator on as it happens). and I only have one usable gfx card, which is currently being utilized. Unfortunately the frames that are being generated cannot be shared in the public domain. 

I knew other people wanted this feature, so I raised this as a pull request knowing/hoping that other people would test it and confirm the results.

In an ideal world I would have provided an output, and if I commit a more substantial change in future, I will.

Cheers

torzdf:
bloody hell, @Clorr. That was quick!

Thanks!

PR #254, opened 2018-03-08
image sort tool

iperov:


Clorr:
Thanks for the PR, i think the tool is useful. However i would prefer to integrate it as a standalone utility rather than integrate it to faceswap.py because each script added to faceswap.py has it's import loaded and it's already a problem with the 3 main command we have now.

I'd rather go for a tools/ folder where we put utility scripts which will be run directly. Can you do this?

Clorr:
Otherwise, there is one drawback I see here, but maybe it is not a problem. As images are renamed, isn't this a problem if we wanted to use names to reconstruct a movie? It is for now the recommended way to convert a movie to extract images with ffmpeg and with a name that is the frame number

iperov:
> As images are renamed, isn't this a problem if we wanted to use names to reconstruct a movie?

How you can reconstruct movie from aligned 256x256 faces?
Aligned folder only for trainer.

Clorr:
Ah ah you're right, I messed up extract with source images...

iperov:
> I'd rather go for a tools/ folder where we put utility scripts which will be run directly. Can you do this?

ok i will try

PR #256, opened 2018-03-08
Add multi-GPU support.

bryanlyon:
Added multi-GPU support through -g --gpu flag.

This has been tested on all 3 models (lowmem, original and GAN).  It works with the default GPUs being 1.  Currently does not use multi GPUs for convert, as that was untested and may cause other issues.  However, models can change the number of GPUs in use without issue.

Save in faceswap is already multi-GPU friendly and does not require any further changes.  One restriction will be against using the built in Keras checkpointing function as it is currently not Multi-GPU safe.  However, other checkpoint modules exist which are supported.

### DIFF
--- a/plugins/Model_GAN/Model.py
+++ b/plugins/Model_GAN/Model.py
@@ -10,6 +10,8 @@ from keras.optimizers import Adam
 
 from lib.PixelShuffler import PixelShuffler
 
+from keras.utils import multi_gpu_model
+
 netGAH5 = 'netGA_GAN.h5'
 netGBH5 = 'netGB_GAN.h5'
 netDAH5 = 'netDA_GAN.h5'
@@ -21,8 +23,9 @@ class GANModel():
     img_shape = (img_size, img_size, channels)
     encoded_dim = 1024
     
-    def __init__(self, model_dir):
+    def __init__(self, model_dir, gpus):
         self.model_dir = model_dir
+        self.gpus = gpus
 
         optimizer = Adam(1e-4, 0.5)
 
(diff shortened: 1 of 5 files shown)


kellurian:
Looks good man.  I like what you did with this. Is it possible to bring this over to the convert and extract processes because these take uber long currently?

bryanlyon:
It can't be used in extract, that is done outside of the models.  Perhaps another method could make the extract multi-GPU.  It's already able to do multiple CPU threads, but there is no way to force it to use CPU if you have a GPU enabled dlib.  That's something I might look at later, but it's a big change.  For my needs, I use a server with just CPUs to do the extract, so that my GPUs are free to train.  I can then use -j 24 to max out my dual 6 core CPUs.  CNN is a bit slow, but with the newest patch, you can --skip-existing to finish up the missed faces with CNN much quicker.

It can probably be moved to convert, but I haven't done enough testing, and in my experience, convert takes a less than 1/4 a second per picture, so actually might hit bandwidth limits before computation limits.

bryanlyon:
I have tested and confirmed that conversion is safe as well with the multigpu.  This latest commit should enable -g --gpus on conversion as well.

Extract is still done with different techniques and requires a different implementation for multi GPU support.

plugins/Model_GAN/Model.py
```diff
@@ -56,6 +59,11 @@ def one_minus(x): return 1 - x
         # img as input => generates encoded represenation and reconstructed image => determines validity 
         self.adversarial_autoencoderA = Model(img, [reconstructed_imgA, out_discriminatorA])
         self.adversarial_autoencoderB = Model(img, [reconstructed_imgB, out_discriminatorB])
+        
+        if self.gpus > 1:
+            self.autoencoder_A = multi_gpu_model( self.adversarial_autoencoderA , self.gpus)
```
Clorr: shouldn't it be:
`self.self.adversarial_autoencoderA = multi_gpu_model( self.adversarial_autoencoderA , self.gpus)`?

bryanlyon: Oops, good catch.  Yes.  My testing was minimal on GAN since my GPU only has 4gb of ram and I get OOMs on that one.  I was borrowing another GPU and only ran enough of a test to verify that the model file is loaded and saved correctly.  I've fixed and committed.

bryanlyon:
One note on how MultiGPU works, since it works by mirroring the model on all GPUS then splitting up the batches across all the GPUs, you want to increase the batch size by a multiple of the number of GPUs you are using.  I.E. if you can normally run batch sizes of 32 and use 2 gpus, you should increase the batch size to 64.  Otherwise, you're using all the GPUs, but you're leaving performance on the table.

bryanlyon:
I am currently rewriting this to merge into the new master.  There is no point in merging this branch for now.

bryanlyon:
This has been resubmitted with the update to the new plugins in #272.

PR #258, opened 2018-03-09
Histogram matching with option -mh

oatssss:
cv2's seamless clone already does colour correction, but it sometimes doesn't act the way you want. Here's another option to try.

### DIFF
--- a/plugins/Convert_Masked.py
+++ b/plugins/Convert_Masked.py
@@ -6,10 +6,10 @@ import numpy
 from lib.aligner import get_align_mat
 
 class Convert():
-    def __init__(self, encoder, blur_size=2, seamless_clone=False, mask_type="facehullandrect", erosion_kernel_size=None, **kwargs):
+    def __init__(self, encoder, blur_size=2, seamless_clone=False, mask_type="facehullandrect", erosion_kernel_size=None, match_histogram=False, **kwargs):
         self.encoder = encoder
         self.erosion_kernel = None
-        self.erosion_kernel_size = erosion_kernel_size       
+        self.erosion_kernel_size = erosion_kernel_size
         if erosion_kernel_size is not None:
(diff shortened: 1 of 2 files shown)


ruah1984:
i thought it already merge in https://github.com/deepfakes/faceswap/pull/217 ?? any way thanks for everything

oatssss:
#217 isn't merged yet so this one can go in separately if people want

Apollo122:
@oatssss can we use this together with seamless clone to remove the flicker? If we specify -seamless true and -mh true will it first use seamless clone then mh to correct the flicker?

oatssss:
@Apollo122 it'll use seamless, then match the histogram afterwards so it should reduce flickering.

oatssss:
@Apollo122 when seamless is on, seamless clone is the last step in the process to patch it back into the image. Seamless does not use a facehull mask to patch it back in since cv2 handles creates its own mask automatically based on context. What I can do is if both seamless and histogram matching are on, after the seamless clone, pass it through to be color-corrected then hand it over to the regular facehull mask to patch it back in, instead of seamless cloning it in. Trying to seamless clone it back in after color-correction might just bring the flicker back, not sure.

Apollo122:
Great! It would be a nice feature. Seamless clone is useful except flickering but we can minimize it with hm

oatssss:
Closing since this was merged alongside #217.

Apollo122:
@oatssss any progress on seamless+hm ? I will train a new model with GAN and and curious what the results would be

kvrooman:
I had an error thrown for me if i use the -mh flag after the -m flag when i convert.  Can't find model etc or something...  It worked if I put the -mh flag before the -m flag in the command line though. Might be useful to change the flag name to something more unique ( -MH ? ) or do you know why that would have happened for me?

PR #236, opened 2018-03-02
fix #233, #237, + working with any batch size

iperov:
fix #233 - primordial bug when dlib recognizes less faces than FakeApp, reason: rgb-bgr order affects chance of face recognition. Numpy.imread and cv2.imread loads same image in different rgb-bgr order.

fix #237

fix working with any batch size
batch size = 1

python_2018-03-02_19-01-59

2:

python_2018-03-02_19-02-37

3:

python_2018-03-02_19-03-18

 etc

PR #259, opened 2018-03-09
Improving performance of extraction. Two main changes to improve the …

babilio:
…most recent modifications to extract: 1st FaceLandmarkExtractor would try to use cnn first, then try hog. The problem was that this reduced the speed by 4 for images where cnn didn't find anything, and most of the times hog wouldn't find anything either or it would be a bad extract. For me it wasn't worth it. With this you can specify on input -D if you want to use hog, cnn, or all. 'all' will try cnn, then hog like FaceLandmarkExtractor was doing. cnn or hog will just use 1 detection method. 2nd change is a rehaul of the verbose parameter. Now warnings when a face is not detected will just be shown if indicated by -v or --verbose. This restores the verbose function to what it once was. With this change I was able to process 1,000 per each 4 minutes regardless if faces were detected or not. Performance improvement just applies to not detected images but I normally will have lots of images without clear faces in my set, so I figured it would impact others. Also the introduction of 'all' would allow trying other models together more easily in the future.


kvrooman:
Agreed, the detect with all detectors in dlib_detectors was unnecessarily reusing the less accurate hog model after running the computationally more intensitve cnn model even if the user specifys just the cnn. Good suggestion to add an -all flag to retain current code logic.

There are some other optimization issues I was looking at as well in the extractor ( scaling to 2048, upsampling, calling the dlib detector as a batch instead of with single images... )

lib/FaceLandmarksExtractor/FaceLandmarksExtractor.py
```diff
@@ -105,25 +105,27 @@ def get_pts_from_predict(a, center, scale):
...
     global keras_model
     global is_initialized
-    if not is_initialized:       
-        
-        dlib_cnn_face_detector_path = os.path.join(os.path.dirname(__file__), "mmod_human_face_detector.dat")
-        if not os.path.exists(dlib_cnn_face_detector_path):
+    if not is_initialized:
+        try:
+            dlib_cnn_face_detector_path = os.path.join(os.path.dirname(__file__), "mmod_human_face_detector.dat")
+        except:
```
iperov: wtf ?

babilio:
Yeah @iperov I wrongly thought that creating the path would fail if it it didn't find the file but it's just a string. So I changed that.

I know that English is not your first language but I'm gonna ask you to be more cordial and friendly that just typing WTF! You can say, "Please see this" and it will have the same effect. We are all just trying to improve the tool. I can't tell if you get upset with people and their edits or if it's just language but in any case there is no need to be aggressive when everyone is working with the same goal.

Clorr:
@babilio I agree with you

babilio:
I understand, we're good :)

lib/FaceLandmarksExtractor/FaceLandmarksExtractor.py
bryanlyon: I think what he means here is that a try/except without actually handling the error is an inappropriate solution to error handling.  The exception should either be dealt with (or if ignored, at least an error printed to the console) or passed up to the caller to handle.

iperov:
Because I very mad !
Why I must check all PR's to detect ppl something erasing I wrote before ?
Latest repo commit also erased my batch size fix #236.
And your PR before my code review was tried to erase important code.

babilio:
I tested my changes and it was working fine before I made pull request. The detector was loading fine. I understand what would have happened if it didnt found the directory would have not been good. You went ahead and and forced to use 2 detection methods even though that is not what the command line argument says. No one got angry or upset over that. I tried to improve it. Then you found something wrong and suggested I improve, i did. That's how open source work. Not everyone here is a master programmer, we are trying to collaborate with ideas.

I don't think it is fair that you have to write things 2 times because someone erases them. People should be more careful when writing over stuff but also that is what pull requests are for so that everyone can review them.

It's an ongoing process, it is not perfect and it's not going to be always the way you want. I think you are very talented and have made great contributions. Even in professional teams people break stuff. So there is no need to get mad over a program.

On the other thread I was agreeing with you that people should submit proof of why their changes are better and I agree that it is not fair for you to write things twice but the project has become very awesome by having more people involved and with a little patience everything is going to be alright.

iperov:
ok

babilio:
@Clorr I took care of conflicts that popped up and retested, this should be good to go

PR #255, opened 2018-03-08
image sort tool and PluginLoader.get_available_models()

iperov:
https://github.com/deepfakes/faceswap/issues/252
usage:
```
python.exe faceswap\sorttool.py -i %WORKSPACE%\data_src\aligned -by similarity
python.exe faceswap\sorttool.py -i %WORKSPACE%\data_src\aligned -by blur
```

PR #251, opened 2018-03-08
Add Improved AutoEncoder model.

acsaga:
Add a new autoencoder model, I call it Improved AutoEncoder (IAE).
In the same training time, this model generates better(sharper and natural-looking)  faces than the original model.

In IAE, autoencoder_A and autoencoder_B use the same encoder and decoder.
Between the encoder and decoder, IAE explicitly adds intermidiate layers for latent varibles.

An (ugly) illustration for the model structure:
image

Issue #208, opened 2018-02-18
Merging w/ multiple faces in frame


enniowatson:
Running into an issue when I try to convert. If there are multiple faces in the frame, Face B gets swapped onto all the faces. Face B also gets swapped onto solo faces in different shots that are not the target face. I'm coming from FakeApp, where any faces deleted from the /aligned folder are skipped during conversion. That doesn't appear to be the case here (though it does appear to skip those files during training). I tried using the filter, with the clearest front view of the target face I had. That resulted in several frames of the desired face getting skipped while several frames of the non-target faces were still swapped. Do I have to manually edit the alignments.json file or is there an easier workaround?

Apollo122:
I raised this issue before but no changes yet. https://github.com/deepfakes/faceswap/pull/155#issuecomment-365079915
I proposed a utility function to check entries in json against data/A and delete the absent ones. But until it is implemented manual work is only way

enniowatson:
Any idea why this hasn't been addressed? Seems like an easy fix, as other versions have this functionality. It's already tedious enough deleting unwanted frames from the /aligned folder. Having to then delete parts of the .json file is a hundred times more annoying, and can cause probably even more problems if you mess up the syntax.

Apollo122:
Because all of our programmers on strike :) no seriously they are all missing so either we wait for them or do it ourselves. I'm c#/Java guy so don't really know python.

ruah1984:
You mean in UK right now, all University staff strike haha.

cojosao:
How difficult would it be to create a standalone script that would parse the alignments file and delete entries that don't have corresponding pictures in A/aligned? I'm no programmer but perhaps this would be a temporary fix until something can be incorporated into the main branch? I'm also in a similar spot since most of what I want to do involves two people in a scene.

oatssss:
Seems simple enough to do in Python, I'll try and do this today.

Apollo122:
Great! Just mind that alignment file is not necessary to be a json file. I think there are couple supported extension. You can see them in script

Apollo122:
Will try this. Can you make the default path for this to "./aligned"? Like in the original scripts? People are used to copy this folder to the folder that contains the photos that they want to convert.

oatssss:
It's in already.

Apollo122:
@cojosao  @enniowatson can you test this? Mind that you have to run extract script again to get the alignment file. Its format has been changed.

RunetX:
[...]

PR #260, opened 2018-03-09
Convert: Exclude deleted faces

oatssss:
Use option `-a` / `--input-align-dir` during conversion to specify the directory of aligned faces to convert. This aligned directory should be the one generated with the extract script. Any faces that were deleted from this directory will be skipped during conversion. By default, the converter will look for an aligned folder at `{input_dir}/aligned` where `{input_dir}` is the value you specified for option `-i` / `--input-dir`.

### DIFF
--- a/lib/cli.py
+++ b/lib/cli.py
@@ -35,7 +35,6 @@ class DirectoryProcessor(object):
         self.create_parser(subparser, command, description)
         self.parse_arguments(description, subparser, command)
 
-
     def process_arguments(self, arguments):
         self.arguments = arguments
         print("Input Directory: {}".format(self.arguments.input_dir))
--- a/lib/cli.py
+++ b/lib/cli.py
@@ -58,6 +58,7 @@ class DirectoryProcessor(object):
             pass
     
         self.output_dir = get_folder(self.arguments.output_dir)
+
         try:
             try:
                 if self.arguments.skip_existing:
--- a/scripts/convert.py
+++ b/scripts/convert.py
@@ -6,7 +6,7 @@ from pathlib import Path
 from tqdm import tqdm
 
 from lib.cli import DirectoryProcessor, FullPaths
-from lib.utils import BackgroundGenerator, get_folder
+from lib.utils import BackgroundGenerator, get_folder, get_image_paths
 
 from plugins.PluginLoader import PluginLoader
 
@@ -27,7 +27,16 @@ class ConvertImage(DirectoryProcessor):
                             dest="model_dir",
                             default="models",
                             help="Model directory. A directory containing the trained model \
-                    you wish to process. Defaults to 'models'")
+                            you wish to process. Defaults to 'models'")
+
+        parser.add_argument('-a', '--input-aligned-dir',
+                            action=FullPaths,
+                            dest="input_aligned_dir",
+                            default=None,
+                            help="Input \"aligned directory\". A directory that should contain the \
+                            aligned faces extracted from the input files. If you delete faces from \
+                            this folder, they'll be skipped during conversion. If no aligned dir is \
+                            specified, all faces will be converted.")
 
         parser.add_argument('-t', '--trainer',
                             type=str,
@@ -119,6 +128,8 @@ class ConvertImage(DirectoryProcessor):
         # Note: GAN prediction outputs a mask + an image, while other predicts only an image
         model_name = self.arguments.trainer
         conv_name = self.arguments.converter
+        input_aligned_dir = None
+        self.input_aligned_dir_path = None
 
         if conv_name.startswith("GAN"):
             assert model_name.startswith("GAN") is True, "GAN converter can only be used with GAN model!"
@@ -129,6 +140,16 @@ class ConvertImage(DirectoryProcessor):
         if not model.load(self.arguments.swap_model):
             print('Model Not Found! A valid model must be provided to continue!')
             exit(1)
+
+        try:
+            self.input_aligned_dir_path = self.arguments.input_aligned_dir
+            input_aligned_dir = self.arguments.input_aligned_dir
+            if input_aligned_dir is None:
+                raise Exception()
+            self.input_aligned_dir = map(lambda path: Path(path), get_image_paths(input_aligned_dir)) # Removes escaped backslashes from paths
+            print(self.input_aligned_dir)
+        except:
+            print('Aligned directory not found. All faces listed in the alignments file will be converted.')
 
         converter = PluginLoader.get_converter(conv_name)(model.converter(False),
             blur_size=self.arguments.blur_size,
@@ -166,6 +187,11 @@ class ConvertImage(DirectoryProcessor):
         except:
             return False
 
+    def check_skipface(self, filename, face_idx):
+        aligned_face_name = '{}_{}{}'.format(Path(filename).stem, face_idx, Path(filename).suffix)
+        aligned_face_file = Path(self.input_aligned_dir_path) / Path(aligned_face_name)
+        return aligned_face_file not in self.input_aligned_dir
+
     def convert(self, converter, item):
         try:
             (filename, image, faces) = item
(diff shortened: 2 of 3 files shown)


iperov:
Cannot understand why need this.

and where is option to disable this check ?

bryanlyon:
I would prefer a tool that removes deleted faces from the alignments.json file since that will make the collection more portable to other devices.  But otherwise, this functionality is incredibly useful, as the filter option is very frustrating.

babilio:
@iperov right now if you you have multiple faces or poorly detected faces and delete them from your aligned folder, when you run convert it will still patch all the faces that were detected when creating the alignments. This produces bad results. So this is helpful. See #208 However I don't think the that it should be on by default and look automatically for /aligned folder.  We should test how this impacts the performance. Right now I have a local merge of this + #253 and it is taking 2 hours to convert 8,000, it took me 40 mins to extract. I am not sure of which of the two could be causing this. I can test more tomorrow night

iperov:
Sorry, but nothing useful.

I will try to explain (sry bad english).

Consider real case of usage.
We extracted 1500 frames from dst video. There are two faces in video.
Faces was extracted in random order such as [0][1][0][1] or [0][1][1][0] 
SO will you delete manually second face in 1500 frames ?? 
Me not. Its waste of time.
Also we cannot apply my sorttool https://github.com/deepfakes/faceswap/pull/255 for dst video, because its renaming files, then files will not match alignments.json.
So better track mask in video editor and cut unwanted faces, or track main face mask and overlay it on original video.

oatssss:
@iperov maybe you can write your sort tool to modify the alignments file with the changes it makes. I was thinking of doing a similar sort, but not as a separate tool, as an actual step in extraction. Eventually I wanted to assign a face into 3 categories: left-profile, front, and right-profile. Since this would be done during extraction, we'd know which of the 3 folders the face would be going into and so can write it into the alignments file.

iperov:
@oatssss , sry I cannot, too complex and hard too test. Also @Clorr tolds to write it as separate tool, so I already rewrited it.
Also there are too many PRs including GAN which totally rework many files. 
GAN PR is bottleneck now for anything new.

oatssss:
@iperov why is the gan PR a bottleneck?

babilio:
@iperov as it is the tool you made doesn't work for convert because any faces you delete after sorting will still get patched on convert.

I am interested in using masks in source video though like you do what video editor would you recommend for this? When i used to use fakeapp i would go over and delete the 1,000s of faces of the other person manually haha

ruah1984:
@oatssss i think @iperov  not means like this. currently we have about 20 pull request here, every people are working different thing to make things happen, and some may have some conflict, don't put it in your heart. we just need to have some one to organized all the request, then will be good. cheers UP

oatssss:
@babilio haha I literally just used iMovie

@ruah1984 (dunno why it didn't tag you) no worries, I'm not sad! I was really just curious why @iperov considers the gan PR a bottleneck. I don't think it's a good thing if contributors feel like their work is being held up by something else. I tried to separate out things I added to the gan PR into separate PRs of their own.

iperov:
@oatssss for example
https://github.com/deepfakes/faceswap/pull/251

Clorr wrote:

> I'm going to test it ASAP. Just a remark, we are moving to a new folder structure in the latest commits. You can check my repo to see how it is now. Recommended structure would be for you:
> 
> Create Model_IAE folder in plugins
> Create init.py like this one
> Put your files in this folder

he refers to new folder structure based on GAN commit, which still not merged.

ruah1984:
@iperov , actually i have try the GAN V2 , the training is good. should be not a big problem, this is the reason why Clorr are using @oatssss GAN PR to run and try it. I think @oatssss still working with some idea and improvement. Don't worry it will merge soon. i not a programmer, i also hope i can help to compile it all , what i can do now is test and feedback only with my "heart" ware(Hardware). I will try Clorr Model_IAE model tonight as well.

oatssss:
@iperov the folder structure change isn't specific to gan, it's really just to have it more organized as more people add models to the plugins directory. You're right though, we should have the change in master already, I'll try and do it tomorrow if nobody beats me to it

Apollo122:
People already tried this PR under the issue and GAN PR. @oatssss if you are comfortable you can merge it.

torzdf:
Just my 2 cents: I think this would be better as a separate tool. In my use case, at least, I do not train on every face discovered in a 25fps+ video, therefore I have no need to keep the extracted faces after generating alignments.json. 

Ideally workflow would be:

- Extract all frames to generate alignments.json
- Manually delete unwanted faces from extracted faces folder
- Run tool to update alignments.json
- Delete all faces from the now unneeded faces folder

The proposed solution requires that faces are kept, when they are no longer required.

This is just my opinion for my use case, and either way it would certainly be useful. Motion tracking and masking (as I currently do) can be fiddly!

bryanlyon:
I disagree on this being a separate tool.  If you don't want to use it you wont provide the command line option.  This being a part of the conversion makes it most convenient as you know that it wont process deleted faces as long as the command line option is there.  If it's a separate tool then you'll have to remember to run that tool

Though I agree that the alignment.json should be modified too (if requested).

However, I would note that I find that a short training session on the faces on the actual video data (after completing general training on pictures excluding the target video) helps to create a better result.

babilio:
@bryanlyon I mostly agree with you the problem is that right now even if you don't provide the command line option it looks automatically for "aligned" folder and it's basically turned on. I might be misunderstanding the commits, so please do tell me if I'm wrong.

I think aligned commonly used folder name by people. So basically this is on by default.  I am not sure that the performance decrease is negligible since it does searching every time for the image on aligned folder. So as it is, with my understanding, I don't agree with the pull request. If it was 100% optional, I would be ok with it. If someone could test performance of convert with this and without this and it is negligible, I'd be ok with this. I can test it later today myself. 

I think this is a very awesome feature, but if I have a video where I don't have to delete faces I don't want the feature on if it adds to convert time.

Again my understanding might be off cause I can be kind of stupid, so let me know if my concern is invalid.

babilio:
As for it being a separate tool the idea is that you can run it to clean up the actual alignment files. This can be useful not only for convert, but to perform extraction again with different detector or options, and for portability. So you could run the tool without having to do convert.

oatssss:
@babilio this looks for an aligned folder in the default location, but if it doesn't find one, it runs with the original behaviour of just converting everything from the input (not aligned) directory. However, if it does find an aligned folder, but it's empty, it won't convert anything since it thinks you deleted everything from the aligned folder.

babilio:
So would the only way to not use this is to have my "aligned" folder named something else. Do you or anyone else have an idea of how this affects performance of convert?

I think the scenario I am thinking more of is like if I have a video of frontal, nicely extracted faces that I don't delete, it is inefficient to verify they exist in the align folder. Specially if it is looking for that folder "under the hood".

I definitely would like hear what others think. And I can run tests to compare performance too.

Apollo122:
@babilio Why would you have a empty folder named "aligned" if you don't want to use this feature? Also I don't think this folder named aligned is created automatically so I don't see the problem

bryanlyon:
Checking for the existence of a folder or a file will be negligible.  Unless you're accessing files over a network, you'll be able to verify hundreds or thousands of files a second without difficulty.  That said, this can be fixed so it only works if you use the -a option.  If you're not interested in the feature, don't include the aligned folder command line option.

babilio:
@Apollo122 It is not empty it is full of well aligned faces that do not need to be checked. 
It is not created automatically but is what newbies would use and if it is slow it would discourage use. 

I am just raising the issue because I ran this pull request among merges with others and it made my convert time really bad (even without seamless). I cannot say this is the culprit but I wanted to bring it up. For all I know it could have been a bad conflict resolution I made between pulls what caused this, so if it seems negligible to you all I don't want to hold back the merge of the PR. Thanks

babilio:
I got home tested this branch by itself. Works as expected and no drop on it/s. Sorry for all the comments. I support this PR

oatssss:
Closing since this was merged alongside #217.

PR #262, opened 2018-03-09
Use k-nn for face filtering

alhoo:
The current implementation of face detection uses a threshold for determining if a face matches the reference image. This commit allows the user to add multiple reference images and multiple negative reference images and then uses the old threshold in addition to K-nearest neighbours method to determine if some negative reference is more similar. This is useful if we have a video of multiple actors that look very similar.

### DIFF
--- a/lib/FaceFilter.py
+++ b/lib/FaceFilter.py
@@ -10,36 +10,43 @@ class FaceFilter():
     def __init__(self, reference_file_paths, nreference_file_paths, threshold = 0.6):
         images = list(map(face_recognition.load_image_file, reference_file_paths))
         nimages = list(map(face_recognition.load_image_file, nreference_file_paths))
-        self.encodings = list(map(lambda im: face_recognition.face_encodings(im)[0], images)) # Note: we take only first face, so the reference file should only contain one face.
-        self.nencodings = list(map(lambda im: face_recognition.face_encodings(im)[0], nimages)) # Note: we take only first face, so the reference file should only contain one face.
+        # Note: we take only first face, so the reference file should only contain one face.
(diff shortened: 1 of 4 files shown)


ruah1984:
may i know what is K-NN?

alhoo:
K-NN is an algorithm that searches for the K number of nearest results when deciding which would be the correct result.

In this case we are looking at K reference images which may be positive (of the actor we are interested in) or negative (some other similar looking actors). If most of the reference images that are close to the face we are looking at are positive, then we keep the face. Instead if there are some negative reference images that are closer to the current face, then we discard the image.

iperov:
there is nothing related to face recognition in changed files, only for face filtering

ruah1984:
thanks @iperov , i thought is something about the Neural network like K-neural network. i really like neural network right now.

alhoo:
Ok I meant the face filtering feature. Filter based on k-nearest neighbours.

alhoo:
A small example on how to use the knn-filtering, if you have example images of personA, personB, personC and personD:

    python3 faceswap.py extract -i video/ -o faces/ -f personA/204635.jpg personA/2015555.jpg -n personB/3266.jpg personC/463.jpg personD/653262.jpg

iperov:
best for me extract all faces, sort them, and delete unwanted.

Yuxishan2:
How to sort them? Is there any skills to sort the images?When I extracted the faces, I found it hard to distinguish the faces of other person's from the ones which I want to extract.

Clorr:
@Yuxishan2 This PR is more made to use one or more image sas a filter faces at extract time. Sorting is done in sorttool (#278 and #254 )

PR #265, opened 2018-03-09
Several SQlite3 based optimizations.

ppmdo:
I propose the use of a sqlite3 database to optimize the faceswap in several ways.

The DB stores the frames processed with the Extract script, and marks the ones where a face was found.

This allows two things:
1. To skip already processed frames based on the DB data, instead of comparing input and output directories.
2. To use the Convert script only on those frames where a face was found.

The DB also stores the frames which have already been converted, so It's also possible to skip already converted frames with the --skip-existing argument.

Finally, I think that for future development, having a DB can ease many things: reporting, web-ui, comparison of results with different types of extractors, etc.

### DIFF
--- a/faceswap.py
+++ b/faceswap.py
@@ -10,6 +10,7 @@ from lib.utils import FullHelpArgumentParser
 from scripts.extract import ExtractTrainingData
 from scripts.train import TrainingProcessor
 from scripts.convert import ConvertImage
+from scripts.db import DBManager
 
 def bad_args(args):
     parser.print_help()
@@ -24,6 +25,8 @@ if __name__ == "__main__":
         subparser, "train", "This command trains the model for the two faces A and B.")
     convert = ConvertImage(
         subparser, "convert", "Convert a source image to a new one with the face swapped.")
+    db = DBManager(
+        subparser, "db", "Manage Sqlite3 database.")
     parser.set_defaults(func=bad_args)
     arguments = parser.parse_args()
     arguments.func(arguments)
--- a/lib/FaceLandmarksExtractor/FaceLandmarksExtractor.py
+++ b/lib/FaceLandmarksExtractor/FaceLandmarksExtractor.py
@@ -182,5 +182,6 @@ def extract(input_image, use_cnn_face_detector=True, all_faces=True, scale_to=20
             landmarks.append ( ((  int(left/input_scale), int(top/input_scale), int(right/input_scale), int(bottom/input_scale) ),pts_img) )
     else:
         print("Warning: No faces were detected.")
+        return None
         
     return landmarks
--- a/lib/cli.py
+++ b/lib/cli.py
@@ -6,6 +6,7 @@ from pathlib import Path
 from lib.FaceFilter import FaceFilter
 from lib.faces_detect import detect_faces, DetectedFace
 from lib.utils import get_image_paths, get_folder
+from lib import db
 from lib import Serializer
 
 class FullPaths(argparse.Action):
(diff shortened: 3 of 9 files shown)


bryanlyon:
A database makes the system a lot more complicated and less portable.  SQLite is not as bad as others, but JSON is not a very big performance cost and is still easy to use.

In fact, from a quick look around, I think JSON is actually more efficient than SQLIte.  In general, the advice online is to stick with JSON unless you need ACID or relational data.  Neither of which is important to our use cases.  Honestly, I'd want to see some concrete reason why we need SQLite before I'd want to use it.

https://stackoverflow.com/questions/8652005/json-file-vs-sqlite-android
https://softwareengineering.stackexchange.com/questions/190482/why-use-a-database-instead-of-just-saving-your-data-to-disk
https://softwareengineering.stackexchange.com/questions/235707/using-a-relational-database-vs-json-objects-for-event-activity-data

Clorr:
Hi, thanks for the PR. I also think the db is a bit overkill for now, but maybe someday it could be useful. However if you want to go that way, you should make an abstraction layer so that info is stored/read the same way, wether we have json or db. If you want more info on abstraction layer, let me know, I can explain more

ppmdo:
Hi. I get the point of the DB overkill. In fact, @Clorr I've already included an abstraction layer in lib/db.py. I wouldn't expect the user to type or use SQL directly, so there are several write/read functions in that file to ease those tasks.

Anyway, I think the benefits of storing the status and results of the extract and convert scripts are worthy. So, If the idea is to stick to JSON, I can refactor the code to do that. What do you think?

To get a clear idea of what I'm intending by storing this data (in JSON or DB), this is what the data looks like after running Extract.py:

id          filename         extracted   face_found  converted 
----------  ---------------  ----------  ----------  ----------
1           output_0767.jpg  true        true        false     
2           output_0710.jpg  true        true        false     
3           output_0028.jpg  true        false       false     
4           output_0002.jpg  true        false       false     
5           output_0713.jpg  true        true        false     
6           output_0701.jpg  true        true        false     
7           output_0026.jpg  true        false       false     
8           output_0020.jpg  true        false       false     
9           output_0008.jpg  true        false       false     
10          output_0715.jpg  true        true        false   

This allows, for example, to know:

1. Which frames failed to be extracted.
2. Which frames have a face. Which afterwards allows to run the convert script **only** in these frames.
3. Which frames have already been extracted or converted. Which allows to run the scripts **only** on the pending ones.

These increases the efficiency of the scripts by avoiding reprocessing of already extracted/converted frames; and by focusing on converting only those frames in which the extractor found a face.

Clorr:
i think your intention is good. The thing is just that I don't want to rely on a db, that's why i prefer we can have the choice of JSON(default) or DB.

Also, what you did in db.py is not really an abstraction layer, it is just an interface (which is nice already). An abstraction layer is like a plugin architecture, it let's you switch between implementation classes seamlessly. We can do that already with your proposal, but we would still end up with some self.conn around and db related code outside db.py

Also, there are some lines that are not related to DB in your PR, like what resides in face_detect.py. This should be avoided because we can end up with big mess without knowing, if you think your modifications are worth it, you can do a separate PR

ppmdo:
@Clorr Thanks for your clarification on the abstraction layer. I think the issue about the modifications being worthy is not up to me; I mean, I can do the coding for the json-db part, but if the benefits I see, are not beneficial for the rest of the users I think we could call this a dead end.

Clorr:
No problem, close this PR if you don't think it is relevant

PR #272, opened 2018-03-11
Add Multi-GPU support

bryanlyon:
I added multi-GPU support to the new model layout.  Currently, Original is not tested (due to OOM on my 2x 4gb 970s).  LowMem is not tested with the current commit due to it not being available since the new pluginloader misses it.

This is essentially the same patch as #256 but modified for the new plugins.

GAN has been tested better with this patch due to it not failing horribly on my system.

Not every NN model has been added to multi-gpu support due to overhead and lack of improvement on small models.  This patch will benefit from larger batches, since a batch gets split across all GPUs.   (64 batch on 2 gpus will actually put 32 pictures on each GPU.)  Right now, this is not done automatically, because due to (slight) ram overhead, it might not properly double with the available memory.

### DIFF
--- a/lib/FaceLandmarksExtractor/FaceLandmarksExtractor.py
+++ b/lib/FaceLandmarksExtractor/FaceLandmarksExtractor.py
@@ -105,25 +105,26 @@ def get_pts_from_predict(a, center, scale):
     c += 0.5
     return [ transform (c[i], center, scale, a.shape[2]) for i in range(a.shape[0]) ]
 
-def initialize(scale_to=2048):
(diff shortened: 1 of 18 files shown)


Clorr:
Should I merge now, or do you want to add multi-gpu in iae before?

bryanlyon:
I could merge my patch for that model into this branch and have both closed at once if you want.

bryanlyon:
I see you've already merged the IAE.  I went ahead and just added it here.  It's ready to merge whenever you are.

PR #278, opened 2018-03-11
sorttool argument changes and added -by face 

iperov:
changed arguments
now valid arguments:
-by blur
-by hist (default)
-by face

'by face' helps sort faces extracted from internet photos

example:

-by hist:
fsviewer_2018-03-11_19-06-12

-by face:
fsviewer_2018-03-11_19-09-43

### DIFF
--- a/sorttool.py
+++ b/sorttool.py
@@ -5,6 +5,8 @@ import operator
 import numpy as np
 import cv2
 from tqdm import tqdm
+import face_recognition
+
 if sys.version_info[0] < 3:
     raise Exception("This program requires at least python3.2")
 if sys.version_info[0] == 3 and sys.version_info[1] < 2:
@@ -28,16 +30,18 @@ class SortProcessor(object):
                              
         parser.add_argument('-by', '--by',
                             type=str,
-                            choices=("blur", "similarity"), # case sensitive because this is used to load a plugin.
+                            choices=("blur", "hist", "face"),
                             dest='method',
-                            default="similarity",
+                            default="hist",
                             help="Sort by method.")
 
     def process(self):        
         if self.arguments.method.lower() == 'blur':
             self.process_blur()
-        elif self.arguments.method.lower() == 'similarity':
-            self.process_similarity()
+        elif self.arguments.method.lower() == 'hist':
+            self.process_hist()
+        elif self.arguments.method.lower() == 'face':
+            self.process_face()
             
     def process_blur(self):
         input_dir = self.arguments.input_dir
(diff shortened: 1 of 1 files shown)


i5L4NDOF5T4BiLiTY:
Works even better than version 1, -by face to check no other faces sneaked in there, then -by blur to get rid of the trash, then -by hist to look over the dataset in all its optimised glory.

Clorr:
Ok to merge this. One request though. Can you put this file in a `tools/` folder and name it a bit more explicitly (something like sort_extracted.py, or something you find relevant)?

iperov:
@Clorr i dont think it good idea, because following this logic, \scripts\ also can be run directly, but they cant

bryanlyon:
This is no different from any other linux program's directory layout; tools/ is implied to refer to self-contained tools that help the project without but other folders are usually utilities or sub-files used by the main program.  That said, a simple catchall added to files in scripts/ could warn users that they're not directly executable.

iperov:
bryanlyon also other directories same warn.

better solution is move all prog to 
src\
then tools\ is ok

Enyakk:
I noticed that the -by face sorting has problems with faces (side profiles, low angles) originally aligned+detected by CNN, resulting in a semi-random order. That is because face_recognition by default uses "hog" for face_encodings, so those "CNN-exclusive" faces fail when re-detected by HOG.

I was unsuccessful to modify face_recognition to engage CNN-mode. Maybe it's possible to re-use the alignments.json or do they not contain the necessary information?

iperov:
@Enyakk 5 point model used in faceencodings

```
def face_encodings(face_image, known_face_locations=None, num_jitters=1):
    raw_landmarks = _raw_face_landmarks(face_image, known_face_locations, model="small")
```

there is no chance to use 68 point model outside the lib face_recognition.

Also I dont know how landmarks transformed to 128D vector for compare

https://github.com/davisking/dlib/blob/master/tools/python/src/face_recognition.cpp compute_face_descriptors

Enyakk:
As a test I successfully enabled the 68 point model by directly putting face_recognition into the same directory as sorttool.py and changing the default predictor in api.py to "large".

`def face_encodings(face_image, known_face_locations=None, num_jitters=1):'
...
'raw_landmarks = _raw_face_landmarks(face_image, known_face_locations, model="large")`

iperov:
@Enyakk and what ? you cant change the lib.

iperov:
@bryanlyon and how to import lib from tools\ ?

Clorr:
@iperov you have to use:
`from ..lib.foo import foo, bar`
You have to add `.` in front of path to go to upper dir

iperov:
you think this is good architecture ? :)

Clorr:
Fair enough for a utility. If you want to make better architecture, you can add a tools.py at the root and make it run with a `tools.py sort`

Clorr:
(and this way your imports will be relative to root folder)

PR #280, opened 2018-03-12
Fix broken multigpu GAN loading.

bryanlyon:
Fix for loading into the multi GPU model when it needs to load/save the original model.

#272 accidentally included a GAN implementation that would not properly load the model causing cumulative corruption over each load/save.  This patch fixes that issue and adds additional safety on saving/loading.

I apologize for the error.  I'd already checked and verified Original, LowMem and IAE but GAN2.0 slipped through due to how it processes the data differently from 1.0 and my rush to add the multi gpu support.  I'll make sure to check better in the future..

### DIFF
--- a/plugins/Model_GAN/Model.py
+++ b/plugins/Model_GAN/Model.py
@@ -116,6 +116,9 @@ class GANModel():
         netGA = Model(x, decoder_A(encoder(x)))
         netGB = Model(x, decoder_B(encoder(x)))
 
+        self.netGA_sm = netGA
+        self.netGB_sm = netGB
+
         try:
             netGA.load_weights(str(self.model_dir / netGAH5))
             netGB.load_weights(str(self.model_dir / netGBH5))
@@ -125,10 +128,8 @@ class GANModel():
             pass
 
         if self.gpus > 1:
-            self.netGA_sm = netGA
-            self.netGB_sm = netGB
-            netGA = multi_gpu_model( netGA , self.gpus)
-            netGB = multi_gpu_model( netGB , self.gpus)
+            netGA = multi_gpu_model( self.netGA_sm , self.gpus)
+            netGB = multi_gpu_model( self.netGB_sm , self.gpus)
 
         return netGA, netGB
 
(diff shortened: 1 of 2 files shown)


Clorr:
Thnaks for testing and providing a solution. I'm just wondering: are you sure you need 2 new vars? Because they seem to reference the same objects as the original vars so maybe the fix is just displacing the `multi_gpu_model` call?

bryanlyon:
Actually, yes, those variables are necessary.  At least for now.  https://keras.io/utils/#multi_gpu_model has the warning of "To save the multi-gpu model, use .save(fname) or .save_weights(fname) with the template model (the argument you passed to multi_gpu_model), rather than the model returned by multi_gpu_model."  In order for this to happen, we need to save the reference to the original model before the multi_gpu_model call.

The Original model avoided this issue by saving the decoder and encoder weights instead of saving the whole model.  I could modify the GAN model to use this method, but it would double the number of files used to save, and would invalidate all old model saves.  In order to avoid this, I could write a save/load module that loads any version of save and converts it.  In addition, I could make the module save data to a single file.  This was beyond the scope of a quick solution fix, since people relying on the GAN model with more than one GPU would have to wait on using the multi GPU support until I built the new module, which would take some time.

bryanlyon:
I also went ahead and fixed the missing Multi GPU support for GAN128.  I hadn't been able to test it (and still can't) but at least it's there so it will run now instead of failing.  I invite anyone with multiple 6gb+ gpus to test it and let me know if there are any issues.  I believe it's likely to work though since it's so similar to GAN.

Clorr:
Ok, that's fine for me. Do I merge now, or do you wait for feedback?

bryanlyon:
I'd suggest merging now, this is bug fixing and doesn't do any feature adds.  (except for GAN128, which just brings it to parity with GAN).

Clorr:
Here we go-oo!

Kirin-kun:
I confirm it's working (at least with 1 GPU)

Issue #233, opened 2018-03-01
Some faces with one eye hair covered can't be recognized


modelsex:
*First THANKS A LOT for all contributors' hard work!
*Always make a compare test after big change, test with same source 1000 pics (kar801 -> kar1800) , compare with FakeApp1.1 & latest faceswap commit 232d931. 
*Test files (this is porn, sorry~) :  https://drive.google.com/open?id=1DKP0Uci97FfriuKcYEbDT2Qp6rKN1h1g

## Expected behavior
Not sure, limitation ? or possible to improve ? 

## Actual behavior
FakeApp1.1 extract rate is 988/1000
faceswap -D cnn extract rate is 943/1000

![comparetest](https://user-images.githubusercontent.com/35594182/36897500-b8a78cc8-1e51-11e8-9af5-d58945203958.png)

Notice that some faces - specially one eye covered by hair can't be extract. Example: kar1086 -> kar1090, these 5 pics can be extract normally in FakeApp, but failed in faceswap. Compare kar1085 with kar1086, no big gap in these 2 pics, just corner of the eye be covered by hair in kar1086.  

## Steps to reproduce
python faceswap.py extract -i D:/project4/data_A1/ -o D:/project4/data_A1/output/ -D cnn

## Other relevant information

- **Operating system and version:** Windows, macOS, Linux 
Windows10
Python3.6.4
CUDA9.0
dlib 19.9
The others env same as requirements-gpu-python36-cuda9.txt

oatssss:
From what commit did you do the extraction from?

modelsex:
Latest commit 232d931

iperov:
this is `dlib cnn` issue
anybody know how to decompile fakeapp v1.1?

Kirin-kun:
There's a program "a.exe" in the alignement subdirectory of FakeApp 1.1


C:\FakeApp\alignment>a.exe -h
RuntimeError: module compiled against API version 0xc but this version of numpy
is 0xb
usage: a.exe [-h] [--one-face] [--all-faces] [--startFrame STARTFRAME]
             [--maxFrames MAXFRAMES] [--fileType FILETYPE]
             [--processor PROCESSOR]
             input_dir [output_dir] [output_file]

positional arguments:
  input_dir
  output_dir
  output_file

optional arguments:
  -h, --help            show this help message and exit
  --one-face
  --all-faces
  --startFrame STARTFRAME
  --maxFrames MAXFRAMES
  --fileType FILETYPE
[...]

PR #281, opened 2018-03-12
Removing extra dlib detector call from Extraction Toolchain -- instead of sending both BGR and RGB images, just send RGB 

kvrooman:
Splitting PR into two to order to simplify discussion...

Current Code Status:
The FaceLandmarksExtractor.py code will issue multiple call to the dlib face detector by first sending an image in BGR format ( referring to the order of the color channels in an image .. blue, green, red ) and then making a second call in RGB format. The BGR image first sent is called ( input_image ) and the RGB image sent second is called ( input_image_bgr ...which is confusing ? )This issue was primarily caused by the non-standard way cv2.imread loads images ( in BGR format unlike all other toolchains )

Background:
The code used to send only the BGR image to dlib detector. This had subpar performance. See https://github.com/deepfakes/faceswap/pull/236  This was fixed by sending the RGB image in addition in a second detector call by inverting the channel order.

Proposed Change:
Dlib documentation states that the face detection model was explicitly trained on RGB images and should only provide satisfactory results with that format. In a first pass of a speed optimization of the toolchain, I want to propose eliminating the first call to the detector with the BGR image and only send the RGB image

Most landmark models and 1adrianb's face_alignment code also only use RGB

Initial Testing:
My limited first testing shows no difference at all in face detection ability or performance in detection. There is a speed-up. I welcome additional testing by the community. I will post additional results on varied test videos shortly.


iperov:
I'm shocked how ppl can ruin the code

kvrooman:
Despite that comment in there, for some reason, both the RGB and the BGR versions of the image are still sent to the detector. See current code below. This is unnecessary as the BGR version is redundant and will perform worse than the RGB version. Parring down the loop to use just the RGB version is good practice and should speed things up?  @ iperov Is there a reason not to?

 input_images = [input_image, input_image_bgr]
for current_detector, input_image in ((current_detector, input_image) for current_detector in dlib_detectors for input_image in input_images):

detected_faces = current_detector(input_image, 0)

kvrooman:
PS - the current code is also mislabeled. As cv2.imread loads all images as BGR, we have to convert it to RGB in order for it to play well with Dlib. At the very least the original code 

input_image_bgr = input_image[:,:,::-1].copy()

should be changed to 

input_image_rgb = input_image[:,:,::-1].copy()

as it is incorrect and misleading

iperov:
totally disagree with this commit.

oatssss:
@iperov I haven't completely looked into the code change, but the counter-argument @kvrooman gave seems completely reasonable. Could you elaborate on why you disagree? Also I AREA performs worse than LINEAR when upscaling, maybe there should be a conditional in there to use the best one for the scenario.

iperov:
> Could you elaborate on why you disagree?

sry no, 
1) too hard explain in english
2) fight for code in every pr is too exhaust me. I better leave project.

bryanlyon:
This PR seems reasonable, 2 questions though.

Have you tested this in a variety of situations?  Does it impact the odds of getting a "hit" on the face match?  I've thought of pruning this area of the code to improve face recognition, I'm not too worried about speed, since I usually have to run it multiple times with different inputs to get it working anyway.  I figured that even if the colors were wrong, that it might actually get a few additional hits this way, even with the colors mixed up, it can still run.

lib/FaceLandmarksExtractor/FaceLandmarksExtractor.py
```diff
@@ -143,12 +143,11 @@ def extract(input_image, detector, verbose, all_faces=True, scale_to=2048):
...
 
     input_scale = scale_to / (w if w > h else h)
-    input_image = cv2.resize (input_image, ( int(w*input_scale), int(h*input_scale) ), interpolation=cv2.INTER_LINEAR)
-    input_image_bgr = input_image[:,:,::-1].copy() #cv2 and numpy inputs differs in rgb-bgr order, this affects chance of dlib face detection
-    input_images = [input_image, input_image_bgr]
+    input_image = cv2.resize (input_image, ( int(w*input_scale), int(h*input_scale) ), interpolation=cv2.INTER_LANCZOS4)
+    input_image = input_image[:,:,::-1] #cv2 loads into BGR format, dlib model is trained on RGB format, switch to RGB
  
     detected_faces = []
-    for current_detector, input_image in ((current_detector, input_image) for current_detector in dlib_detectors for input_image in input_images):
```
oatssss: lmao, the chaining of these for loops has lost me. Why do you remove them here?

kvrooman: it's pretty odd to me too. I understand it to mean that if there are multiple detectors ( i.e. hog and CNN ) and multiple images in input_images( BGR and RGB), loop over every unique combination of the two and send that combo to the current_detector call. the following len break line aborts out of this loop on the first image with found faces and it won't check any remaining detector/image combos. ( if you load the hog model first, this would never check CNN model.... , if you load BGR first, this would never check RGB ) . As Dlib firmly states his model is trained on RGB faces, you want that image first in the list ( now second and mislabeled as bgr). on limited testing, found no difference in number of found faces when eliminating the list of two images and just using only the RGB. testing on several thousand extracts today, so should be able to report back if any.

bryanlyon: This is the part that was running the RGB and BGR though. It was making a list of all detectors, and then the image in RGB and BGR then looping through them each so that every detector was fed both RGB and BGR. That is the whole reason for the PR.

The change is to replace the double loops (one loop on dlib_detectors and one loop on input_images = [input_image, input_image_bgr]). By removing the BGR version, he didn't need to loop on both, so only loops on the dlib_detectors so all detectors are used on the RGB version of the image only.

babilio: Have you guys seen this #236 . Apparently recognition was not good without bgr

kvrooman:
the ironic thing is that input_image_bgr is actually in RGB format and mislabelled. so unless you run the detector with that file, you get the subpar performance.

bryanlyon:
I'm willing to believe that the previous error was a bug if the current testing is showing good results.  I can't test extraction right now, but if anyone else wants to test and respond with their results.  This is more of an "optimization" than a bug, so we can leave it in while we test for a bit.

iperov:
here comparison of face swappers
https://www.deepfakes.club/best-hardware-software-deepfakes/

deepfakesclub wrote: https://www.reddit.com/r/SFWdeepfakes/comments/82btao/portable_python_installs_of_faceswap_and_dfaker/dv94je7/
> It detects more correct faces overall

so if you merge this crap PR, faceswap will detect same amount of faces as FakeApp.

Conclusion:

STOP OFFER CHANGE FUNDAMENTALS OF FACE RECOGNITION WITHOUT 100% PROOFS AND LONG TESTS WITH VARIOUS CONDITIONS

@Clorr please close this pr.

bryanlyon:
None of the articles (or the comment) you reference show any information about the BGR implementation.  Like I said, I would like to see tests that show this actually works as expected and doesn't lower results before we merge, but demanding it be closed simply because you don't think this is correct when initial results say it is inappropriate.

Nobody is saying we have to add this today or without testing, we want to make sure it actually helps, before it's added, but the initial results are promising and it makes sense that sending the BGR might not do well, especially with the CNN mode which is trained for RGB only data.  I understand that you have a concern that this might reduce quality of detection, but if it doesn't, then we should add it.  The only way to know is to test it.   @kvrooman has done some initial testing and shows improved speed with no loss in detection and that alone means that this is worth checking out to see if it's worthwhile.

iperov:
> I would like to see tests that show this actually works as expected and doesn't lower results before we merge

agree. I already wrote: 100% PROOFS AND LONG TESTS WITH VARIOUS CONDITIONS

> sending the BGR might not do well

disagree. Sending BGR and RGB detect more faces.
https://www.deepfakes.club/best-hardware-software-deepfakes/ https://www.reddit.com/r/SFWdeepfakes/comments/82btao/portable_python_installs_of_faceswap_and_dfaker/dv94je7/
because there is only one thing in FaceSwap make difference from FakeApp is rgb-bgr scan.

Also FaceSwap extract performance better than FakeApp, thx for my pytorch port. Nobody need more performance.
Why we spending our time and energy to discuss this suxx ?

@Clorr please close it

babilio:
Right now the level of detection is really good, so I'd be hesitant unless this PR can really show that it has the same detection capabilities. If it ain't broke don't fix it kind of deal. It would be helpful when people post PR's to offer some actual quantifiable gains and losses, like found x faces out of y amount of images in z minutes, vs a, b, c before. @iperov had his reasons for this and is opposing, so the onus of proof should be on the changes being made. He doesn't need to find papers to justify that his work is working cause it is working. If someone is trying to make it faster then show that.

babilio:
Yes for this type of changes there should be testing and proof. Ideally from person doing the PR. Specially cause @iperov wrote the original code and is opposing it. The level of detection is really good right now, so he shouldn't need to produce papers about BGR implementation. The new changes should show improvements and drawbacks clearly if any. This is clearly discouraging @iperov from all the work he did porting from pytorch, if there are clear benefits lets show'em and go for it but otherwise...

When it is a plugin that's optional and does it's own thing that is fine but when it's core code we can't just go off of theory.

kvrooman:
@iperov You can be a bit more cordial, rather than just saying I suxx. I understand that I'm new to this and that you're passionate but I have the same goal as you ....

I am testing this further by the way. I posted this PR now at this stage for comment and review ( like I thought I was supposed to ? ) I don't understand why you're yelling to close this item when I've tried to logically explain why the code likely doesn't need some extra lines of code and extra calls to the detector to get the same performance/found faces. Sorry if I offended you by editing your implementation but I thought this was the whole point of open source code...?  

Btw, this was just an intermediate step on my way to vectorizing the whole extract chain as the dlib detector can take vectors of images as inputs and can process them as batches...

bryanlyon:
I am doing testing as well.  Testing on one person's system is helpful, but less so than running it on multiple sets and on multiple platforms.

Obviously @iperov is and @babilio are right, this needs to show improvement to be accepted, but a PR like this is the best way to show a wider audience that it is available and get that test data.

I would also like to work with you on the extract.  My focus is on multi-gpu support and right now, the extract chain is the hardest part to add that support to.  (except for CNN only, which I could add relatively easily, but it'd take me touching a half dozen files to get working in the same way as train/convert).

babilio:
There is a problem where dlib runs out of memory really easily. So I'm not sure but batch processing could run into that. Keep in mind that the project is supposed to be usable with people with lower end cards.

Clorr:
This PR adresses details that are beyond my understanding. What I can see here is that the proposed code is somewhat clean and should be respected. My thinking is:
- If this messes up things completely this should be seen and the PR won't be merged.
- If it highly improves things, this should also be seen and this could be merged.
- If the difference is very low, it does not deserves so much hard discussion, and we can talk about this calmly.

Also, @iperov, you should consider that if you say things that are too strong, it will deserve your credibility (although I'm sure you know what you are saying as this was discussed in #233 and it generated the #236 PR). So I'm sure this can be talked about calmly and with proper arguments... 

At some point #233 offers an interesting test case. As @iperov  showed that his modification have improved things, we should at least test that this PR does not bring regression on that test case.

lib/FaceLandmarksExtractor/FaceLandmarksExtractor.py
bryanlyon: #236 is a not exactly proof that BGR helps, since it was only BGR going to the detector before that patch and patch added RGB. The patch obviously improved the results, but the question now is if keeping the BGR is necessary. This patch removes the BGR and just send the RGB to the detector, so it needs some testing, but shows promise at this time.

bryanlyon:
Yes, however, before #233 we were using BGR only, which is not what CNN or HOG were designed for.  @iperov's change, to use RGB and BGR was helpful.  This PR is just saying that using RGB may have been all that was necessary and that using both RGB and BGR is unnecessary.

Here is an initial test of my own test data.

The command I ran is:
python3 faceswap.py extract -i content/ -o Testdata/bgr/ -j 20
and 
python3 faceswap.py extract -i content/ -o Testdata/nobgr/ -j 20

Without patch:
3271/3271 [21:11<00:00,  2.57it/s]
Alignments filepath: /superm/data/Projects/Valentina_Nappi/NFBusty Snatched/Fuck2/alignments.json
Writing alignments to: /superm/data/Projects/Valentina_Nappi/NFBusty Snatched/Fuck2/alignments.json
-------------------------
Images found:        3271
Faces detected:      3271
-------------------------
Done!
ls Testdata/bgr/ | wc
    965     965   26055
With patch:
100%| 3271/3271 [18:42<00:00,  2.91it/s]
-------------------------
Images found:        3271
Faces detected:      3271
-------------------------
ls Testdata/nobgr/ | wc
    965     965   26055

The detection was EXACTLY the same.  The exact same number of faces were found and each face was identical.  The folders were identical.  The only difference was in execution time. 21:11 without the patch and 18:42 with the patch.

Obviously, this is not complete.  I didn't test CNN (I don't have GPUs available right now) or a comprehensive set of data (literally impossible), but these results tell me that this PR has some merit, at least for further testing.  But I think it's likely that just using RGB instead of RGB and BGR makes sense.

kvrooman:
Testing with the 1,000 images found in https://github.com/deepfakes/faceswap/issues/233

Current Master Code:
py faceswap.py extract -i E:\shots\testing -o E:\shots\testing\faces_master -D cnn
1.  985 faces found in 1,000 frames   3:52 time --- 4.31 it/s
2.  985 faces found in 1,000 frames   3:53 time --- 4.29 it/s

Proposed PR Code:
py faceswap.py extract -i E:\shots\testing -o E:\shots\testing\faces_PR -D cnn
1.  985 faces found in 1,000 frames   3:33 time --- 4.68 it/s
2.  985 faces found in 1,000 frames   3:36 time --- 4.61 it/s

No difference in faces. Slight speed increase

iperov:
you using bad test data.
Use video with high amount of faces and high angles near 90 deg.

bryanlyon:
My test data was with video with 90 degree faces.  I've now tested over
6000 images extracted with almost identical results.  BGR had one more
false positive than the RGB only where a blue-green curtain got recorded as
a face.

RGB only was consistently (slightly) faster.

I think this PR is proving that if for nothing else, it makes the code
clearer and a tiny bit faster (~10%).

kvrooman:
Doing a diff/compare on the alignment .json files shows ~200k differences for the test example I used (985 faces) Looks like mostly single digit shifts ( i.e. 146 -> 147 or 92-> 91 ) but there are also notable shifts in the bounding boxes for some faces. Interesting... I'll need to compare accuracy and try to get a copy of the gated Fan et al. paper, but most of the other academic papers use RGB as input to the alignment models...

kvrooman:
Doing a comparison to 1adrianb and his face_alignment scripts ( see his code below )

'''def get_landmarks(self, input_image, all_faces=False):
        if isinstance(input_image, str):
            try:
                image = io.imread(input_image)
            except IOError:
                print("error opening file :: ", input_image)
                return None
        else:
            image = input_image

        detected_faces = self.detect_faces(image)'''
Later on the same image is passed to the get_predics function.

He uses scimage (io.imread) to do the imread of the images. The scimage documentation states that it loads images as RGB format unlike the cv2.imread which loads in BGR format.  If we don't invert the channels before sending the resulting image to the face detector and the landmark predictor, this is an implementation change from 1adrianb....I can evaluate the differences in accuracy if we deviate from that implementation as we currently do, but I strongly suspect that using 1adrianb's code will be better?

http://scikit-image.org/docs/dev/api/skimage.io.html#skimage.io.imread

torzdf:
@iperov 

> you using bad test data.
> Use video with high amount of faces and high angles near 90 deg.

You cannot possibly know this,

I apologise for putting this here, but there is no appropriate place on github.

You cannot go around attacking everyone who wants to make any amendments to your code. What has been suggested here is entirely reasonable and is worthy of further investigation. It is neither an attack on you nor the excellent code that you have implemented. 

Downvoting and throwing tantrums whenever anyone wants to make changes only serves to discourage and drive people away from the project. Most people who work on projects like these are amateurs and are looking to improve themselves as well as the code base. No one comes here with the intention of making things worse. If you do not agree with a change, the onus is on you to articulate why rather than ranting at the contributor. Ideally then both parties walk away having learned something.

The nature of open source is that changes will be proposed, some will be accepted, some will be rejected, you will agree with some changes, you will disagree with others. Some of these changes will affect code that you have worked on. However, this is not your code. It is the project's code and it belongs to anyone and everyone who wants to work on or fork the project. This happened as soon as you submitted code to the project.

If this isn't agreeable to you, then maybe open source is not for you? Or perhaps you could create your own fork of the project where you have more control over what is or isn't accepted. Alternatively you can understand that anyone who submits code wants to improve the project. They won't always succeed, but everyone will appreciate feedback given in a constructive manner. I would ask that you try to take a more measured approach in future.

Back on topic, I plan to perform a sizeable amount of extractions today, so I will run with and without this PR and report back my findings.

Clorr:
@kvrooman I think splitting your PR is a good move. Did you submit another one? or is the 2nd part dependent of this one?

kvrooman:
Tested 10,000 frames from multiple movies containing multiple scenes with differing lighting, zoom levels, poses, and profiles shots.

**Current Master Code:**
py faceswap.py extract -i E:\shots\testing -o E:\shots\testing\faces_master -D cnn

6,599 faces found in 10,000 frames --- 37:49 time --- 4.41 it/s
manually sorted through and found 65 false positives

**Proposed PR Code:**
py faceswap.py extract -i E:\shots\testing -o E:\shots\testing\faces_PR -D cnn

6,656 faces found in 10,000 frames --- 30:22 time --- 5.49 it/s
manually sorted through and found 75 false positives

false positives are mostly zoomed out shots with a whole person in the frame of the picture.
Haven't evaluated landmark / alignment file yet but assume there will also be minor adjustments as mentioned previously.

torzdf:
Ok, I ran some tests, and got some fairly unexpected results, to be honest. On the whole I would say that this is a positive change, but I would like to see some more results, because this looks fairly inconclusive to me right now. I can't help but think there may be some other factors at play in my set up, but here they are anyway, run with:
` -D cnn -r on`

```
                |Time                   |Frames
Title   Frames  |Old     New    Diff    |Old    New     Diff
psp1    3996    |24:39   22:35   2:04   |3959   3985    26 
psp2    2122    |12:45   19:04  -6:19   |1959   1962    3
plow    4496    |30:43   25:23   5:20   |3903   3904    1 
sb      552     |03:06   03:07  -0:01   |552    552     0
----------------|-----------------------|------------------
TOTAL   11166   |71:13   70:09   1:04   |10373  10403   30
----------------|-----------------------|------------------
```

kvrooman:
Regarding the more faces found... the current code will abort the code loop below and not evaluate the RGB image in the detector if "any" faces are found in the BGR image. I also found more faces in my test...

Potential scenarios that result in more faces found ( As I strongly suspect the RGB detector is more robust and accurate and 3 people have reported the PR finding more faces, one of the following is true )
1. The detector finds a foot in the BGR image and thinks its a face and quits the loop. The RGB detector would have found a foot and a face or just a face.
2. The detector finds a frontal face in the BGR image and quits the loop. The RGB detector would have found a frontal face and a profile face.
3. The detector finds a normal sized face in the BGR image and quits the loop. The RGB detector would have found a normal sized face and a tiny size and a large face.
4. The detector finds a face of Person A in the BGR image and quits the loop. The RGB detector would have found a face of Person A and Person B.

```
for current_detector, input_image in ((current_detector, input_image) for current_detector in dlib_detectors for input_image in input_images):
        detected_faces = current_detector(input_image, 0)
        if len(detected_faces) != 0:
            break
```

You can isolate for this test by flipping the order of the image list from

```
input_images = [input_image, input_image_bgr]
```
to
```
input_images = [ input_image_bgr, input_image]
```
Once again the "input_image_bgr" is the image actually in RGB format...

babilio:
Thanks for the through testing!

kvrooman:
Just flipped the single line in the current master code
```
input_images = [input_image, input_image_bgr]
```
to
```
input_images = [ input_image_bgr, input_image]
```
and re-ran on my test set

py faceswap.py extract -i E:\shots\testing -o E:\shots\testing\faces_reverse_order -D cnn
6,663 faces found in 10,000 frames --- 35:23 time --- 4.71 it/s

So the BGR image does provide some small value...a 0.105% increase in faces over just the RGB alone. Could change based on dataset, of course. Interestingly enough the 7 faces obtained above in addition from running just the RGB alone were shots of someone in very reflective sunglasses ( i.e. green/blue blobs over their eyes )

**so the choices we have**
1.current code ( run BGR first, then quit if face found, else run RGB )
       6,599 faces found in 10,000 frames --- 37:49 time --- 4.41 it/s 
2.PR code ( run just RGB )
       6,656 faces found in 10,000 frames --- 30:22 time --- 5.49 it/s
3.swap image_list order in current code (( run RGB first, then quit if face found, else run BGR )
       6,663 faces found in 10,000 frames --- 35:23 time --- 4.71 it/s

I'm biased to 2 and the speed-up but I'll defer to the community's thoughts.

Clorr:
@kvrooman thanks for testing this in details. I don't think this subject deserves so much work though. AFAIK [the source of dlib specifies RGB input](https://github.com/davisking/dlib/blob/09f36dbf4771947c01514a160c70d8a6d1a02aa7/tools/python/src/cnn_face_detector.cpp#L37
) which was already what was done before, but with bad naming. `cv2.imread` [loads in BGR](https://www.learnopencv.com/why-does-opencv-use-bgr-color-format/) and the revert of channels does a RGB. And on that, I agree with your PR.

On the question whether we should loop over multiple channels or not, I think that this is not **needed** but can still help in some cases. My **intuition** on this is that, whatever the channels order, you will have a prediction, and mixing the channel will give you different result on edge cases.

My conclusion is that, by respect for the original work, we should keep the 2 channels, but named correctly. If the speed increase really matters to you, I would suggest to handle this as an option such as --detector-channels, where we can specify RGB, BGR, RGB+BGR. Also this way, anyone who thinks his config works best can still use it.

kvrooman:
@ Clorr, sounds alright. Fine with the compromise of switching the order of the RGB and BGR image list and sending both to the detector. ( also cleaning up the naming ). I'll modify the PR tonight and resubmit

Clorr:
@kvrooman there is a conflict here because `plugins/Extract_Crop.py` was removed in a previous commit. Anyhow this file did not need an update. Can you update the PR so that I can merge?

kvrooman:
Should be good

iperov:
Conflicting files
lib/FaceLandmarksExtractor/FaceLandmarksExtractor.py

PR #285, opened 2018-03-12
Output Sharpening Added

JayantPythonLover:
Two new types of output sharpening methods have been added.

One that deals with a Box Blur method and the other with a Gaussian Blur method.

Box Blur method can be called using argument '-sh bsharpen' --- This method is not dynamic and can produce strong sharpening on your images. Sometimes it can yield great results and sometimes entirely the opposite.

Gaussian Blur method can be called using argument '-sh gsharpen' --- This method is dynamic and tries to adjust to your data set. As a result, while the sharpening effect might not be as strong as bsharpen, it is bound to produce a more natural looking sharpened image.

By default the parameter is set to none which will not run any sharpening on your output.

Preview (quality deteriorated by the GIF but you get the idea):

test

----

As requested I did not another example. This one was trained on a FRESH Original Model with just a data set related to the output I wanted.

The loss values were at 0.013 when I stopped the training and made the conversion. Both Seamless Clone and Match Histogram are enabled across all iterations.

sharpen

In this example it might seem like the Box Sharpen is performing much better than Gaussian but I'd say it is otherwise. The original footage also has a certain amount of quality loss and blur. So the result of the Gaussian Blur is more accurate.

The Box Sharpen however pushes it a notch further and makes it really sharp (with some extra noise) ..which does make it look better in this case (might not in every case).

Check out video examples below. You'll get what I mean.

Left: No Sharpen | Right: Box Sharpen

https://i.imgur.com/Q9ZYlZG.mp4

Left: No Sharpen | Right: Gaussian Sharpen

https://i.imgur.com/sA7NPKF.mp4

---

Which one to go for? The depends on your Dataset and personal preference.

### DIFF
--- a/plugins/Convert_Masked.py
+++ b/plugins/Convert_Masked.py
@@ -6,7 +6,7 @@ import numpy
 from lib.aligner import get_align_mat
 
 class Convert():
-    def __init__(self, encoder, trainer, blur_size=2, seamless_clone=False, mask_type="facehullandrect", erosion_kernel_size=None, match_histogram=False, **kwargs):
+    def __init__(self, encoder, trainer, blur_size=2, seamless_clone=False, mask_type="facehullandrect", erosion_kernel_size=None, match_histogram=False, sharpen_image=None, **kwargs):
         self.encoder = encoder
         self.trainer = trainer
         self.erosion_kernel = None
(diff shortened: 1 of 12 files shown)


bryanlyon:
I support this PR.  This is a well developed change and clearly demonstrated, while I don't need it myself, I can see definite benefits for others.  I appreciate that you made it an option and built in the context of the architecture.

Apollo122:
Thanks for the PR @JayantPythonLover . Could you add more output pictures? How does it look with other options enabled such as seamless clone, histogram match?

JayantPythonLover:
@Apollo122 The examples shown above are with Seamless Clone and Match Histogram.

I always use those options so I made sure that this works without issues with both of those options enabled.

JayantPythonLover:
Added another example with a different output video and a freshly trained model.

Also added comparison footages of No Sharpen and the two sharpen methods so you can see what can be achieved.

Clorr:
Nice PR, let's go

Issue #287, opened 2018-03-13
GAN128 does not work with the multi_GPU_model


bryanlyon:
GAN128 does not support the multi_GPU_model

## Expected behavior

Running with multiple GPUs and the matching -g --spus command line option it should run the GAN128 plugin spread across all GPUs using the Keras multi_gpu_model which automates much of this task.

## Actual behavior

The current build gives the following error whenever attempting to use -g > 1
('The name "model_4" is used 2 times in the model. All layer names should be unique. Layer names: ', ['input_6', 'lambda_1', 'lambda_2', 'model_6', 'model_4', 'model_4'])

Error occurs BEFORE the GPU is activated (happens as soon as the multi_gpu_model is initialized.
## Steps to reproduce

Attempt to run GAN128 with -g 2 or more

## Reasons it can't be fixed yet.

Right now, Keras mutli_gpu_model cannot support multiple output layers.  I'm opening this issue as a reminder and explanation as to why this problem cannot be solved.  In order to accommodate this, I will be issuing a "fix" that blocks GAN128 from multi gpu for the time being.

https://github.com/keras-team/keras/issues/8962

## Other relevant information

- **Operating system and version:**
Windows and Linux 
- **Python version:**  3.5.2 and 3.6.4
- **Faceswap version:** 3f04e8c https://github.com/bryanlyon/faceswap/commit/3f04e8cd06e1816e6aa87f3826ebb919cfa983b2 
- **Faceswap method:** multi GPU

PR #288, opened 2018-03-13
Fix gan128

bryanlyon:
This is a fix to GAN128 to patch out the multi gpu until Keras updates their upstream issue (as mentioned in #287 )

In addition, while I was debugging, I fixed the exception handling in train.py so that it properly passes errors that it doesn't catch..


plugins/Model_GAN128/Model.py
```diff
@@ -42,7 +44,7 @@ class GANModel():
 
     def __init__(self, model_dir, gpus):
         self.model_dir = model_dir
-        self.gpus = gpus
+        self.gpus = 1 #See https://github.com/deepfakes/faceswap/issues/287
```
Clorr: I'd prefer you let the original line and put below it an assert, with a failure message pointing to the issue you mention in the comment

scripts/train.py
```diff
@@ -178,7 +178,7 @@ def processThread(self):
                 print('Saving model weights has been cancelled!')
             exit(0)
         except Exception as e:
-            print(e)
+            raise e
```
Clorr: Somehow I agree that the print(e) is not convenient as it make a more readable error message but that masks the details. However I would be prefere this to be made in another PR

bryanlyon:
Sure, I'll send another PR and remove it from this one.

bryanlyon:
I'll do an assert instead of just doing single.  I'll update this soon.

bryanlyon:
I've made the changes as requested.  If there are not other concerns, this is ready to merge.

PR #292, opened 2018-03-13
Converted LowMem model to the new structure

seranus:
Converted LowMem model to the new structure, removed the old LowMem model file


Clorr:
Thanks for the PR!

Note that previous LowMem was using files from Original plugin. However I'm fine with this PR because it makes sure that a modification on Original (which shouldn't happen normally) won't break the LowMem. If everyone is fine with that, we can merge it without further discussion

bryanlyon:
I agree on principle, I haven't checked this PR, but if it works, I support
it.  My only concern is that maybe low memory should be a configuration on
original instead of its own model.  That is something that needs a better
architecture to do though so this is a good solution for now.

Clorr:
@bryanlyon I understand your point, but it is quite different than a multi_gpu_model. Here we modify the layers structure so it is not so generic. 

We could go in a parameter for layers as FakeApp did, but I think it will lead to problems as a model can't be reloaded if it does not have the same layer configuration so people may lose their work by thinking they can play freely with params. The question is also the same for ENCODER_DIM. If we make this a param some people who don't understand the importance of the param they will loose their saved models. 

I prefer that people advanced enough start doing manual modifications and endorse the responsibility of it than adding a param and having issues that says that one can't reload his model and has to be told that he lost it's work

Clorr:
One thing though is that I would like to modify the name of the models so they are different than the original. I'm thinking about something with a prefix like for IAE

bryanlyon:
I actually have been thinking about this, there is a way to load the model
from the folder no matter how the model is structured, we just need to
build a mechanism to save the whole model and reload it.  It's obviously a
big step though, so it'll probably be a good option for a project.

seranus:
For clarity reasons we could add something like an AE prefix to Original and all modifications of it like AE_Original, AE_LowMem, AE_Intermediate so users would know it's a similar model

ruah1984:
We have IAE ,  VAE-GAN , GAN, XGAN, WGAN from others (Shaoanlu). prefix the name will be better.

bryanlyon:
I agree with the changes of model names, but not with this PR and I think it shouldn't be done until we have a model system like my proposed project in the project tab.

Since Clorr has said this is good and my check shows it all works I will merge it.

PR #293, opened 2018-03-14
Fixed error handling in train.py

bryanlyon:
Fixed the error handling in train.py so it doesn't swallow tracelogs.

This is an improvement to follow Python best practices of raising any exceptions that aren't handled so that tracelogs don't get lost.

### DIFF
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -178,7 +178,7 @@ class TrainingProcessor(object):
                 print('Saving model weights has been cancelled!')
             exit(0)
         except Exception as e:
-            print(e)
+            raise e
             exit(1)
 
     def set_tf_allow_growth(self):
@@ -200,4 +200,4 @@ class TrainingProcessor(object):
                 cv2.imwrite('_sample_{}.jpg'.format(name), image)
         except Exception as e:
             print("could not preview sample")
-            print(e)
+            raise e


Clorr:
2 things here:
- If we make a move, we should do it in every place it would help so we should handle other scripts
- I would do a `print("You encountered an error : {} . See below for details" % e)` or something like that to help not experienced users to have the basic message

bryanlyon:
As to your first concern, I agree we should make sure all code follows the proper pattern.  This patch fixes train.py since train.py is the only place this anti-pattern is used.

As to your second concern, I don't print a message, since the traceback already makes it very clear that it is an error message and what exactly the error is.  In fact, printing a separate error message would add no data because by it's nature, the traceback already includes the error and it's details.

Here is an example caused by a simple typo:

Exception in thread Thread-1:
  File "c:\python\python36\Lib\threading.py", line 916, in _bootstrap_inner
    self.run()
  File "c:\python\python36\Lib\threading.py", line 864, in run
  File "C:\Users\Documents\GitHub\faceswap\scripts\train.py", line 181, in processThread
    raise e
  File "C:\Users\Documents\GitHub\faceswap\scripts\train.py", line 147, in processThread
    model = PluginLoader.get_model(trainer)(get_folder(self.arguments.model_dir), self.arguments.gpus)
  File "C:\Users\Documents\GitHub\faceswap\plugins\Model_GAN\Model.py", line 54, in __init__
    self.netGA, self.netGB = self.build_generator()
  File "C:\Users\Documents\GitHub\faceswap\plugins\Model_GAN\Model.py", line 112, in build_generator
    encoder = Encoderr()
NameError: name 'Encoderr' is not defined

This is an error that would be squashed by the current handling and would only return:

NameError: name 'Encoderr' is not defined

Note that the full traceback is just as clear as the print, but includes all the exact information needed to fix the problem right away.  Other errors could be far more insidious such as.

PermissionError: [Errno 13] Permission denied
or 
io.UnsupportedOperation: not writable

What was denied?  Without a traceback we'd have to hunt to find this problem ourselves.  Traceback is a deliberate part of Python and it's error handling.  Actually changing that may cause confusion to anyone who has seen a Python error before.

Printing added information for the user is one thing, but trying to catch an exception here to do that is not the right time since you'll not be able to actually give them any more useful information (except MAYBE a link to the github to report problems) and will only make it more difficult to debug and understand.

Clorr:
Both convert and extract mask the exception, but I admit that if we did a raise here, it would stop completly the process and may be annoying as user wouldn't be able to process other valid items.

My second proposition was not to say to print a message instead of raising, it was more adding a message before raising. I was suggesting that because many users are scared by tracebacks and don't know which part to read. I agree that the error message is already at the end of the traceback, but newbies don't see it. 

However in my proposition, I agree that users would have to scroll to see the "user friendly" message and many won't do it as they will see the traceback and won't see anything else.

bryanlyon:
Both Convert and Extract do capture exceptions, but those are exceptions whose results are known and so should not stop all execution.  Instead they provide the basic information such as 'Failed to extract from image: {}. Reason: {}'.format(filename, e) which will say that this particular image was not converted and give the error it got for it.  (which will usually be not found or permission) and should give all the information needed to solve for it.

train.py was capturing all exceptions in the whole model and/or trainer and swallow them.  The difference in implementation here is huge, the ones in convert and extract are properly capturing and dealing with the exception, train.py was only intending to capture KeyboarInterrupt (and in a lesser sin, the preview sample) but was swallowing all other exceptions.

I agree that all of our exception handling could be improved, but for now, these train.py exceptions are the only ones causing problems with debugging.

PR #298, opened 2018-03-15
Adding tools.py as main script for using tools, as well as integrating all feature requests from #255 and #278

AbysmalBiscuit:
Add tools.py command and control script for use as the main interface for various tools. The structure and approach is the same as faceswap.py

Add many new features to tools/sort.py: various new sorting methods, grouping by folders, logging file renaming/movemeng, keeping original files in the input directory and improved cli options documentation. Argument parsing has been re-written to inteface with tools.py

Add __init__.py empty file in tools directory for python to register it as a module so that sort.py and future tools can be easily imported.

### DIFF
--- a/tools/sort.py
+++ b/tools/sort.py
@@ -9,6 +9,7 @@ from tqdm import tqdm
 import face_recognition
 from shutil import copyfile
 import json
+import re
 
 
 if sys.version_info[0] < 3:
@@ -19,12 +20,17 @@ if sys.version_info[0] == 3 and sys.version_info[1] < 2:
 
 class SortProcessor(object):
     def __init__(self, subparser, command, description='default'):
+        self.arguments = None
+        self.changes = None
         self.parse_arguments(description, subparser, command)
 
     def parse_arguments(self, description, subparser, command):
         parser = subparser.add_parser(
                 command,
-                help="This command lets you sort images using various methods.",
+                help="This command lets you sort images using various methods."
+                     " Please backup your data and/or test this tool with a "
+                     "smaller data set to make sure you understand how it"
+                     "works.",
                 description=description,
                 epilog="Questions and feedback: \
                         https://github.com/deepfakes/faceswap-playground"
@@ -41,16 +47,18 @@ class SortProcessor(object):
                             default="__default",
                             help="Output directory for sorted aligned faces.")
 
-        parser.add_argument('-g', '--grouping',
+        parser.add_argument('-f', '--final-process',
                             type=str,
                             choices=("folders", "rename"),
-                            dest='grouping',
+                            dest='final_process',
                             default="rename",
-                            help="'folders' sorts input the files them into\
-                                  folders.\
-                                  'renaming' renames the input files to group\
-                                  them.\
-                                  Default: rename")
+                            help="'folders': files are sorted using the "
+                                 "-s/--sort-by method, then they are "
+                                 "organized nto folders using the "
+                                 "-g/--group-by grouping method. "
+                                 "'rename': files are sorted using the "
+                                 "-s/--sort-by then they are renamed."
+                                 "Default: rename")
 
         parser.add_argument('-t', '--ref_threshold',
                             type=float,
@@ -63,7 +71,7 @@ class SortProcessor(object):
                                   is. \
                                   For face 0.6 should be enough, with 0.5\
                                   being very discriminating. \
-                                  For face-cnn 17 should be enough, with 12\
+                                  For face-cnn 7.2 should be enough, with 4\
                                   being very discriminating.\
                                   For hist 0.3 should be enough, with 0.2\
                                   being very discriminating.\
@@ -108,25 +116,36 @@ class SortProcessor(object):
                                   '--log-file', then a 'sort_log.json' file \
                                   will be created in the input directory.")
 
-        parser.add_argument('--log-file',
+        parser.add_argument('-lg', '--log-file',
                             dest='log_file',
                             default='__default',
                             help="Specify a log file to use for saving the\
                                   renaming or grouping information.\
                                   Default: sort_log.json")
 
-        parser.add_argument('-by', '--by',
+        parser.add_argument('-s', '--sort-by',
                             type=str,
                             choices=("blur", "face", "face-cnn",
                                      "face-cnn-dissim", "face-dissim", "hist",
                                      "hist-dissim"),
-                            dest='method',
+                            dest='sort_method',
                             default="hist",
-                            help="Sort by method.\
-                                  When grouping by folders face-cnn-dissim and\
-                                  face-dissim default to face, \
-                                  and hist-dissim defaults to hist.\
-                                  Default: hist")
+                            help="Sort by method. "
+                                 "Choose how images are sorted. "
+                                 "Default: hist")
+
+        parser.add_argument('-g', '--group-by',
+                            type=str,
+                            choices=("blur", "face", "face-cnn", "hist"),
+                            dest='group_method',
+                            default="__default",
+                            help="Group by method. "
+                                 "When -fp/--final-processing by folders "
+                                 "choose the how the images are grouped after "
+                                 "sorting. "
+                                 "Default: non-dissim version of "
+                                 "-s/--sort-by method")
+
         parser = self.add_optional_arguments(parser)
         parser.set_defaults(func=self.process_arguments)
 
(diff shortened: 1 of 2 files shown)


Clorr:
Seems fine, is everyone ok for merge?

JayantPythonLover:
Very nice PR. Fits well into my work flow. And the good thing is it has the options to retain the file names which is important because that info is used in the alignment data.

Seems to have no issues so far with the rest of the code.

bryanlyon:
I see no problem with this PR.  I am not sure it encourages correct development though, since many people will use it solely for import and lead to imports not being where they are used.

iperov:
--grouping
set default - rename

tools/sort.py
```diff
@@ -6,127 +7,586 @@
...
+        parser.add_argument('-g', '--grouping',
+                            type=str,
+                            choices=("folders", "rename"),
+                            dest='grouping',
+                            default="folders",
+                            help="'folders' sorts input the files them into\
+                                  folders.\
+                                  'renaming' renames the input files to group\
+                                  them.\
+                                  Default: folders")
```
iperov: default: rename

tools/sort.py
```diff
@@ -6,127 +7,586 @@
...
+
+        elif self.arguments.grouping.lower() == 'rename':
+            if self.arguments.method.lower() == 'blur':
+                self.process_blur()
+            elif self.arguments.method.lower() == 'face':
+                self.process_face()
+            elif self.arguments.method.lower() == 'face-cnn':
+                self.process_face_cnn()
+            elif self.arguments.method.lower() == 'face-dissim':
+                self.process_face()
```
iperov: process_face_dissim

tools/sort.py
```diff
@@ -6,127 +7,586 @@
...
+            elif self.arguments.method.lower() == 'hist':
+                self.process_hist()
+            elif self.arguments.method.lower() == 'hist-dissim':
+                self.process_hist()
```
iperov: process_hist_dissim

iperov:
and where is face-cnn-dissim ?

AbysmalBiscuit:
@iperov I hadn't realized you also wanted a face-cnn-dissim, I'll add it in.

tools/sort.py
```diff
@@ -6,127 +7,586 @@
...
+
+        print ("Done.")
+
+    def process_face_dissim(self):
+        input_dir = self.arguments.input_dir
+        output_dir = self.arguments.output_dir
+
+        print ("Sorting by face dissimilarity...")
+
         img_list = [ [x, face_recognition.face_encodings(cv2.imread(x)) ] for x in tqdm( self.find_images(input_dir), desc="Loading") ]
```
iperov: `img_list = [ [x, face_recognition.face_encodings(cv2.imread(x)), 0 ] for x in tqdm( self.find_images(input_dir), desc="Loading") ]`

iperov:
face-cnn-dissim:

```
    def process_face_cnn_dissim(self):
        from lib import FaceLandmarksExtractor

        input_dir = self.arguments.input_dir
        output_dir = self.arguments.output_dir

        print ("Sorting by face-cnn dissimilarity...")

        img_list = []
        for x in tqdm( self.find_images(input_dir), desc="Loading"):
            d = FaceLandmarksExtractor.extract(cv2.imread(x), 'cnn', True)
            img_list.append( [x, np.array(d[0][1]) if len(d) > 0 else np.zeros ( (68,2) ), 0 ] )

        img_list_len = len(img_list)
        for i in tqdm ( range(0, img_list_len-1), desc="Sorting"):
            score_total = 0
            for j in range(i+1,len(img_list)):
                if i == j:
                    continue
                fl1 = img_list[i][1]
                fl2 = img_list[j][1]
                score_total += np.sum ( np.absolute ( (fl2 - fl1).flatten() ) )

            img_list[i][2] = score_total

        print ("Sorting...")
        img_list = sorted(img_list, key=operator.itemgetter(2), reverse=True)
        self.process_final_rename (output_dir, img_list)

        print ("Done.")
```

AbysmalBiscuit:
@iperov I've added in the fixes/changes, thanks for pointing them out! :)

Clorr:
Let me know when it's ready to merge

AbysmalBiscuit:
@Clorr as far as I see it, it should be ready to merge. But it could be worthwhile to wait until @iperov has a look at it to make sure.

Apollo122:
I would make 'folders" default option of grouping. It's safer than changing the name because of alignment file.

Apollo122:
I'm grouping faces by face-cnn and getting many "No faces were detected" messages. I extracted those faces with cnn.

iperov:
I didnt tested folders, so all fine for me.

> I would make 'folders" default option of grouping. It's safer than changing the name because of alignment file.

is this dangerous ? just restart align. Source files never touched.

AbysmalBiscuit:
@Apollo122 'rename' is the default because @iperov is the author of the original and wrote half the tool, so if he wants it to be that way then I don't really see an issue with that. :)
Also there is the `-k/--keep` option that allows you to keep all the original files and the `-o/--output` option for outputting the sorted/grouped files to a different directory. The two combined mean that you input files are untouched and you get your sorted/grouped files.

Can you please provide the command you used for grouping by face-cnn along with the console output/any traceback messages you received?
When I tested it it worked.

@iperov I think this comes back to the issue that people want to use sort after they have extracted 2000 faces from several minutes of video, or in their folders where they put all the faces that they have extracted, which could also be in the 1000s of images.
However I don't see this as an issue since as I said above, there are options to avoid any risk.

Apollo122:
@AbysmalBiscuit I'm not asking it for myself as I change the defaults before using it. Rename option will be problematic for inexperienced users. He will sort his images and get renamed files and during convert he will get problems. Then we'll see lots of issue posts.

It was a basic command tools.py sort -i whatever -o whatever/sorted -k -by face-cnn (grouping default)
Note that I sorted 16k png's and it missed 80-100 of them. I was weird because I got those images using cnn. Also at the end of the sorting, created folders remained empty for awhile I guess it was still copying the files. After a minute or two I saw the files. Although folder 0 remained empty. A message would be useful for the situations like this, letting the user know he should wait a little after script is done

iperov:
Apollo122 just dont use this tool, if it too dangerous too you.

iperov:
lets merge ?

AbysmalBiscuit:
@iperov I have actually found a bug in the face-cnn method when grouping by folders, but it was a 1 line fix (wrong score calculating method was being called).

I did start to test it extensively and found a problem where the grouping doesn't work very well if the data isn't pre-sorted. I know how to fix this, which should greatly improve how well the program groups into folders, but it will take me a few hours to change it and to test it.
As an aside the fix that I came up with is actually what I wanted to do right from the start with the grouping by folders, but couldn't think of how to implement it back then. :)

So I'd rather wait just a little longer until I sort this out. (Pun not intended)

@Apollo122 I understand the concern for inexperienced/non-programmer users. The best thing I can think of would be to put a disclaimer/warning message when you use `tools.py` or `sort` that tells people to backup their data and/or to test the tool on a small set of data to make sure they understand how it works.

Could you please elaborate on what you mean by having missed 80-100 images; is it that it didn't group them at all, or they were put into individual forlders?

When the program is running it should display the progress of the stage it's on, i.e. Loading, Sorting or Grouping, and last Moving.
To me it sounds like you're on Windows, and explorer has a tendency not to refresh quickly when new files/folders are added.
I mostly use GNU/Linux, so I don't have that problem, in the sense that files appear instantly in the folders after they are moved.

Apollo122:
@AbysmalBiscuit
It couldn't detect faces on those images. It logged "No faces were detected" during sorting.
Yes I'm testing on Windows. No I see the folders but they stayed empty for awhile, this problem occurred because I used keep option. And I had to wait the copy operation but without indicator I assumed it bugged out

iperov:
I can fix "No faces were detected"
but it required to change FaceLandmarksExtractor.py

iperov:
but i will not fix it in near future

AbysmalBiscuit:
I refactored and redesigned the program.
I wrote something like a very basic test file to check that all the option permutations work, and it seems to be ok.

Original functionality is exactly the same as before (to sort by renaming with hist as the sorting method):
`python tools.py sort -i <input_dir>`

To sort by one method and group by another:
`python tools.py sort -i <input_dir> -o <output_dir> -k -s face -g hist`

AbysmalBiscuit:
I amended the previous commit since it didn't get signed for some reason.

iperov:
sry, I will not test your refactorings again.

AbysmalBiscuit:
@Clorr 
I have written a unittest script that tests all the options and possible combinations of options. 
Everything passed with OK, so as far as I see it, it is ready to be merged.

Clorr:
Ok, merged. Thanks for this work ;-)

iperov:
and where is rename functionality ?

iperov:
@AbysmalBiscuit

iperov:
ok all fine for me

iperov:
but with new architecture too hard to integrate something new.
All like welded together.
This is why I hate optimizations.

iperov:
@AbysmalBiscuit how I can add 'regularizer' into your architecture ??

Clorr:
@iperov if you want to add a new file in `tools/` , it is ok. What is 'regularizer' for?

Issue #213, opened 2018-02-20
check for duplicates in extract folder


flipflopbboi:
Hello all,
I have been having trouble with cloud servers shutting down unexpectedly so I edited the original `extract.py` to not overwrite if the image has already been processed in a previous run.

Note that I am currently assuming an `idx` of `0` (i.e. single face was found in photo, usually denoting successful face extraction - all extracted images with nonzero index have been failures from what I ve seen, please enlighten me further!)

This can be handy since somebody may update his image db but should not wait for complete re-extraction!
Note that this is on an earlier version I pulled from this repo so not directly applicable, but I am sure this can be implemented extremely quickly.

Just thought I'd share this idea: you can have a `-no` flag in the extract command to prevent overwriting.
Thoughts? Thanks to all contributors for the good work!

``` python
import os
def process(self):
        extractor_name = "Align" # TODO Pass as argument
        extractor = PluginLoader.get_extractor(extractor_name)()

        try:
            for filename in self.read_directory():
                output_file = self.output_dir / Path(filename).stem
                output_file_to_check = os.path.abspath(str(output_file) +
                                                            '0' +
                                                            Path(filename).suffix)
                if os.path.isfile(output_file_to_check):
                    print('File {} already exists, will not overwrite'.format(output_file_to_check))
                else:
                    image = cv2.imread(filename)
                    for idx, face in self.get_faces(image):
                        resized_image = extractor.extract(image, face, 256)
                        cv2.imwrite(str(output_file) + str(idx) + Path(filename).suffix, resized_image)
                
        except Exception as e:
            print('Failed to extract from image: {}. Reason: {}'.format(filename, e))
```

fetchlister:
Just a heads up, https://cloud.google.com/free/ gives you 300 dollars of free credit. I built a quad core linux system to run facewap on and added a 3584 cuda core Tesla P100. Trained models on it like a charm and only ran out of the free 300 after about a week, plenty of time to train models on a P100. I pay for it now because its honestly a great price for that many Cuda cores and the storage and cpu are basically free - you can get a model down to 0.02 in about 15 hours ona free instance. and it pretty much never goes down. Always been up for me.

flipflopbboi:
Hi, thanks for the suggestion! I am actually running on a google colab notebook which provides free access to a Tesla K80. Problem is you need a good internet connection to stay connected otherwise the notebook stops :/  
I would be very interested in instruction on how to setup a vm on gcp! 
[...]

PR #301, opened 2018-03-18
[WIP] Alignment file editing tools

torzdf:
If, like me, you don't want to keep tonnes of faces on your hard drive after you've generated your alignments file, then this tool may be useful for you.

The feature list is slowly building. At the moment it does:

- remove - Remove faces that have been deleted from your aligned faces directory from the alignments file (Similar to the convert command this does this outside of the process so you do not need to keep the aligned faces on your hard disk after extraction)
- reformat - change the format of your existing alignments file to json, yaml or pickle
- frames - check the alignments file against the source frames to identify frames with multiple/no faces, frames that don't exist in alignments file and frames in alignments file where the source file doesn't exist.

**Useage:**
The basic useage is:
`python tools.py alignments -a <path to alignment file> {remove, reformat, frames} [task options]
`

**remove**
`python tools.py alignments -a <path to alignment file> remove -f <path to folder containing aligned faces>
`
Optionally you can add the -d flag to specify a different destination format. It defaults to output in the same format as the source file.

A basic workflow might be:
- Remove any of your false-positives/unwanted faces from your aligned faces directory
- Run alignment_tools.py, to remove these faces from your alignments file and ensure that they won't be mapped back during the convert process (your original alignments file is backed up)
- Delete your unwanted aligned faces directory

**reformat**
`python tools.py alignments -a <path to alignment file> reformat -d <json, yaml or pickle>
`
NB: PyYAML must be installed to use the YAML encoder/decoder. 

This will change the format of your alignments file into pickle, json or yaml. The original alignments file will be backed up.

**frames**
`python tools.py alignments -a <path to alignment file> frames -f <path to folder containing source frames> -t <missing-alignments, missing-frames, no-faces, multi-faces> -o <console, file, move>
`

This performs various reports/actions on your source frames by comparing with the alignments file. Specifically:
- missing-alignments: Source frames that exist that don't have a corresponding match in the alignments file
- missing-frames: Entries in the alignments file that don't have a corresponding source frame
- no-faces: Source frames that have no faces identified within the alignments file
- multi-faces: Source frames that have multiple faces identified within the alignments file.

The output can be:
- console: Prints the results to the console window
- file: Saves the results to a text file within your source frame folder
- move: Moves the files that are found into a new sub-folder within your source frame folder. NB: selecting this for missing-frames will default to console output, as there would not be any frames to move.

_______________________________
If anyone else has any suggestions/feedback, please let me know and I will look to implement/abandon as necessary.

Expect bugs, this is an early upload.

To do:
- Merge multiple alignment files into one.
- Remove whole frames from an alignment file
- Identify potential false positives

### DIFF
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -178,7 +178,7 @@ class TrainingProcessor(object):
                 print('Saving model weights has been cancelled!')
             exit(0)
         except Exception as e:
-            print(e)
+            raise e
             exit(1)
 
     def set_tf_allow_growth(self):
@@ -200,4 +200,4 @@ class TrainingProcessor(object):
                 cv2.imwrite('_sample_{}.jpg'.format(name), image)
         except Exception as e:
             print("could not preview sample")
-            print(e)
+            raise e


ruah1984:
https://github.com/deepfakes/faceswap/pull/298

AbysmalBiscuit:
As @ruah1984 points out this will not be at all compatible with `tools.py` as the main tools calling script.

I have had a look at this PR and I like what you're trying to do, however I tried running the program and it doesn't work. There are a lot of problems and mistakes all throughout the program, which makes sense since it's an early upload.

However it would take more time to explain everything and provide suggestions via code review than to just fix them myself. 
@torzdf if you would like after I'm done, I could submit a PR to your PR with my changes/fixes, or I could open a new PR. :)

torzdf:
Yeah, this is an early draft, thrown together in about half an hour. There are issues, and I am modifying the code at the moment to iron out a few bugs, and refactor a fair bit.

I only work with committed code, so if/when that PR gets committed, I'll make the necessary changes. to comply. 

I wouldn't edit this yet as I'm still actively working on this (today at least), so will publish changes intermittently.

A couple of things I'm currently fixing:

- Traverse alignments in reverse order to mitigate changing index numbers
- Rename aligned faces to match their new index numbers (i.e if face xxxx_0.png gets deleted, after running this process any files named xxxx_1.png etc will no longer match with the alignments file, so they need to be renamed)

Please let me know of anything obvious which stands out to you, as I can go in and fix.

AbysmalBiscuit:
I'll do a quick review of your code then.

tools/alignment_tools.py
```diff
@@ -0,0 +1,186 @@
...
+
+    def __init__(self, alignments_dir):
+        self.extensions = ['json', 'p', 'yaml', 'yml']
+        self.alignments_file = None
+        self.alignments_format = None
+        self.alignments = None
+
+        self.load_alignments(alignments_dir)
+
+    def alignments_file_exists(self, alignments_base, extension):
```
AbysmalBiscuit: I think it would be better to get the user to specify a path to a target alignments file as an input.
This would allow you to avoid all the checks for there being a single file.

tools/alignment_tools.py
```diff
@@ -0,0 +1,186 @@
...
+            return json.loads(alignments.read())
+        elif self.alignments_format == 'p':
+            return pickle.loads(alignments.read())
+        elif self.alignments_format in ('yaml', 'yml'):
+            try:
+                import yaml
+            except ImportError:
+                print("You must have PyYAML installed to use YAMLSerializer")
+            return yaml.load(alignments.read())
+        alignments.close()
```
AbysmalBiscuit: The file will never be closed if the method returns somewhere within the if-elif-elif loop.

You need to use a `with open() as f:` block instead of the way you're doind it now.

tools/alignment_tools.py
```diff
@@ -0,0 +1,186 @@
...
+        if self.alignments_format == 'json':
+            return json.loads(alignments.read())
+        elif self.alignments_format == 'p':
+            return pickle.loads(alignments.read())
+        elif self.alignments_format in ('yaml', 'yml'):
+            try:
+                import yaml
+            except ImportError:
+                print("You must have PyYAML installed to use YAMLSerializer")
+            return yaml.load(alignments.read())
```
AbysmalBiscuit: This is an error that is also present in `lib/Serializer.py`, you will get an error thrown on attempting to return if the yaml serializer could not be imported, which defeats the purpose of the try-except block.

It's better to do it like so:
```python
try:
    global yaml
    import yaml
    return yaml.load(alignments.read())
except ImportError:
    print("You must have PyYAML installed to use YAMLSerializer")
    exit(1)
```

tools/alignment_tools.py
```diff
@@ -0,0 +1,186 @@
...
+
+    def check_alignments_file(self, alignments_base):
+        ''' Checks that there is just one alignment file '''
+        file_count = 0
+        for extension in self.extensions:
+            file_count += self.alignments_file_exists(alignments_base, extension)
+        if file_count != 1:
+            raise ValueError ('1 alignments file should exist, but {} were found'.format(file_count))
+
+    def read_alignments(self):
```
AbysmalBiscuit: Both for reading and writing the files it would be simpler in terms of logic to just assign a serializer method to an instance variable during the initial arguments processing, and then to call the serializer that way.

tools/alignment_tools.py
```diff
@@ -0,0 +1,186 @@
...
+    def load_alignments(self, alignments_dir):
+        ''' load the data from the alignments file '''
+        alignments_base = os.path.join(alignments_dir, 'alignments')
+        try:
+            self.check_alignments_file(alignments_base)
+            self.alignments = self.read_alignments()
+        except ValueError as v:
+            print(v)
+            sys.exit()
+        except:
```
AbysmalBiscuit: Redundant except, if any other errors occur they will be raised automatically. If this may suppress the error output via the empty `raise`.

tools/alignment_tools.py
```diff
@@ -0,0 +1,186 @@
...
+            raise
+
+    def backup_alignments(self):
+        ''' Backup copy of old alignments '''
+        now = datetime.datetime.now().strftime("%Y-%m-%d--%H_%M_%S")
+        backupfile = '{}_{}{}'.format(os.path.splitext(self.alignments_file)[0], now, os.path.splitext(self.alignments_file)[1])
+        print('Backing up old {} to {}'.format(self.alignments_file, backupfile))
+        os.rename(self.alignments_file, backupfile)
+
+    def write_alignments(self):
```
AbysmalBiscuit: See comments about the `read_alignments()` method.

tools/alignment_tools.py
```diff
@@ -0,0 +1,186 @@
...
+                        required=True)
+    parser.add_argument('-f', '--faces_folder',
+                        type=str,
+                        dest='faces_dir',
+                        help='Input directory of source A extracted faces.',
+                        required=True)
+
+def remove_alignments(arguments):
+    ''' Remove items from alignments file '''
+    faces = Faces(arguments.faces_dir)
```
AbysmalBiscuit: This is minor, but to avoid having to do `faces.faces` later on, you can just do:
```python
faces = Faces(arguments.faces_dir).faces
```

tools/alignment_tools.py
```diff
@@ -0,0 +1,186 @@
...
+                index = int(face_filename[face_filename.rindex('_')+1:])
+                original_file = '{}{}'.format(face_filename[:face_filename.rindex('_')], os.path.splitext(face)[1])
+                if self.faces.get(original_file, '') == '':
+                    self.faces[original_file] = [index]
+                else:
+                    self.faces[original_file].append(index)
+        except ValueError as v:
+            print(v)
+            sys.exit()
+        except:
```
AbysmalBiscuit: Redundant except with a raise that may mask the error output/traceback.

tools/alignment_tools.py
```diff
@@ -0,0 +1,186 @@
...
+            file_count += self.alignments_file_exists(alignments_base, extension)
+        if file_count != 1:
+            raise ValueError ('1 alignments file should exist, but {} were found'.format(file_count))
+
+    def read_alignments(self):
+        ''' Read the alignments data from the correct format '''
+        print('Loading {} alignments from {}'.format(self.alignments_format, self.alignments_file))
+        alignments = open(self.alignments_file, 'r')
+        if self.alignments_format == 'json':
+            return json.loads(alignments.read())
```
AbysmalBiscuit: When used in a `with open() as f:` block you can just use the `load` method from all these libraries. I'm not sure, but maybe you don't even have to do `alignments.read()` with the way you're doind it now.

Samle if you would assign an instance serializer method:
```python
with open(self.alignments_file, 'r') as alignments:
    self.alignments = self.serializer.load(alignments)
```

tools/alignment_tools.py
```diff
@@ -0,0 +1,186 @@
...
+    ''' Remove items from alignments file '''
+    faces = Faces(arguments.faces_dir)
+    alignment_data = AlignmentData(arguments.alignments_dir)
+    del_count = 0
+    print('\n-------------------------------------\n Processing compare...')
+    for key, value in alignment_data.alignments.items():
+        if len(value) == 0 or len(value) == len(faces.faces.get(key,'')):
+            continue
+        for idx, val in enumerate(value):
+            if idx not in faces.faces.get(key,[-1]):
```
AbysmalBiscuit: You are trying to remove the faces, hence it should be:
```python
if idx in faces.faces.get(key, [-1]):
```
Or with a more Pythonic syntax:
```python
if idx in faces.faces[key][-1]:
```

tools/alignment_tools.py
```diff
@@ -0,0 +1,186 @@
...
+                        required=True)
+
+def remove_alignments(arguments):
+    ''' Remove items from alignments file '''
+    faces = Faces(arguments.faces_dir)
+    alignment_data = AlignmentData(arguments.alignments_dir)
+    del_count = 0
+    print('\n-------------------------------------\n Processing compare...')
+    for key, value in alignment_data.alignments.items():
+        if len(value) == 0 or len(value) == len(faces.faces.get(key,'')):
```
AbysmalBiscuit: The check for 
```python
len(value) == len(faces.faces.get(key,'')
```
makes it so that no images will be removed, at least when I tested the method where there was only 1 face detected.

Hence I think this would work better:
```python
if len(value) == 0:
```

AbysmalBiscuit:
I've made a gist with all the changes I suggested:
https://gist.github.com/AbysmalBiscuit/81f81252fcf2c769b70a15a0b9f22726

You can run it directly by:
`python alignments.py alignments -j remove -a <path to alignments file> -f <path to aligned dir>`

I did port it to work with `tools.py` though, hence why it's somewhat different in terms of layout.
It adds the option to change the serializer.

tools/alignment_tools.py
```diff
@@ -0,0 +1,235 @@
...
+            raise
+
+    def backup_alignments(self):
+        ''' Backup copy of old alignments '''
+        now = datetime.datetime.now().strftime("%Y%m%d_%H%M%S")
+        backupfile = '{}_{}{}'.format(os.path.splitext(self.alignments_file)[0], now, os.path.splitext(self.alignments_file)[1])
+        print('Backing up old {} to {}'.format(self.alignments_file, backupfile))
+        os.rename(self.alignments_file, backupfile)
+
+    def write_alignments(self):
```
torzdf: This is still very much WIP, but was looking to fix some of the potential bugs I mentioned earlier

torzdf:
Thanks for all of the feedback.

I've just pushed a commit to fix these bugs:

-     Traverse alignments in reverse order to mitigate changing index numbers following a delete
-     Rename aligned faces to match their new index numbers (i.e if face xxxx_0.png gets deleted, after running this process any files named xxxx_1.png etc will no longer match with the alignments file, so they need to be renamed)
-     Assumption that the original file will have the same extension as the aligned face is incorrect.

I'm going to look at your code now and will look to implement where I can.

Cheers

tools/alignment_tools.py
torzdf: Yeah, that's probably correct. Amending

tools/alignment_tools.py
torzdf: amending

tools/alignment_tools.py
torzdf: Yeah, that was a copy/paste and I noticed it. It was actually going to raise a value error into a parent try except block, but that seems overkill so I'm doing this instead.

tools/alignment_tools.py
torzdf: As above, I've ditched altogether and am just handling the import error separately.

tools/alignment_tools.py
torzdf: Yeah, it's not perfect, but I'm going to leave as is, as I plan to (and have) added more variables.

tools/alignment_tools.py
torzdf: As above. Was temporary, now gone

tools/alignment_tools.py
torzdf: No, I think I'm correct here... 

Although my code has changed a bit now, I'll reference the original code:

` for idx, val in enumerate(value):`

This is iterating through all the stored alignments for a given image in the alignments file.

We then check this against the output file list. If the file *doesn't* exist then we want to delete it from the alignments file, hence 

` if idx not in faces.faces.get(key,[-1])` 

If we performed the delete when it *was* in the file list, we would delete all the alignments where we did have faces and leave all those where we didn't.

tools/alignment_tools.py
torzdf: Ok, this is strange. It is simply there to speed up the process. Basically if there are the same number of face files as there are alignments, skip this because there is nothing to do. It works for me, so I'm a bit stumped. Ultimately if there is one image then alignment would be:

 `[
    {
      "h": 213,
      "y": 103,
      "landmarksXY": [
      ...
      ]
    }]`

and faces index list would be:
[0]

The length of both is 1, so it is skipped.

If you have deleted your image, then the default value will kick in (''). The length of this is zero, so the code should kick in.

AbysmalBiscuit:
Hey, my PR was merged so you'll have to update your branch to work with tools.py (my gist, or sort.py should work as decent templates). :)

I haven't had an extensive look at your commits, but they seem to be good.

torzdf:
Thanks. Been a bit snowed at work, but will hopefully update in next couple of days

torzdf:
ref: #302 #208 #213

PR #309, opened 2018-03-24
Adds support for arbitrary image rotations

coldstacks:
Previously only 90, 180, and 270 degree rotations were supported.  This PR adds support for arbitrary-angle image rotations during extraction.

This PR adds two new command-line options to `extract`, both of which require `--rotate-images on` in order to be used:
- `--rotation-angle <step_size>` lets you adjust the step size (default is `90`, as before)
- `--rotation-angle-list <angle1> <angle2> <angle3> ...` lets you enumerate a list of angles to use

I think `--rotation-angle-list` can be especially useful because the user will often know what angles are relevant for their inputs.  For example, it is common for a person's head to tilt a bit to the left sometimes and a bit to the right sometimes, but never go completely sideways or upside-down.  In this case, specifying just a couple of angles such as `--rotation-angle-list 30 330` can expand the extractor's effectiveness considerably.

### DIFF
--- a/lib/cli.py
+++ b/lib/cli.py
@@ -41,6 +41,7 @@ class DirectoryProcessor(object):
     num_faces_detected = 0
     faces_detected = dict()
     verify_output = False
+    rotation_angles = None
 
     def __init__(self, subparser, command, description='default'):
         self.create_parser(subparser, command, description)
@@ -60,8 +61,19 @@ class DirectoryProcessor(object):
             self.serializer = Serializer.get_serializer(self.arguments.serializer or "json")
         print("Using {} serializer".format(self.serializer.ext))
 
+        try:
+            if self.arguments.rotate_images == "on":
+                print("Rotation angle list")
+                print(self.arguments.rotation_angle_list)
+                if self.arguments.rotation_angle_list is not None:
+                    self.rotation_angles = self.arguments.rotation_angle_list
+                else:
+                    self.rotation_angles = range(self.arguments.rotation_angle, 360, self.arguments.rotation_angle)
+        except AttributeError:
+            pass
+
         print('Starting, this may take a while...')
-        
+
         try:
             if self.arguments.skip_existing:
                 self.already_processed = get_image_paths(self.arguments.output_dir)
--- a/lib/cli.py
+++ b/lib/cli.py
@@ -62,13 +62,16 @@ class DirectoryProcessor(object):
         print("Using {} serializer".format(self.serializer.ext))
 
         try:
-            if self.arguments.rotate_images is not None:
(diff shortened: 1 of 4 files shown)


bryanlyon:
I like the idea behind this PR, it's where I wanted the rotation feature to go in the first place, but we made sure that the -r command should be able to handle all the features you have here.  Just add your features to the existing -r command.

In other words, please modify your PR to use - r 90 for specific rotations and some other notation with -r (maybe comma separated) for a list of rotations.

I really appreciate the work as this was on my list to work in very soon.

coldstacks:
Sure, I can modify -r instead.  Do you want to keep the "rotate in increments of size <arg> up to 360 degrees" behavior?  If so, we'll need to distinguish between -r 90 meaning "90, 180, 270" and -r 90 meaning "90 only" -- not hard to do, but want to double-check that is the intention.  (I suspect it will generally be more useful for the user to enumerate the angles they want to check, but I can see an argument for the other way too.)

bryanlyon:
Currently -r on does that.  Keep that the same and add yours as the numbers.

coldstacks:
Ok, updated it so there's just a single -r option.

bryanlyon:
I looked at thus quickly, mostly there but please make "-r on" default to 90 degrees rotations around the circle so that current operation is unchanged?

torzdf:
It's good to see functionality extended.

Quick question. I chose to use flip/transpose on 90 degree rotations because it performs faster than getRotationMatrix. I believe it's marginal, but when you apply over thousands of images it will stack.

Does moving everything to getRotationMatrix cause significant slow down (unfortunately I can't test at the moment)? It may not, but if it does, it may be worth keeping the original method for 90 degree rotations.

coldstacks:
I haven't benchmarked either, but it seems like a safe assumption that the old method would be faster.  Also just noticed that the new `rotate_image` uses positive for counter-clockwise (same as getRotationMatrix), but the old `rotate_image` used positive for clockwise.  I'll do a speed check and update the new function so it is consistent with the old one tonight.

coldstacks:
Tested this just now, and it turns out that the new `rotate_image` using `getRotationMatrix` is (surprisingly) 30-40% faster than the old flip/transpose method.  This is a nice result, because it means there is no need to special-case 90 degree rotations!

I also switched the convention to use positive value = clockwise rotation, for backwards compatibility.  So, existing alignments.json files should work correctly with the new method.

Clorr:
Is that ready to merge?

coldstacks:
@Clorr if you're asking me, yes, as far as I know it's ready to merge.

Clorr:
Let's go

PR #312, opened 2018-03-24
sorttool fix, and added --sort-by face-yaw

iperov:
fixed 'No faces were detected'

speed of face sorting much increased, because 256x256 face already "detected", so no need to detect it via dlib again

added --sort-by face-yaw
result:

37765244-23e3f66c-2ddd-11e8-85da-26e14e1c021c 1


iperov:
@Clorr merge it plz

PR #313, opened 2018-03-24
[WIP] Alignment file editing tools

torzdf:
If, like me, you don't want to keep tonnes of faces on your hard drive after you've generated your alignments file, then this tool may be useful for you.

The feature list is slowly building. At the moment it does:

- remove - Remove faces that have been deleted from your aligned faces directory from the alignments file (Similar to the convert command this does this outside of the process so you do not need to keep the aligned faces on your hard disk after extraction)
- reformat - change the format of your existing alignments file to json, yaml or pickle
- frames - check the alignments file against the source frames to identify frames with multiple/no faces, frames that don't exist in alignments file and frames in alignments file where the source file doesn't exist.

**Useage:**
The basic useage is:
`python tools.py alignments -a <path to alignment file> {remove, reformat, frames} [task options]
`

**remove**
`python tools.py alignments -a <path to alignment file> remove -f <path to folder containing aligned faces>`
Optionally you can add the -d flag to specify a different destination format. It defaults to output in the same format as the source file.

A basic workflow might be:
- Remove any of your false-positives/unwanted faces from your aligned faces directory
- Run alignment_tools.py, to remove these faces from your alignments file and ensure that they won't be mapped back during the convert process (your original alignments file is backed up)
- Delete your unwanted aligned faces directory

**reformat**
`python tools.py alignments -a <path to alignment file> reformat -d <json, yaml or pickle>`
NB: PyYAML must be installed to use the YAML encoder/decoder.

This will change the format of your alignments file into pickle, json or yaml. The original alignments file will be backed up.

**frames**
`python tools.py alignments -a <path to alignment file> frames -f <path to folder containing source frames> -t <missing-alignments, missing-frames, no-faces, multi-faces> -o <console, file, move>
`

This performs various reports/actions on your source frames by comparing with the alignments file. Specifically:
- missing-alignments: Source frames that exist that don't have a corresponding match in the alignments file
- missing-frames: Entries in the alignments file that don't have a corresponding source frame
- no-faces: Source frames that have no faces identified within the alignments file
- multi-faces: Source frames that have multiple faces identified within the alignments file.

The output can be:
- console: Prints the results to the console window
- file: Saves the results to a text file within your source frame folder
- move: Moves the files that are found into a new sub-folder within your source frame folder. NB: selecting this for missing-frames will default to console output, as there would not be any frames to move.
___________________________________________________________
If anyone else has any suggestions/feedback, please let me know and I will look to implement/abandon as necessary.

Expect bugs, this is an early upload.

To do:
- Merge multiple alignment files into one.
- Remove whole frames from an alignment file
- Identify potential false positives

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@
 !lib
 !scripts
 !plugins
+!tools
--- a/tools.py
+++ b/tools.py
@@ -3,6 +3,7 @@ import sys
 from lib.cli import FullHelpArgumentParser
 # Importing the various tools
 from tools.sort import SortProcessor
+from tools.alignment_tools import AlignmentTool
 
 # Python version check
 if sys.version_info[0] < 3:
@@ -27,6 +28,8 @@ if __name__ == "__main__":
     sort = SortProcessor(
         subparser, "sort", "This command lets you sort images using various "
                            "methods.")
+    alignments = AlignmentTool(subparser,
+                         'alignments', 'Perform various edits to an alignments file.')
     parser.set_defaults(func=bad_args)
     arguments = parser.parse_args()
     arguments.func(arguments)
--- a/tools/alignment_tools.py
+++ b/tools/alignment_tools.py
@@ -327,20 +327,8 @@ if __name__ == '__main__':
    
     parser = argparse.ArgumentParser()
     subparser = parser.add_subparsers()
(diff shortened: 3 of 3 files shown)


torzdf:
ref: #301

torzdf:
Closing due to lack of interest/until I have more time to expand functionality

PR #318, opened 2018-03-26
 Fix to an UnboundLocalError due to rename.

JayantPythonLover:
Just a minor fix that arose due to the recent renaming.

### DIFF
--- a/lib/FaceLandmarksExtractor/FaceLandmarksExtractor.py
+++ b/lib/FaceLandmarksExtractor/FaceLandmarksExtractor.py
@@ -173,7 +173,7 @@ def extract(input_image_bgr, detector, verbose, all_faces=True, input_is_predete
             center[1] -= (bottom - top) * 0.12
             scale = (right - left + bottom - top) / 195.0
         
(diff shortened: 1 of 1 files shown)


lib/FaceLandmarksExtractor/FaceLandmarksExtractor.py
```diff
@@ -173,7 +173,7 @@ def extract(input_image_bgr, detector, verbose, all_faces=True, input_is_predete
             center[1] -= (bottom - top) * 0.12
             scale = (right - left + bottom - top) / 195.0
         
-            image = crop(input_image, center, scale).transpose ( (2,0,1) ).astype(np.float32) / 255.0
```
kvrooman: add the following after line 149 instead
`input_image = input_image_bgr[:,:,::-1].copy()
`

#312 added an if statement which could create situations where input_image is not bound. Merge #318  reverts #281 by switching the image sent to the aligner back to the previous logic.  The Unbound error can simply be fixed without changing code design by adding the above line which binds input_image inside the if statement which #312 added

PR #319, opened 2018-03-27
removing aligments.json

iperov:
**removed alignments.json**

You must restart extracts for your models.
All necessary data stored inside png in aligned\ folder.
These pngs can be renamed.
So you can super quickly filter target face in both aligned\ folders by using sort tool (--sort-by hist), and it will works fine in convert !

**sort tool**

`--sort-by` `face face-dissim face-yaw` now uses new pngs with faceswap data inside.
old `face face-cnn` removed, now sort tool no more uses neither dlib nor keras.

### DIFF
--- a/lib/AlignedPNG.py
+++ b/lib/AlignedPNG.py
@@ -1,0 +1,219 @@
+PNG_HEADER = b"\x89PNG\r\n\x1a\n"
+
+import string
+import struct
+import zlib
+import pickle
+
+class Chunk(object):
+    def __init__(self, name=None, data=None):
+        self.length = 0
+        self.crc = 0
+        self.name = name if name else "noNe"
+        self.data = data if data else b""
+
+    @classmethod
+    def load(cls, data):
+        """Load a chunk including header and footer"""
+        inst = cls()
+        if len(data) < 12:
+            msg = "Chunk-data too small"
+            raise ValueError(msg)
+
+        # chunk header & data
+        (inst.length, raw_name) = struct.unpack("!I4s", data[0:8])
+        inst.data = data[8:-4]
+        inst.verify_length()
+        inst.name = raw_name.decode("ascii")
+        inst.verify_name()
+
+        # chunk crc
+        inst.crc = struct.unpack("!I", data[8+inst.length:8+inst.length+4])[0]
+        inst.verify_crc()
+
+        return inst
+
+    def dump(self, auto_crc=True, auto_length=True):
+        """Return the chunk including header and footer"""
+        if auto_length: self.update_length()
+        if auto_crc: self.update_crc()
+        self.verify_name()
+        return struct.pack("!I", self.length) + self.get_raw_name() + self.data + struct.pack("!I", self.crc)
+
+    def verify_length(self):
+        if len(self.data) != self.length:
+            msg = "Data length ({}) does not match length in chunk header ({})".format(len(self.data), self.length)
+            raise ValueError(msg)
+        return True
+
+    def verify_name(self):
+        for c in self.name:
+            if c not in string.ascii_letters:
+                msg = "Invalid character in chunk name: {}".format(repr(self.name))
+                raise ValueError(msg)
+            return True
+
+    def verify_crc(self):
+        calculated_crc = self.get_crc()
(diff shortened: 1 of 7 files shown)


bryanlyon:
While I like the PNG storage feature, I do not like removing alignments.json to get it.  I see no reason that we have to get rid of the alignment.json to add this feature.

iperov:
@Clorr will you adapt it with dfaker ?

iperov:
@bryanlyon 
i see no reason to use alignments.json. It is like ballast. 
Backward compatibility? Are we making OS here? Why collect and support old non-effective solutions?

iperov:
with alignments.json I have to manually delete excess faces in data_B\aligned folder, because I cannot rename them by sort tool. This is especially big problem with videos with large amount of faces.

bryanlyon:
It's not just a matter of backwards compatibility, it's also about jpg support and keeping things simple, editable, and concise.  I agree with the png embedded data, but it should not come at the experience of portability and workflow.

There are edits that I have done that are only possible with a single alignment file.  There are methods and techniques that I see which are only possible with an alignments file.  There are hundreds of datasets with an alignments file.

I'm not saying we can't use embedded png data, just that we shouldn't be abandoning the current system, especially not without a migration plan.

iperov:
what you want to edit in aligned faces in jpg?

iperov:
`There are hundreds of datasets with an alignments file.`
what ?? what you mean ? Example of these datasets please.

iperov:
Datasets which can be downloaded from reddit, such as Cageset and other - contains NO alignments.json file.

Also these aligned 256x256 photos are useless, because contains no aligned landmarks inside. 
They are like obsolete "VHS" cassette, while we are walking to DVD future.

bryanlyon:
I don't mean downloaded datasets, I mean datasets people are using right now, removing support for those datasets is not an acceptable option.

This PR will force everyone to run extraction in all their source data again.  It will force all users to keep large pngs, even when they don't want the aligned faces at all.  It will reduce tool choice and increase complexity of any tools that want to edit the alignments.  It's also very poorly tested and not ready to implement.

I like the png embedded option, but it's not vastly superior and certainly not ready to replace the alignment files that exist now.  I support it as an option, not as a replacement.

Send a PR with it as an option or saving/loading both at once and I'll support it, until then, I think this PR is premature and more of a negative than a positive.

AbysmalBiscuit:
@iperov the VHS cassette reference made me feel pretty old. :p

I think that they may be referring to people's private datasets.
A compromise solution that I can see would be to add an option/functionality that could read an `alignments.json` file and embed the data inside it into the images that the data should be embedded in.  It could also have the functionality of reading a bunch of images and extracting the embedded data into an `alignments.json` file.

Also to maybe allow the choice of how the data should be saved, alignments file or embedding (or both?).

It seems @bryanlyon beat me to explaining this. :)

Kirin-kun:
You could change method with format. Store in tExt for PNG and in EXIF for JPGs. I don't think it would be that expensive (but it's up to you, or to someone else who wants to implement it anyway). The piexif lib seem to be able to do that.

But I'm cool with removing the need of a serialization file. Faceswap is a work in progress anyway.

iperov:
please never use my code from this pr

Apollo122:
I agree with @bryanlyon. Embedding data in png's is useful but removing alignment.json all together is a little bit extreme. It would be better if it's optional.

iperov:
Run extract again for existing trainset is absolutely no problem, especially if you have sorttool to sort target quickly.

99% users dont collecting trainsets, they just trying to swap face in some videos with not good result, and they leaves this idea.
I collecting GBs of trainsets, and I have no problem to extract again, if I will get better results in future, because I have all source videos. 
But you guys cant understand this. You are thinking too narrowly. 

I cant fight for every pr, so I decided to leave you. I already regretted that ported face_alignment to keras, you are not worthy of it.  I going to fully refactor crap faceswap architecture from scratch, and remove most of suxx unusable things from it. Users of my windows release will extract faces again with no problem.

@Clorr please dont use code from this pr for your dfaker impromevents.

Bye all, and P.S. sorry bad english.

PR #320, opened 2018-03-28
Revert " Fix to an UnboundLocalError due to rename." -- Bind the variable in question, rather than replace it with another parameter

kvrooman:
Reverts deepfakes/faceswap#318 and fixes Unbound Local Error

#318 found an Unbound Local Error. Proposed new solution binds the offending variable in the case where it isn't bound. #318 instead modified the offending variable by renaming it later in the program to a different variable with a similar sounding name?


JayantPythonLover:
Seems good.

PR #315, opened 2018-03-25
Only import DLIB related modules when required and minor import relocations

torzdf:
DLIB is a memory hog. Firing it up every time faceswap is run can lead to memory issues. This PR changes the imports so that any DLIB related modules are only loaded when required.

I have also moved any imports within the code that I have amended (e.g. tensorflow for allow_growth) to the top, so imports are all together. 

Threading was imported twice for train.py, so this has been fixed.

I think I got them all, but let me know if not. I tried to keep this at the highest level possible.

Tested working with extract, sort, train and convert.

PR #325, opened 2018-04-03
Built in GUI [ALPHA]

torzdf:
Disclaimer: Still under development. There will be bugs, code needs to be cleaned up and is mainly untested (works on Extract for sure). Use at your own risk.

So with that said. I have started to build a GUI. Some people don't like a command line, so this is to cater for them.

I have put up a PR early as I would like some feedback from other platforms whilst I continue to iron out the kinks.

image

Some design choices:

- tkinter is used as it comes with Python and should be cross platform (linux users may need to install the python3-tk package for their distro). Please let me know what commands you need to issue, if any, to run the GUI on your platform, as I will look to run a check and generate a message if not installed, with some useful pointers. No additional modules need to be added to requirements.txt.
- This is not a pure, packaged GUI. It stills needs to be launched from the commandline, with all the previous pre-requisites:
`python faceswap.py gui`
No further options are required nor supported. I have no interest in keeping up and supporting compiled versions of the code. This also ensures that the gui comes as default and is just an option. You can use it, or you can use the commandline. Choice is yours. This is also why I have added it as an additional script. To keep choice, and not forcing the user one way or the other.
- Yet again, I have tried to leave the base of the code well alone. HOWEVER, I have had to significantly alter how argparse arguments are now loaded in (adding them to a list of dictionaries). I have kept the format as close to the original as possible, and I'm hopeful that this won't be too much of a roadblock. This is so I can read them back out into the GUI. If anyone knows a way that I can read the argparse options out without having to make this change then please let me know and I will amend.
- Related to above, the main reason I have made the change to the argparse options is to keep the GUI self-supporting. That is, if someone adds a new argument, then that argument will automatically get added to the relevant page of the GUI. No further development on the GUI will be required to support it. I am well aware that the last thing people probably want to do is support a GUI, so I have attempted to mitigate that in it's entirety. I have also taken on board how quickly other GUIs have gone out of date, hence adding it as an option to the base code.
- The option to save and load configs has been added. This gives a convenient way to be able to load frequently used configs, or save different configs for different purposes. This has been done for the global level (extract, train and convert) and for individual commands. If you have a global config saved, and you just load the config for the extract page (for example), then only the extract section of that config will be loaded.
- Tasks (extract, train, convert) are launched in a separate thread. This ensures that the GUI remains responsive and is to support future development.
- It is definitely function over form. I can't possibly test over multiple platforms, so it is deliberately kept quite basic, using defaults where possible, 

Issues:
- I'm fairly certain that passing the main parser through to the GUI is not the way to go. If anyone has a better way to do this (likely), please let me know
- I'm having issues getting the right hand pane to scale along the x-axis when expanding the window out, so I have had to fix the width for now. This is because you can't put scrollbars onto frames with tkinter, so I have had to use a canvas, which then causes autofilling issues.
- Implementing PR #315 for this would be useful, as I suspect I am hammering Dlib even more now.
- This is the first time I have developed a GUI in python, so go easy on me. Any feedback, if given in a constructive manner is hugely appreciated.
- Error handling and exiting the thread gracefully is non-existent, so underlying script can can lock upon errors.

Future Development:
- If there is interest in this part of the project, then I plan to add in the preview window, loss graph and console out to the GUI, so that the user can stay in one place. I want to get basic functionality bug free first though, and make sure there is demand.
- Similarly, if there isn't interest for this, that's all good too. Happy to keep this in my own fork.
- Extract frames from video, merge frames back to video.


torzdf:
Idiot me removed the actual script execution in final testing before pushing. This has been re-inserted.

GUI is currently successfully running a convert.

torzdf:
I do plan to add that. 

I just want to get basic functionality tested and merged first, then I will be looking to add on a few bells and whistles. Graphing loss is high on the list.

bryanlyon:
This was done exactly right and after testing it several times (including faking up some new arguments) I feel that this is good to add.  Thanks for doing it with so few modifications, since that makes me confident enough to commit it right away.

torzdf:
Woah, really. Lol, I've done quite a significant code rework since this push. 

No biggy, I'll roll back to this version and continue development from there.

Thanks!

ruah1984:
how to run this GUI??

fat-tire:
Also might not want to use the exact phrase "running train" as that's slang, for [well, er... ummm](https://www.urbandictionary.com/define.php?term=Running%20train)  <cough> maybe just "_training..._" works?

(See #334 for more.  Must say this is a really nice addition)

torzdf:
lol.

Well, it's just because I pull the command name in and use that. Maybe I'll change it to Executing:

PR #331, opened 2018-04-04
Enable filename migration for lowmem

bryanlyon:
Right now lowmem and Original use the same filenames.  Since their models are incompatible, this could lead to failure by attempting to  load lowmem weights in original or vice versa.  This name change migration will rename the files existing lowmem models in order to prevent problems and make things more clear for the future.

We should keep this code in for a reasonable time, then remove the filename migration logic once everyone has migrated to the new filenames.  We may want to change original's filenames as well to match the new filename format but I have not done this for consistency reasons as it's less likely to get all users of original model to migrate to the new filenames than the lowmem model.

Recommend that all users either manually rename the lowmem files or run this commit against all their lowmem models to migrate to the new filenames.

### DIFF
--- a/plugins/Model_LowMem/AutoEncoder.py
+++ b/plugins/Model_LowMem/AutoEncoder.py
@@ -1,8 +1,15 @@
 # AutoEncoder base classes
 
-encoderH5 = 'encoder.h5'
-decoder_AH5 = 'decoder_A.h5'
-decoder_BH5 = 'decoder_B.h5'
+encoderH5 = 'lowmem_encoder.h5'
+decoder_AH5 = 'lowmem_decoder_A.h5'
+decoder_BH5 = 'lowmem_decoder_B.h5'
+
+#Part of Filename migration, should be remopved some reasonable time after first added
+import os.path
+old_encoderH5 = 'encoder.h5'
+old_decoder_AH5 = 'decoder_A.h5'
+old_decoder_BH5 = 'decoder_B.h5'
+#End filename migration
 
 class AutoEncoder:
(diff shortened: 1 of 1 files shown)


bryanlyon:
I'd be fine with that myself, but I've found very few people read here, and we don't have good documentation, so renaming by default seemed like the cleanest option.

torzdf:
Of course, it's also impossible to mitigate against every occurrence. For example if someone selected the lowmem option and accidentally selected their original model folder, it would rename and overwrite their original model.

My personal opinion (and it is my opinion, it's probably not correct!), is that I don't like implementing code that you know needs to be removed down the line, because it invariably gets forgotten.

bryanlyon:
I understand, and I wrote this code for that eventuality.  It can be left in indefinitely with no ill effects even if it is forgotten to be removed.  Honestly, I would prefer to completely refactor the whole model system with the model architecture saved along with the weights.  This is something that is much bigger than would be feasible in a single PR.  This change was just to prevent the few errors that come from misusing the weights from a wrong model.

torzdf:
Nah, it's all good. Just playing devil's advocate to an extent. :) I reckon you should go ahead and chuck this in.

I like the idea of saving the model architecture, to be honest, the Machine Learning part of the code is still just dark arts to me at the moment ;)

I'm currently in the process of rewriting the scripts/cli section as we have some mammouth classes and functions that are becoming fairly unmanageable, mainly along the lines of separating out the arguments from the actual functions and conforming more towards the PEP 8 standard. Hopefully this isn't treading on anyone's toes.

PR #332, opened 2018-04-05
Fixes to couple of CNN detector issues with the DLIB import changes

JayantPythonLover:
I am not sure if it is just me but with the DLIB import changes, I had multiple errors when using the CNN detector -- such as not extracting faces unless on HOG and sorting OOM'ing out on face-cnn sorts.

I've made the following changes that seem to resolve it for me.

@torzdf Please take a look and see if this is alright.

### DIFF
--- a/lib/cli.py
+++ b/lib/cli.py
@@ -252,6 +252,7 @@ class DirectoryProcessor(object):
             self.verify_output = True
 
     def load_filter(self):
+        import_FaceFilter()
         nfilter_files = self.arguments.nfilter
         if not isinstance(self.arguments.nfilter, list):
             nfilter_files = [self.arguments.nfilter]
@@ -263,7 +264,6 @@ class DirectoryProcessor(object):
         filter_files = list(filter(lambda fn: Path(fn).exists(), filter_files))
         
         if filter_files:
-            import_FaceFilter()
             print('Loading reference images for filtering: %s' % filter_files)
             return FaceFilter(filter_files, nfilter_files, self.arguments.ref_threshold)
 
--- a/tools/sort.py
+++ b/tools/sort.py
@@ -302,6 +302,7 @@ class SortProcessor(object):
         return img_list
 
     def sort_face_cnn(self):
+        import_face_recognition()
         import_FaceLandmarksExtractor()
 
         input_dir = self.arguments.input_dir
@@ -331,6 +332,7 @@ class SortProcessor(object):
         return img_list
 
     def sort_face_cnn_dissim(self):
+        import_face_recognition()
         import_FaceLandmarksExtractor()
 
         input_dir = self.arguments.input_dir
@@ -368,7 +370,8 @@ class SortProcessor(object):
             l = ( (fl[27][0]-fl[0][0]) + (fl[28][0]-fl[1][0]) + (fl[29][0]-fl[2][0]) ) / 3.0   
             r = ( (fl[16][0]-fl[27][0]) + (fl[15][0]-fl[28][0]) + (fl[14][0]-fl[29][0]) ) / 3.0
             return r-l
-            
+        
+        import_face_recognition()
         import_FaceLandmarksExtractor()
         input_dir = self.arguments.input_dir
     


lib/cli.py
```diff
@@ -263,7 +264,6 @@ def load_filter(self):
         filter_files = list(filter(lambda fn: Path(fn).exists(), filter_files))
         
         if filter_files:
-            import_FaceFilter()
```
torzdf: I still don't get why this would cause issues. I couldn't replicate but can test again tomorrow. 

By moving this to the top of load_filter, it means dlib will be launched for this on Extract regardless of whether we're filtering faces or not as self.process_arguments has:

`self.filter = self.load_filter()`

with no check on whether it's needed.

There may be a bug in the load_filter code though:

`        if filter_files:`
`            import_FaceFilter()`
`            print('Loading reference images for filtering: %s' % filter_files)`
`            return FaceFilter(filter_files, nfilter_files, self.arguments.ref_threshold)`

It looks like FaceFilter only gets called if there is a value for filter_files. If there is a value for nfilter_files and no value for filter_files then the FaceFilter doesn't get called :/

Unless I'm missing something, I think the check should probably be:
`if filter_files or nfilter_files:`

with my import_FaceFilter() remaining where I put it and the print statement updated to reflect the change.

I have never used this functionality, so may be off the mark

tools/sort.py
```diff
@@ -302,6 +302,7 @@ def sort_face_dissim(self):
         return img_list
 
     def sort_face_cnn(self):
+        import_face_recognition()
```
torzdf: I can't see that face_recognition is used in this method. Just FaceLandmarksExtractor

tools/sort.py
```diff
@@ -331,6 +332,7 @@ def sort_face_cnn(self):
         return img_list
 
     def sort_face_cnn_dissim(self):
+        import_face_recognition()
```
torzdf: I can't see that face_recognition is used in this method. Just FaceLandmarksExtractor

tools/sort.py
```diff
@@ -368,7 +370,8 @@ def calc_landmarks_face_yaw(fl):
             l = ( (fl[27][0]-fl[0][0]) + (fl[28][0]-fl[1][0]) + (fl[29][0]-fl[2][0]) ) / 3.0   
             r = ( (fl[16][0]-fl[27][0]) + (fl[15][0]-fl[28][0]) + (fl[14][0]-fl[29][0]) ) / 3.0
             return r-l
-
```
torzdf: I can't see that face_recognition is used in this method. Just FaceLandmarksExtractor

i5L4NDOF5T4BiLiTY:
I cant reproduce the issue either.
extract -D cnn -r on, extract -D cnn -r on -bt 8, sort -s face-cnn, all worked fine on a unmodified faceswapmaster I downloaded like two hours ago.

torzdf:
Ok, I've just hit a similar issue, but it's weird. On my version it extracted no faces. Then I ran on a rollbacked version, and it extracted so many false positives it was unreal. I tried again, and I got a ton of unspecified CUDA errors. 

I rebooted and then rollbacked version didn't extract any faces either.

I'm figuring the issue is with the CUDA install somewhere, or with the frames, because on another frameset it is extracting just fine.

I'm going to try some stuff to see if I can identify the issue, but at the moment it's looking like it isn't related specifically to this PR. I now need to find if it's specific to my/your install, the frames or something else.

torzdf:
^^ Ignore this. I've just rolled back to a version from long ago and I'm now having issues with previously working framesets. 

I think there must be something wrong with my install. Going to blow away my VM and start again, but I will continue to test

JayantPythonLover:
I've been going running through the code again to see exactly what is causing this issue for me.

Your changes seem fine to me and should work as it did before but clearly something is missing somewhere. Just trying to see if I can find a fix.

torzdf:
For what it's worth, something got borked with my install. After blowing everything away and re-installing everything is working as expected.

Issue #330, opened 2018-04-04
CAN I stop "training" after epoch "number"? automatically


LLALLA0925:
If  I do train.py -ep 400, this sequence stops at 399, and do NOTHING!!
I want to make train.py to stop( like I push "ENTER" or "ctrl + c" ) when epoch reachs to 400... 
and get out to CMD(i am using cmd version python) to use other COMMAND without "ENTER"...

I am not good at ENGLISH and python.
Can u help me?

andykdy:
I haven't tested this but it should work. 
First time submitting a PR.

torzdf:
@LLALLA0925 I have a PR currently up for testing (#367) that finishes training when target epochs are reached. It only works, however, if the -p (preview) flag is enabled. Feel free to test.

torzdf:
see #372

torzdf:
Fixed in latest commit for preview mode

PR #333, opened 2018-04-05
#330 Save Model with Epoch flag

andykdy:
#330 suggests that the network does not automatically save the model if an epoch is set manually via the --epoch flag.

The changes makes it so that the model is saved after the training loop.

### DIFF
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -204,7 +204,6 @@ class TrainingProcessor(object):
                     self.save_now = False
                     
             model.save_weights()
-            print('Training complete\nModel has been saved')
             exit(0)
         except KeyboardInterrupt:
             try:
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -197,12 +197,12 @@ class TrainingProcessor(object):
                     model.save_weights()
 
                 if self.stop:
-                    model.save_weights()
-                    exit()
+                    break
 
                 if self.save_now:
                     model.save_weights()
                     self.save_now = False
+                    
             model.save_weights()
             print('Training complete\nModel has been saved')
             exit(0)
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -203,7 +203,9 @@ class TrainingProcessor(object):
                 if self.save_now:
                     model.save_weights()
                     self.save_now = False
-
+            model.save_weights()
+            print('Training complete\nModel has been saved')
+            exit(0)
         except KeyboardInterrupt:
             try:
                 model.save_weights()


torzdf:
Makes sense. It would probably be worth also changing:

>                  if self.stop:
>                      model.save_weights()

exit()

to

>                  if self.stop:
>                      break

andykdy:
That makes sense. This is my first time making a PR
Can I edit this PR to include that change?

bryanlyon:
Just make changes to the same branch and it'll automatically update here.

My only suggestion is to remove the print, since the save_weights() already writes out a success message.

andykdy:
Changes have been made

LLALLA0925:
Thanks andykdy!! I used this new change. 
  for epoch in range(0, self.arguments.epochs):

                save_iteration = epoch % self.arguments.save_interval == 0

                trainer.train_one_step(epoch, self.show if (save_iteration or self.save_now) else None)

                if save_iteration:
                    model.save_weights()

                if self.stop:
                    break

                if self.save_now:
                    model.save_weights()
                    self.save_now = False
            model.save_weights()
            exit(0)
        except KeyboardInterrupt:
            try:
                model.save_weights()
            except KeyboardInterrupt:
                print('Saving model weights has been cancelled!')
            exit(0)
        except Exception as e:
            raise e
            exit(1)
I think this changes only def processThread(self):
but I want to finish all train.py to start convert.py and to fffmpeg automatically.

Someone said the reason why train.py not finish is 
train.py  is waiting "ENTER" or "ctrl + c"
maybe here part :
def process(self):
        self.stop = False
        self.save_now = False

        thr = threading.Thread(target=self.processThread, args=(), kwargs={})
        thr.start()

        if self.arguments.preview:
            print('Using live preview')
            while True:
                try:
                    with self.lock:
                        for name, image in self.preview_buffer.items():
                            cv2.imshow(name, image)

                    key = cv2.waitKey(1000)
                    if key == ord('\n') or key == ord('\r'):
                        break
                    if key == ord('s'):
                        self.save_now = True
                except KeyboardInterrupt:
                    break
        else:
            input() # TODO how to catch a specific key instead of Enter?
            # there isnt a good multiplatform solution: https://stackoverflow.com/questions/3523174/raw-input-in-python-without-pressing-enter

        print("Exit requested! The trainer will complete its current cycle, save the models and quit (it can take up a couple of seconds depending on your training speed). If you want to kill it now, press Ctrl + c")
        self.stop = True
        thr.join() # waits until thread finishes

I made cmd command (ffmpeg -> extract -> **train -> convert** -> ffmpeg) sequence
but sequence of ( ... **train -> convert** .. ) doesn't work.. 
you change and original code are just waiting key enter or ctrl+c...

can u help me??

bryanlyon:
@LLALLA0925 If you want to make it work just for you, remove the input() line.  It's not something we'll add to the main code, but it'll make yours run like you want.  I'm not sure you'll love the results on epoch only, but that at least will get you working.

andykdy:
Ahh, I think I see what you mean @LLALLA0925 
You want the entire process to be chained one after the other so that the entire pipeline is traversed automatically. 
Like bryanlyon said, I don't think that's within the scope of this repository and you'll have to make the change yourself

torzdf:
Just out of interest, is there any reason why you wouldn't want the process to exit when it reaches the target epoch?

bryanlyon:
I am fine with it exiting when it reaches the target epoch when the epoch is selected, but that change will also disable the ability to press "enter" to exit training for people not using epoch training.

andykdy:
Would that be the case with my changes? 
When '\n' is pressed the thread will chance 'self.stop' to True which will save the weights and stop training. 
If that's not the case, how could I change the code to accommodate the issue?

bryanlyon:
@andykdy Your changes are fine, I was referring to the changes I recommended to @LLALLA0925 of removing the input() line.  That change will work with his use case, but hurt everyone else's.  Your changes are fine so I'm going to merge them into staging for people to experiment with and verify.

I've merged this into staging. Closing it here, since it's in staging and will get promoted if there are no issues.

Issue #334, opened 2018-04-06
GUI breaks faceswap.py on non-display machines.


fat-tire:
I guess this is for @torzdf primarily.

A few quick observations on that recent [GUI pull](https://github.com/deepfakes/faceswap/commit/2c11eb164988b6beda5805d2c5034ac91982cc70):

1.  As stated in #325, Linux does crash without tk support in python installed  On Ubuntu/Mint/debian for example you'll need to do a `sudo apt install python3-tk`.

2.  You also get an error if you're running on a system without an active x session (such as w/ a headless server-- google cloud instance, etc.)

`_tkinter.TclError: no display name and no $DISPLAY environment variable`

Reverting that commit fixes the issue.  The error occurs even when you don't use the "gui" command.  Probably should do a few checks:

*  If the DISPLAY env variable isn't set, [skip the creation of the gui](https://github.com/deepfakes/faceswap/blob/2c11eb164988b6beda5805d2c5034ac91982cc70/faceswap.py#L29).

*  Ditto if the gui argument isn't there.  This should be runnable w/ any dependency on a ui session.  Could even bypass the import of tk if it isn't installed, no display, (or gui command isn't being used) as it uses memory...

Aside from that this looks like a cool addition.  Nice job.  Any reason you're not using a prettier toolkit?  Doesn't qt look native in Linux/Windows/Mac OS these days?

dfacct:
GUI also breaks faceswap.py if you aren't currently in the faceswap directory, which was never a requirement before.
`_tkinter.TclError: couldn't open "icons/open_folder.png": no such file or directory`

torzdf:
Of, thanks, I'll look to push commits to squash these today

torzdf:
@fat-tire 

> Aside from that this looks like a cool addition. Nice job. Any reason you're not using a prettier toolkit? Doesn't qt look native in Linux/Windows/Mac OS these days?

Yeah, there are definitely prettier tools out there. tkInter is used as it is part of the standard Distribution. I wanted to avoid adding more requirements where necessary. Whether it should still be part of the standard distro is another question. Ultimately, though, it gets the job done, even if it looks like it was built for WindowsXP ;)

Of course, if someone wants to make a GUI in qt, there is nothing stopping them adding another option ;)

fat-tire:
[Updated](https://github.com/deepfakes/faceswap/pull/336#issuecomment-379501812) #336 with some errors I saw.

fat-tire:
[...]

PR #336, opened 2018-04-06
GUI v0.1.0a

torzdf:
Please see #325 for more detailed information.

This PR addresses the issues described here: #334

Tested for non installation of python3-tk.
Tested for loading icons when launching outside of the faceswap directory
Cannot test for non-existence of X-Console as I do not have a machine to test on at the moment.

Also, major code refactor (under the hood to aid further development). Very basic error handling for the underlying faceswap.py process.

As before, bugs ahoy, please continue to report and I will continue to address.

### DIFF
--- a/scripts/gui.py
+++ b/scripts/gui.py
@@ -1,12 +1,14 @@
 import sys
-from threading import Thread
 import tkinter as tk
 
+from os import path
+from threading import Thread
 from tkinter import ttk
 from tkinter import filedialog
 
 from lib.cli import FullPaths
 from lib.Serializer import JSONSerializer
+
 
 class FaceswapGui(tk.Tk):
     ''' The Graphical User Interface '''
@@ -16,11 +18,15 @@ class FaceswapGui(tk.Tk):
 
         self.opts = options
         self.parser = parser
-        self.icofolder = tk.PhotoImage(file='icons/open_folder.png')
-        self.icoload = tk.PhotoImage(file='icons/open_file.png')
-        self.icosave = tk.PhotoImage(file='icons/save.png')
-        self.icoreset = tk.PhotoImage(file='icons/reset.png')
-        self.icoclear = tk.PhotoImage(file='icons/clear.png')
+        
+        pathscript = path.realpath(path.dirname(sys.argv[0]))
+        pathicons = path.join(pathscript, 'icons')
+        self.icofolder = tk.PhotoImage(file=path.join(pathicons,'open_folder.png'))
+        self.icoload = tk.PhotoImage(file=path.join(pathicons,'open_file.png'))
+        self.icosave = tk.PhotoImage(file=path.join(pathicons,'save.png'))
+        self.icoreset = tk.PhotoImage(file=path.join(pathicons,'reset.png'))
+        self.icoclear = tk.PhotoImage(file=path.join(pathicons,'clear.png'))
+
         self.helptext = tk.StringVar()
         self.statustext = tk.StringVar()
         self.serializer = JSONSerializer
--- a/scripts/gui.py
+++ b/scripts/gui.py
@@ -1,21 +1,49 @@
 import sys
-import tkinter as tk
 
 from os import path
 from threading import Thread
-from tkinter import ttk
-from tkinter import filedialog
 
 from lib.cli import FullPaths
 from lib.Serializer import JSONSerializer
 
+# Users without tkinter distrubtion packages or without an X-Console will error out
+# importing tkinter. Therefore run a check on these and only import if required 
+tk = None
+ttk = None
+filedialog = None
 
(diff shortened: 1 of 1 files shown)


fat-tire:
I'm still seeing this error w/o $DISPLAY set:

```
  File "faceswap.py", line 30, in <module>
    subparser, guiparsers, parser, "gui", "Launch the Faceswap Graphical User Interface.")
  File "/content/faceswap/scripts/gui.py", line 430, in __init__
    self.root = FaceswapGui(self.opts, self.parser)
  File "/content/faceswap/scripts/gui.py", line 49, in __init__
    self.gui = tk.Tk()
  File "/usr/lib/python3.6/tkinter/__init__.py", line 2020, in __init__
    self.tk = _tkinter.create(screenName, baseName, className, interactive, wantobjects, useTk, sync, use)
_tkinter.TclError: no display name and no $DISPLAY environment variable
```

It looks like check_display() only returns if DISPLAY is set [in cases where "gui" is the first argument](https://github.com/torzdf/faceswap/blob/gui/scripts/gui.py#L41).  In a headless server-type situation, you would NOT include "gui" as an argument, so check_display() always returns True (when gui isn't invoked, if that makes sense.)  In other words, in a headless situation, the check fails checking for "gui".  You shouldn't require "gui" to check for the DISPLAY being set.  put yet another way, the default behavior right now is to assume that DISPLAY *is* set unless proven otherwise by saying "gui" and NOT setting DISPLAY.

You might want to check if gui is any argument rather than just the first BTW.

Finally-- even when I DO add gui as an argument (hoping it would fail due to lack of DISPLAY) to force the check, I get failure on the DISPLAY check:

```
  File "faceswap.py", line 30, in <module>
    subparser, guiparsers, parser, "gui", "Launch the Faceswap Graphical User Interface.")
  File "/content/faceswap/scripts/gui.py", line 425, in __init__
    if not check_display(cmd) or not import_tkinter(cmd):
  File "/content/faceswap/scripts/gui.py", line 41, in check_display
    if command == 'gui' and not environ['DISPLAY']:
  File "/usr/lib/python3.6/os.py", line 669, in __getitem__
    raise KeyError(key) from None
KeyError: 'DISPLAY'
```

Might you have meant `environ.get('DISPLAY')` rather than `environ('DISPLAY')`?  You can check if it's set to `None` to avoid this crash.  You might debug this by trying something other than DISPLAY (like TESTDISPLAY) that you know has NOT been set.

Hope this helps.

bryanlyon:
One way to fix this would be to encase the call in a try except block.  Make sure you're only capturing the errors that match and let the other errors through.

fat-tire:
I don't think you have to go that far.  Assume that gui is not enabled unless proven otherwise, then:

if any parameter == "gui" and 'DISPLAY' in environ:
    (now import gui stuff and create object...)

torzdf:
Thanks for the feedback. Annoyingly it was the one thing I couldn't test (although I did echo $DISPLAY on a headless machine to see what I would get).  Full Traceback is useful.

torzdf:
Hopefully latest commit got it. Annoyingly I would have caught that I had forgotten to split the GUI check from the DISPLAY check if I'd tested. I only include the GUI check to spit out a useful error message.

I still can't test, but please let me know :)

fat-tire:
Looking forward to it!   Request:  could you add a flag to output that graph to a file?   Request #2-- I think [tensorboard](https://www.tensorflow.org/programmers_guide/summaries_and_tensorboard) counts as GUI right? Maybe add a flag to dump a log every so often so TB can be attached and progress monitored that way too?  Cheers!

torzdf:
I'm trying to avoid faceswap.py having any dependencies on the GUI, or faceswap.py having to have any code to specifically support it. Obviously there is only so far I can go with this approach!

I will look into tensorboard, but in the meantime I'm just looking to generate a graph with matplotlib based on the loss values output to console, as these are trivial to hook into. Saving the graph from the GUI should be straightforward, but it won't work without the GUI. Porting the code to some kind of file writer as part of train.py should be straightforward though, so it could be looked into if we decided that was the way to go.

torzdf:
@bryanlyon @Clorr Is it possible to merge this, as it's a bugfix that will effect anyone using faceswap.py in an headless environment.

Thanks

johnc2k:
I am still getting the message "Could not detect a display. The GUI has been disabled" on Windows 10.
Im assuming I need to define the $DISPLAY variable in Windows?

torzdf:
Arggh. Annoyingly I found this issue when testing today. I can push a fix, or you can ammend gui.py in scripts from:

`if not environ.get('DISPLAY', None):`

to:

`if not os.environ.get('DISPLAY', None) and os.name != 'nt':`

(Line 47)

johnc2k:
That fixed it thanks

PR #341, opened 2018-04-11
Fixed GAN128 multi gpu issue.

leftler:
Added explicit names to netGA and netGB's outputs to fix the name conflict error when using multiple gpus. 

This has been tested on a single GTX 1080 with no `-g` flag passed in and has been tested on a Azure NC12 VM with two K80 cards, both with no `-g` flag passed and with `-g 2`

This is replacing the assert added to solve #287

### DIFF
--- a/plugins/Model_GAN128/Model.py
+++ b/plugins/Model_GAN128/Model.py
@@ -45,7 +45,6 @@ class GANModel():
     def __init__(self, model_dir, gpus):
         self.model_dir = model_dir
         self.gpus = gpus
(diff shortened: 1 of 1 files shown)


torzdf:
Can someone with multi gpus test this? I only have one unfortunately.

bryanlyon:
I haven't tested (and can't due to vram) but changing the output names explicitly makes sense.  This bug was completely preventing execution, so if it executes, it should be a good add.

torzdf:
Ok, I think we should merge then, we can always revert if there are issues. It's one of the reasons why I think it would be useful to have a Staging branch though.

PR #349, opened 2018-04-13
Model backups before saving

fakinated:
This adds a backup mechanism - the model directory gets renamed into `<previous name>_bk` before saving weights and sorts images before processing, so the frame range parameter (`-fr`) makes sense (before the order of images would depend on the implementation of the file system/OS).

### DIFF
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -4,6 +4,7 @@ from os.path import basename, exists
 
 from pathlib import Path
 from scandir import scandir
+import os
 
 image_extensions = [".jpg", ".jpeg", ".png", ".tif", ".tiff"]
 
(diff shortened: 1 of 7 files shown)


torzdf:
I don't use the frame range parameter. What was the effect prior to sorting images, and what does this fix do?

fakinated:
The frame range parameter, when specified as `-fr a-b`, only processes the ath to the bth frame - however the order of the frames was completely arbitrary before I sorted the list of images.

The script would just trim the list of images as such: `imgs[a:b]` which doesn't give any benefit if you don't know the order in which the images are read.

torzdf:
Ok, thanks, I figured as much, so I figure it was just trimming random frames before. I guess this feature can't be used too much if no one noticed ;)

fakinated:
Exactly. But how the images end up in the list depends on the operating system and maybe even the file system. As I'm using Linux and the files lie on an ext4 partition, it might possibly work with e. g. Windows and ntfs...

PR #352, opened 2018-04-15
GUI v0.2.1b

torzdf:
Thanks to some help from @AbysmalBiscuit  and feedback from users I now have version 0.2.1 ready for merging. There is quite a lot added.

For background see here: #325. In a nutshell this is a cross-platform GUI for faceswap.py

#### New since last version:

- More robust subprocess handling
- Layout changes
- GUI support for tools.py (thanks @AbysmalBiscuit)
- Tooltips
- Graphing (finally!) - You will need to install the matplotlib python module
- Console out within GUI
- Preview pane within GUI
- Size adjustable panels

Pressing 'Enter' on the Preview Window will not work within the GUI. You must press the 'Terminate' button to stop and save (It may look like it's hanging the GUI, but that's just it doing it's thing).

Run it with:

`python faceswap.py gui`

When posting bug reports, please include the full traceback (if relevant). You should be able to pull it from the embedded console out, but if you can't the gui can be run in debug mode as:

`python faceswap.py gui -d`

This will print the console out to the main terminal window instead of the embedded window in the GUI.

#### Known issues:

- DPI for high resolution monitors (will look at for next release)
- Sometimes graphs don't reset when training is terminated. This is annoying but not game changing (will look at for next release)
- ToolTips don't work on macOS (will look at for next release)
- FIXED: ~~~GAN not graphing.~~~
- FIXED: ~~~Warning output when generating graphs~~~
- FIXED: ~~~Windows version not saving model when terminating training~~~
- FIXED: ~~~tools.py did not actually load the GUI~~~
- USER ACTION REQUIRED. ~~~~macOS fails to detect DISPLAY.~~~~ macOS needs to have xquartz installed to work. Will add hint in error message
- FIXED: ~~~macOS locks up when changing tabs.~~~

#### Feature Requests and possible implementations (to be looked at for next release):

- Extract video frames and recompile swapped frames back to video
- Include trend stats with graph (e.g average loss over each quarter, rolling average etc)
- For the GAN trainer, add Preview tabs for: Masked, Raw, and Mask Previews
- Keep a continuous graph of all training activity (maybe store data in a sqlite db?)
- Current session, and all time spent training record (e.g. 24 hours current session ; 160 hours total)
- Save graphs to file
- Zoom in/out graphs
- Add progress bar
- Load tools.py into the main faceswap.py gui

#### Screenshots:
Windows 10 Native:
w10

macOS Native:
screen shot 2018-04-15 at 10 31 56

X-Forwarding on Windows from Ubuntu:
gui_preview1

gui_preview2

GAN graphing:
splitgraph


AbysmalBiscuit:
@torzdf I'm already a contributor, so no worries about that. :)

I'm also finishing up some small improvements for the GUI and a new tool that will be an ffmpeg wrapper for the common tasks that have to be done for faceswapping, so I'll be contributing quite a bit shortly.

Issue #305, opened 2018-03-21
IAE model "swap-model" option bug


woshin:
https://github.com/deepfakes/faceswap/blob/e0fa8c071ebfaf249ff13b2e3002adb9f5ca6715/plugins/Model_IAE/AutoEncoder.py#L25

Clorr:
Feel free to make a correction directly in Github

torzdf:
Fixed in #353

PR #353, opened 2018-04-16
Fix IAE's swap bug

bryanlyon:
IAE has a swap bug detailed in #305 where the models aren't swapped.  Fixed here.


torzdf:
This commit removes the backup model directory from #349 

Although in retrospect, I think backup model dir should probably be a util that gets called to reduce redundancy

bryanlyon:
It shouldn't interfere with #349 at all.  That PR only touches the save and this PR only touches the load.  It doesn't even change the files, it simply makes sure that when swap is enabled it loads A as B and B as A.  In fact, both sets of code should ever run in a single operation since only convert uses the swap flag and it doesn't ever save.  This is actually how swap is handled in all other models, it was just incorrectly broken in IAE.

plugins/Model_IAE/AutoEncoder.py
```diff
@@ -39,11 +38,6 @@ def load(self, swapped):
             return False
 
     def save_weights(self):
-        model_dir = str(self.model_dir)
```
torzdf: Maybe I'm misreading this, but it looks like it's pulling out this code, which was just merged in. If I'm not misreading, then I think it's just some over-zealous conflict resolution ;)

bryanlyon:
There are more diffs now since #349 was merged in.  There was only one line changed in my original, the bugfix is only the line (face_A,face_B) = (inter_AH5, inter_BH5) if not swapped else (inter_BH5, inter_AH5)

bryanlyon:
I've updated the branch with the merges so it's using the same base.  It should show correct now.

PR #363, opened 2018-04-21
Adding new tool effmpeg ("easy"-ffmpeg) with gui support. Extend gui functionality to support filetypes.

AbysmalBiscuit:
Adding a new "easy"-ffmpeg tool with gui- support.
Extend gui functionality to support filetypes when prompting user to select a file, and add new file selection class that allows to switch between choosing a folder and a file based on selected action.

`effmpeg` provides the following functionality:
* extract: extract video into individual frames
* gen-vid: generate a video when given individual frames (with the option to mux in audio during generation using a reference video)
* get-fps: prints the video fps to the terminal
* get-info: prints all the information about the video (includes all meta-data)
* mux-audio: muxes audio from reference video into the input video (output is saved as a separate video)
* rescale: rescales the input video to a new resolution using the provided 'scale'
* rotate: rotates the video using the chosen transpose variable or degrees (if transpose is set, then degree rotation is ignored)
* slice: cuts out a chosen portion of the input video into a new video using the start time and either the stop time or duration (if both duration and end time are set, then duration will be ignored and the end time will be used)


torzdf:
Thanks for this, going to start testing tomorrow.

I've actually completely re-written the lib.cli and scripts section, so I'm going to try to merge both of these into staging at the same time. I would like to get your thoughts on my changes (will raise the PR tomorrow), as whilst I think it's a significant improvement, I've got to the point where I can't see the wood from the trees, so there is likely scope to improve more.

AbysmalBiscuit:
@torzdf sure I'd be happy to give your changes a look.

From the changes I've made to `lib/cli.py` the only ones needed to make the gui and effmpeg function properly are the classes I have added to extend the `FullPaths` class (`DirFullPaths`, `FileFullPaths` and `ComboFullPaths`), and the addition of `"action": FileFullPaths,` to the `alignments_file` argument in the general argument processing class found in `lib/cli.py`.

torzdf:
ffmpy needs to be added to requirements files

torzdf:
Next time I'll try testing prior to merging into staging so a new PR isn't required

AbysmalBiscuit:
@torzdf 

I'll look into these issues, no worries about me having to open a new PR.

I hadn't yet added ffmpy to the dependencies/dependency checks similar to the gui since this was just an alpha test to help me find bugs that I may have missed. :)

As for the video generation I'll try to figure it out. The issue is limited only to gui since this is a tkinter file selection thing.

torzdf:
Couldn't you just use a filedialog.asksaveasfile ?

AbysmalBiscuit:
That's what I meant by using the file saving.

PR #371, opened 2018-04-23
Model backup bugfix

torzdf:
Windows sometimes has permission issues with creating/deleting new directories. This bugfix keeps the backup files within the original model directory by appending .bk onto the end of the model files.

This also ensures that any user files placed within the model directory do not get lost.

### DIFF
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -1,6 +1,6 @@
 import cv2
 import sys
-from os.path import basename, exists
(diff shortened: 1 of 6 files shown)


Kirin-kun:
Crashes if no model files are present.

torzdf:
Ok, I've rolled back for now. I'll look into it and re-push

PR #388, opened 2018-05-07
Overlooked converter code added

andenixa:
Added some lines to coveter that allow using the new model


torzdf:
Can you fix the conflicts please.

andenixa:
@torzdf yes so it seems, I think I'd rather remove this pr and re-do the alteration if that is an option.

torzdf:
Yeah, no problem. I'll close this one, so just resubmit when you're ready.

PR #396, opened 2018-05-13
Load tools.py into the main faceswap.py gui

LordOEQJen:
As mentioned in the to-do-list here:
https://github.com/torzdf/faceswap/projects/1

Adds sort and effmpeg to the menu 
dsktp

### DIFF
--- a/faceswap.py
+++ b/faceswap.py
@@ -2,6 +2,8 @@
 """ The master faceswap.py script """
 import sys
 
+from tools.sort import SortProcessor
+from tools.effmpeg import Effmpeg
 import lib.cli as cli
 
 if sys.version_info[0] < 3:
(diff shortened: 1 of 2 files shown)


AbysmalBiscuit:
It works. There are some small things that I would change that will make it easier to maintain/extend in the future, I'll do a code review to point them out.

The reason why initially tools were put into a separate file was because the collaborators wanted to keep the main faceswap functionality and the tools separate.
@Clorr (and other collaborators) are you fine with merging both the core faceswap functionality and tools into one command file?
As an alternative we could add a `gui.py` or `faceswap-gui.py` file to the root directory and then keep the `faceswap.py` and `tools.py` as cli only (I've tested commenting out the gui stuff from them and it works fine as cli only).

torzdf:
I appreciate the help on this guys, but I'd advise that you hold fire for now.

As I tend to do, I have pretty much rewritten the GUI from the ground up, so the structure is significantly different. If there is a burning requirement for this, it can get merged, but it is likely to be overwritten when I implement the next version.

torzdf:
And actually, having a quick review, this adds the tools to the main faceswap code. There are arguments to be made either way on whether this is desired, but @Clorr and others were not so keen on this, wishing to separate out the main faceswap workflow from helper functions. I'd prefer to respect their wishes on this, and maybe review again in future.

torzdf:
@AbysmalBiscuit 

> As an alternative we could add a gui.py or faceswap-gui.py file to the root directory and then keep the faceswap.py and tools.py as cli only (I've tested commenting out the gui stuff from them and it works fine as cli only).

As the options have now been moved to dicts, my plan is to remove the gui specific code from cli.py, faceswap.py and sort.py and just import the dicts straight into the gui.

PR #402, opened 2018-05-17
Add GPU selection for convertion

bonlime:
Right now there is no explicit way to select GPU to use for convertion. By default GPU 0 is taken, which leads to OOM error if it is busy
Adding one argument will help


bonlime:
Sorry, incorrect branches for pull request

bryanlyon:
There is already a supported way to do this.  See https://devblogs.nvidia.com/cuda-pro-tip-control-gpu-visibility-cuda_visible_devices/

This works on Windows and Linux.  We decided not to duplicate it in faceswap due to complications such as MultiGPU systems, virtual GPUs, and other problems that occur in edge cases.  Please use the CUDA_VISIBLE_DEVICES for this functionality.

PR #418, opened 2018-06-12
OriginalHighRes 128 model update

andenixa:
Required for OriginalHighRes Model to function

### DIFF
--- a/plugins/Model_OriginalHighRes/Trainer.py
+++ b/plugins/Model_OriginalHighRes/Trainer.py
@@ -1,13 +1,11 @@
 
 import time
-
 import numpy
 
 from lib.training_data import TrainingDataGenerator, stack_images
 
 
 TRANSFORM_PRC = 115.
-#TRANSFORM_PRC = 150.
 
 
 class Trainer():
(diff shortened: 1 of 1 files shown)


andenixa:
@Kirin-kun thats confirmed for 6GB cards. Shall be fixed ASAP.

andenixa:
@Kirin-kun I apologize for not testing with 6Gb card prior to make a PR.
Currently I made a PR to staging but you are welcome to test is. I've am able to train it at 6GB gforce gtx9800ti with maximum batch_size of 12. That should be sufficient for good results. 
I could squeeze memory requirements much further, but my intentions were to provide a full scale 128x128 model and I don't want to make a crippled version of it.

Kirin-kun:
Is your original encoder 128 better than, or equivalent to, Iperov's H128?

I got pretty good results with his model.

andenixa:
@Kirin-kun I don't think you have Iperov's H128 with faceswap. You need to use DeepFaceLab for that.
I should think Iperov models are better, specially F128 ones. Definitely faster with his new K.function design. Then again you need another repo to use it. I don't know for sure because I never used his pure H128, 
I made an HD mod with 192x192 tensor size and Shaonlu activation based on Iperov's H128 so I can't compare to the original one (because I don't have any trained Iperov's H128 on hands).

Issue #437, opened 2018-06-21
convert.py doesn't work with OriginalHighRes trainer


agilebean:
## Expected behavior
Training with OriginalHighRes plugin module works like other trainer plugins

## Actual behavior
Training with OriginalHighRes plugin module throws the error:
 `TypeError: join() argument must be str or bytes, not 'PosixPath'` with the following
stack trace:
````
Using TensorFlow backend.
Output Directory: /home/agilebean/faceswap/output
Input Directory: /home/agilebean/faceswap/input_images
Loading Extract from Extract_Align plugin...
Using json serializer
Alignments filepath: /home/agilebean/faceswap/input_images/alignments.json
Aligned directory not specified. All faces listed in the alignments file will be converted
Loading Model from Model_OriginalHighRes plugin...
Traceback (most recent call last):
  File "faceswap.py", line 36, in <module>
    ARGUMENTS.func(ARGUMENTS)
  File "/home/agilebean/faceswap/lib/cli.py", line 81, in execute_script
    process.process()
  File "/home/agilebean/faceswap/scripts/convert.py", line 42, in process
    model = self.load_model()
  File "/home/agilebean/faceswap/scripts/convert.py", line 70, in load_model
    if not model.load(self.args.swap_model):
  File "/home/agilebean/faceswap/plugins/Model_OriginalHighRes/Model.py", line 134, in load
    state_dir = os.path.join(self.model_dir, 'state_{version_str}_{ENCODER.value}.json'.format(**globals()))
  File "/usr/lib/python3.5/posixpath.py", line 89, in join
    genericpath._check_arg_types('join', a, *p)
  File "/usr/lib/python3.5/genericpath.py", line 143, in _check_arg_types
    (funcname, s.__class__.__name__)) from None
TypeError: join() argument must be str or bytes, not 'PosixPath'
````

## Steps to reproduce

TRAINER=OriginalHighRes
python3 faceswap.py train -A  $INPUT_FACE_DIR -B  $STYLE_FACE_DIR -it $NUM_ITERATIONS -t $TRAINER

I tested with all other methods, OriginalHighRes is the only method that doesn't work. 

## Other relevant information
- **Operating system and version:** Ubuntu 16.04 LTS
- **Python version:** 3.5
- **Faceswap version:** 
- **Faceswap method:** GPU

oracle9i88:
I testing it under win10
python=3.5
GPU

also is TyperError :。。。。，not ‘Winpath’

[...]

PR #442, opened 2018-06-22
convert.py doesn't work with OriginalHighRes trainer bugfix

andenixa:


torzdf:
#437

agilebean:
@torzdf will this Windows 10 fix also fix the different problem on Ubuntu?

PR #452, opened 2018-06-28
slight model re-factoring

andenixa:
- removed excess threading code
  - added random kernel initialization to dense layer

added (optional) DDSIM loss function (requires keras_contrib package) which allegedly should give a better perceptual training quality


andenixa:
Is it going to be merged?

PR #464, opened 2018-07-21
Slight OriginaHightRes Model re-factoring

andenixa:
- Kernels are now initialised once
- Might fix red screen of death on multi-gpu configurations

### DIFF
--- a/plugins/Model_OriginalHighRes/Model.py
+++ b/plugins/Model_OriginalHighRes/Model.py
@@ -44,6 +44,9 @@ class EncoderType(enum.Enum):
     SHAOANLU = "shaoanlu"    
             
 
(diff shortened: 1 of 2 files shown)


kellurian:
Works fine for me at 128x128, still get error "ValueError: Error when checking input: expected input_4 to have shape (256, 256, 3) but got array with shape (128, 128, 3)" when I change to 256, but you might not have implemented any changes for increasing the training resolution.

andenixa:
@kellurian my bad, done now but so far it would only support `INPUT_SIZE` of 
- 64, 64
- 128, 128
- 256, 256
I have to look into the code if the trainer library supports intermediate values here, i.e. 176x176

kellurian:
Awesome sauce. Thank you.

PR #477, opened 2018-08-20
alignments.json save interval for extracting

Tvde1:
Added an optional parameter `-si` or `--save-interval` which will save the alignments.json after the set amount of frames.

So `python faceswap.py extract ... -si 5` will save the file after 5 frames.
It will also skip the already extracted files when it's restarted.

### DIFF
--- a/lib/cli.py
+++ b/lib/cli.py
@@ -5,6 +5,7 @@ from importlib import import_module
 import os
 import platform
 import sys
+from math import inf
 
 from plugins.PluginLoader import PluginLoader
 
@@ -409,6 +410,16 @@ class ExtractArgs(ExtractConvertArgs):
                               "help": "Perform extra alignment to ensure "
                                       "left/right eyes are  at the same "
                                       "height"})
+        argument_list.append({"opts": ("-si", "--save-interval"),
+                              "dest": "save_interval",
+                              "type": int,
+                              "default": inf,
+                              "help": "Automatically save the alignments file "
+                                      "after a set amount of frames. Will only "
+                                      "save at the end of extracting by "
+                                      "default. WARNING: Don't interrupt the "
+                                      "script when writing the file because "
+                                      "it might get corrupted."})
         return argument_list
 
 
--- a/lib/cli.py
+++ b/lib/cli.py
@@ -5,7 +5,6 @@ from importlib import import_module
 import os
 import platform
 import sys
-from math import inf
 
 from plugins.PluginLoader import PluginLoader
 
@@ -431,7 +430,7 @@ class ExtractArgs(ExtractConvertArgs):
         argument_list.append({"opts": ("-si", "--save-interval"),
                               "dest": "save_interval",
                               "type": int,
-                              "default": inf,
+                              "default": None,
                               "help": "Automatically save the alignments file "
                                       "after a set amount of frames. Will only "
                                       "save at the end of extracting by "
--- a/scripts/extract.py
+++ b/scripts/extract.py
@@ -27,7 +27,7 @@ class Extract(object):
         self.output_dir = self.faces.output_dir
 
         self.export_face = True
-        self.save_interval = self.args.save_interval
+        self.save_interval = self.args.save_interval if hasattr(self.args, "save_interval") else None
 
     def process(self):
         """ Perform the extraction process """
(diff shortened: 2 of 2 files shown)


lib/cli.py
```diff
@@ -5,6 +5,7 @@
 import os
 import platform
 import sys
+from math import inf
```
torzdf: Best not to have this import as it's not necessary. Make the default `None `and handle saving if a value is passed

lib/cli.py
```diff
@@ -427,6 +428,16 @@ def get_optional_arguments():
                               "help": "Perform extra alignment to ensure "
                                       "left/right eyes are  at the same "
                                       "height"})
+        argument_list.append({"opts": ("-si", "--save-interval"),
+                              "dest": "save_interval",
+                              "type": int,
+                              "default": inf,
```
torzdf: Change to None

scripts/extract.py
```diff
@@ -44,19 +45,33 @@ def process(self):
...
         self.faces.num_faces_detected = faces
 
+    def write_alignments(self):
+        self.alignments.write_alignments(self.faces.faces_detected)
+
+
     def extract_single_process(self):
         """ Run extraction in a single process """
+
+        alignments = self.alignments.read_alignments()
```
torzdf: Why are you reading alignments here? It shouldn't exist at this point. If it's to continue off from where previously left off, then we have 'skip existing' for that.

scripts/extract.py
```diff
@@ -44,19 +45,33 @@ def process(self):
...
+        
+        frame_no = 0
         for filename in tqdm(self.images.input_images, file=sys.stdout):
+            basename = os.path.basename(filename)
+            if basename in alignments:
```
torzdf: See above. This looks like it's duplicating skip-existing code.

scripts/extract.py
```diff
@@ -44,19 +45,33 @@ def process(self):
...
+                continue
             filename, faces = self.process_single_image(filename)
-            self.faces.faces_detected[os.path.basename(filename)] = faces
+            self.faces.faces_detected[basename] = faces
+            frame_no = frame_no + 1
```
torzdf: `frame_no += 1`

scripts/extract.py
```diff
@@ -44,19 +45,33 @@ def process(self):
...
+            if frame_no == self.save_interval:
```
torzdf: `if hasattr(self.args, "save_interval") and self.args.save_interval == frame_no`

Tvde1:
Thanks for the feedback. I've changed the stuff you requested

scripts/extract.py
```diff
@@ -27,6 +27,7 @@ def __init__(self, arguments):
         self.output_dir = self.faces.output_dir
 
         self.export_face = True
+        self.save_interval = self.args.save_interval
```
torzdf: If extraction is being run as part of convert (i.e. no alignments file exists when someone runs convert) then `save_interval` won't exist in `arguments` and the code will error out, so remove this line and see below.

scripts/extract.py
```diff
@@ -44,22 +45,31 @@ def process(self):
...
+
+
     def extract_single_process(self):
         """ Run extraction in a single process """
+        frame_no = 0
         for filename in tqdm(self.images.input_images, file=sys.stdout):
             filename, faces = self.process_single_image(filename)
             self.faces.faces_detected[os.path.basename(filename)] = faces
+            frame_no += 1
+            if frame_no == self.save_interval:
```
torzdf: As per earlier comment, self.save_interval will not necessarily exist, so reference it directly from arguments and make sure it is a passed argument: `if hasattr(self.args, "save_interval") and self.args.save_interval == frame_no`

scripts/extract.py
```diff
@@ -68,6 +78,10 @@ def extract_multi_process(self):
                 file=sys.stdout):
             self.faces.num_faces_detected += 1
             self.faces.faces_detected[os.path.basename(filename)] = faces
+            frame_no += 1
+            if frame_no == self.save_interval:
```
torzdf: As per earlier comment, self.save_interval will not necessarily exist, so reference it directly from arguments and make sure it is a passed argument: `if hasattr(self.args, "save_interval") and self.args.save_interval == frame_no`

scripts/extract.py
Tvde1: I added the check here so we don't have to keep looking if the arg exists. I don't know if it matters much with performance but it's a tiny bit optimization I guess

PR #479, opened 2018-08-21
fix group_blur and group_face_yaw

Tvde1:


Tvde1:
Essentially reverting this bad change
image

Tvde1:
`idx` should loop through all the images. Not restart and fill every bucket with the same set of images.

PR #482, opened 2018-08-26
this skips all pictures in alignments.json

Tvde1:
I want to run extract again, but output to a different folder, for easy cleaning later.

What -s does now, is skip all the files in the output folder.

This will also remove all the files from the alignments.json.

### DIFF
--- a/scripts/extract.py
+++ b/scripts/extract.py
@@ -21,6 +21,7 @@ class Extract(object):
         self.args = arguments
 
         self.images = Images(self.args)
+        self.input_images = self.images.input_images
         self.faces = Faces(self.args)
         self.alignments = Alignments(self.args)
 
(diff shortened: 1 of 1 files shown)


torzdf:
Sorry, I'm confused by what is being attempted to be achieved here.

My understanding of skip existing is to add faces to alignments file where some faces have already been extracted, so you wouldn't want to remove files from the alignments folder.

torzdf:
I'll have to think about this one, because it changes the way it currently works. 

I don't use skip existing, but I *think* that the general use case is for people to be able to extract faces to the same alignments file in batches

`-s, --skip-existing   Skips frames that have already been extracted`

I'd need to play around with it a bit to see how it's meant to work and what the desired outcome would be.

Tvde1:
I had a video of 3000 frames and after the first run, I sorted through all of them etc. And then I noticed some frames were missing.

So I want to run it again and add the frames to the same alignments file.

Only I don't want to have to sort the whole 3000 frames again just for the 100 or something that were new. So I set the output folder to `../extracted_2`.

Tvde1:
I also am in the middle of extracting a video. While it was extracting I deleted pictures of other faces I didn't need.
And when I run extract again it will redo those frames even when they are still in the alignments file.

Tvde1:
This has been added with your new commits right?

Tvde1:
Hm never mind

torzdf:
This should be do-able now in the extract_mp branch (I will look to merge to staging soon).

The -s and (new) -sf options no longer look at the output folder, but instead look at the alignments file. If you specify a different output folder on your second run, you should be good to go.

PR #483, opened 2018-08-26
fix missing-alignments

Tvde1:


torzdf:
Closing as I'm not sure what problem it's solving

Tvde1:
Yeah but it kept crashing for me

Tvde1:
item has been changed to a list of 4 items, the alignment name, extention, filename, and frame index

torzdf:
I think you were using it for the wrong purpose.... depending whether you supply a frames dir or a faces dir, items changes.

For frames items is just a list of filenames, for frames, it is those 4 items listed above.

Missing Alignments is meant for checking against frames, not faces. By removing "exit(0)" from line 565 you basically allowed it to accept a faces folder, hence the crash.

Tvde1:
Oh that is well odd. I think it'd be a good idea to split it up some day.

PR #505, opened 2018-09-20
ignore the pycache files

Tvde1:


torzdf:
I'm confused. pycache is already ignored:

image

Tvde1:
Well my git client showed the temporary files as unstaged. It's also not in the `.gitignore` yet.

AbysmalBiscuit:
@torzdf I think line 22 in the `.gitignore` file may un-ignore the nested __pycache__.

Perhaps changing the commit such that it would apply to all __pycache__ directories would be better, i.e. as suggested [here](https://github.com/github/gitignore/blob/master/Python.gitignore):
```python
__pycache__/
```

.gitignore
```diff
@@ -20,3 +20,5 @@
 !plugins
 !tools
 !tools/lib*
+
+tools/lib_alignments/__pycache__/
```
AbysmalBiscuit: Change this to:
```python
__pycache__/
```
In order to ignore all __pycache__ directories, no matter where they are.

PR #506, opened 2018-09-20
OriginalHighRes model plug-in re-factoring

andenixa:
- Added `HighRes` encoder suitable for training high resolution outputs, i.e. 192x, 256x
- Added `Standard` encoder, a high quality implementation of FS standard autoencoder
- Added optional `DSSIM` loss function to improve training quality; doesn't break current saves if enabled
- Added optional `Subpixel Upscaling` to improve quality; doesn't break current saves / training if enabled
- Major re-factoring to improve stability

### DIFF
--- a/lib/training_data.py
+++ b/lib/training_data.py
@@ -14,8 +14,8 @@ class TrainingDataGenerator():
 
     def minibatchAB(self, images, batchsize):
         batch = BackgroundGenerator(self.minibatch(images, batchsize), 1)
-        for ep1, warped_img, target_img in batch.iterator():
-            yield ep1, warped_img, target_img
+        for ep1, warped_img, target_img, file_names, in batch.iterator():
+            yield ep1, warped_img, target_img, file_names
 
     # A generator function that yields epoch, batchsize of warped_img and batchsize of target_img
     def minibatch(self, data, batchsize):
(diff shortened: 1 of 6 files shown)


andenixa:
Please try it out as Original 256x mode was highly requested of me.

gessyoo:
Thanks andenixa, will give it a try and post result pics.

Eleganze:
Hi,

Tried the 256 version on a dual 1080ti, but only got OOM. Tried adjusting batch size down to 16, with no luck.  Not sure if I did everything right, though. (Changed encoder from original to highres)

Let me know and I will be happy to try out whatever setup you like.

gessyoo:
Eleganze, image size 192x192 works on a single 1080ti, with -bs 32.  Need to change line 70 in Model.py:   IMAGE_SHAPE = 192, 192 # image shape (or 256, 256, defaults to 128, 128).

Eleganze:
Gessyoo, thanks a lot! Got it to work. I had a program running in the background stealing memory - my bad.
I got 192 to work with bs 32, but only when activating 1 gpu. The weird thing is that when I watch the GPU usage, it shows full memory usage from both cards, but only processing usage on 1 of them. Not sure if this is normal, or if its something with the load distribution.

I got 256x256 up and running with bs 8, but it crashed after 40 iterations.

gessyoo:
I've got a 1080ti and a 1070, but 2 gpus (-g 2) only seems to work with extracting, not with training.  Are you running the two1080ti's in SLI?  I thought SLI was needed to train with 2 GPUs.

Eleganze:
Running 192x192 with two cards now (processing on both) bs 64 and activated DSSIM loss. I will leave it running for a day, and see how the results are. After 32 minutes, I’m at 850 iterations and down to around 0,035.

gessyoo:
I've used DFL lab as well.  Have you tried model H128?  Here's one: https://youtu.be/Cs1SkB_8PEo.  FS has the ability to do manual alignments, but it crashes when I try to run it.

torzdf:
> 
> 
>  FS has the ability to do manual alignments, but it crashes when I try to run it.

Do you have any further info on this? I would like it to not be crashing!

@Eleganze I'm currently working on Landmark smoothing which may go some way to sort your issue.

Eleganze:
@Torzdf Thanks! I will check it out. I have problems to get anything else but the trainer to run right now.  Seems to be related to my python version, and how strings should be written in the code.

PR #510, opened 2018-09-28
Fixed face extract with dlib on OS X

1132719438:
- Use pynvx to get GPU info on OS X
- Fixed face extract with dlib


torzdf:
Great work! I can test this from a non-macOS perspective.

1132719438:
@torzdf No functionality is missing for faceswap,  the only reason is that pynvx has not been widely tested on Linux and Windows, so I want to minimize impact to Linux and Windows users. Maybe we can switch to pynvx when it is fully tested.

torzdf:
Ok, I'll get this pushed to staging soon. I just realised that you authored pynvx (presumably just for this purpose!), so thanks for that.

PR #512, opened 2018-09-28
Add timelapse functionality

Tvde1:
So how this works:

if you supply both `--timelapse-input-A` and `--timelapse-input-B` (and optionally `--timelapse-output`), it will look in those folders for files, and every iteration save a preview with the same faces in the same order.

I have yet to create a preview. I think it'll look fun seeing how the model starts picking up certain details.
Also if none of the parameters are given, it will not cost any cpu except for checking if an object is null which I think the compiler can optimise (or else I'd optimise it with a simple boolean).

### DIFF
--- a/lib/cli.py
+++ b/lib/cli.py
@@ -688,7 +688,6 @@ class TrainArgs(FaceSwapArgs):
                               "dest": "timelapse_output",
                               "default": None,
                               "help": "The output folder for the timelapse"})
-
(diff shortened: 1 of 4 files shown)


torzdf:
This is a cool feature, but I haven't had a chance to test it (and am unlikely to in the near future).

If anyone else is using this, please could you feedback with your results.

PR #513, opened 2018-09-28
Add a cli option to remove files after extract

Tvde1:


torzdf:
More background on why this is needed please.

flipflopbboi:
I agree that having this option would be very useful. Imagine you have 5000 thousand images to extract, you launch the extractor and it crashes mid-way (e.g. I use a remote machine with timeout constraints). Now the only option is to relaunch the entire process from scratch; `--skip-exisinting` should prevent double-work , BUT this does not apply to images that had no faces/did not match the filter. As such, you may very well spend hours on re-checking empty images. Hope this makes the use case clearer, thanks to both!

torzdf:
Ok, I do understand the reasoning, I'm just not sure that I like a process that deletes files from the hard disk (I struggle to think of another application which does this). I think a better option would be to expand out from skip-existing, to have two options (skip-existing-faces and skip-existing-frames).

Either way, I would hold fire on this, as I am currently improving extract which may obsolete this. People can still checkout this PR, but I would rather not add a cli option only to remove it again shortly afterwards.

torzdf:
Check if #522 absorbs these issues.

Thanks

torzdf:
This has been obsoleted by #522 

Use skip-existing to have the desired effect (skip existing now skips any frames in the alignments file, regardless of whether a face was found or not, so frames will not be re-extracted).

PR #515, opened 2018-10-01
Update PixelShuffler.py

L4Z3RC47:
Keras 2.2.2 moves normalize_data_format to keras.backend from conv_utils


torzdf:
requirements.txt also will need updating, and testing for Keras version bump everywhere it is used will need to be performed.

PR #523, opened 2018-10-22
Extraction - Speed improvements (#522)

torzdf:
This PR implements speed improvements by better leveraging parallel processing, whilst enabling easier development of alignment and detector plugins if future developments change in this area.

It also implements some additional (minor) features:
- Re-implement Convert 'on the fly'. Convert will now occur on the fly again if an alignments file is not present. *However* it will use Dlib HOG and Dlib Pose Predictor for detection and alignment, so as always it is recommended to run extract first to generate an alignments file. This is useful for quick testing though
- Re-Implement Dlib Pose Predictor alignments. Not as good as Face Alignment Network, but CPU heavy rather than GPU Heavy
- Skip Existing Faces added (cli option -sf). This will skip any frames where faces have been detected in the alignments file. Skip Existing (-s) now skips any frames that do not exist in the alignments file at all
- Size (-sz) option. Faces can now be extracted at different size (other than the default 256) for models that utilize it.

This is beta, so please checkout, test and feedback success/failures (including your python version and OS).

In theory the more VRAM you have the faster your detection should be. You should use the -mp flag for fastest results. You now need to specify the aligner (-A) as well as the detector.

Benchmarks:
1000 Images, average over 5 runs. Times are in seconds from issuing the command to completion running on a GTX 1080
```
		|           time	    |   speed increase
		|current  serial  parallel  |serial	parallel
----------------|---------------------------|------------------------
dlib-cnn	|    516       224     164  |  143%	    234%
mtcnn		|    343       197     134  |   83%	    169%
```

dlib-cnn
mtcnn


kilroythethird:
There seem to be a bug with -bt

> Extracting faces:   0%|                                                                                                                                                                                               | 0/135 [00:00<?, ?it/s]
> Traceback (most recent call last):
>   File "/media/data/home/nope/workspace/faceswap/faceswap.py", line 36, in <module>
>     ARGUMENTS.func(ARGUMENTS)
>   File "/media/data/home/nope/workspace/faceswap/lib/cli.py", line 81, in execute_script
>     process.process()
>   File "/media/data/home/nope/workspace/faceswap/scripts/extract.py", line 48, in process
>     self.run_extraction(save_thread)
>   File "/media/data/home/nope/workspace/faceswap/scripts/extract.py", line 128, in run_extraction
>     self.post_process.do_actions(faces)
>   File "/media/data/home/nope/workspace/faceswap/scripts/fsmedia.py", line 274, in do_actions
>     action.process(output_item)
>   File "/media/data/home/nope/workspace/faceswap/scripts/fsmedia.py", line 310, in process
>     face.landmarksXY(),
> TypeError: 'list' object is not callable

Otherwise looks good, but using MTCNN and dlib as aligner i get ~ 0.2 it/s less than with the current version (Can use FAN because i am using an AMD card and plaidml.keras).
I'll see if i can write my own aligner supporting Keras directly, but i understand that AMD uses are not official supported by this project.

torzdf:
@kilroythethird  I'm guessing you're using CPU as MTCNN isn't built with Keras? The optimizations are purely GPU based (except, maybe dlib-hog, but I didn't benchmark that). You could probably force better cpu utilization by ammending plugins/extract/detect/mtcnn.py line 78 (`alloc = 2048`) to something sensible for your system.

I'd be interested to know if it makes a difference, as I may go in and make it more dynamic.

torzdf:
@kilroythethird -bt bug should be squashed in latest commit

PR #524, opened 2018-10-23
fix docker run command

wAuner:
the install instruction does not work with the provided commands. If the image is named "deepfakes-gpu" then this is what needs to be started with `docker run`. fixed that in steps 5 and 6.


torzdf:
I don't use docker. Can someone who does please confirm this PR.

Thanks

wAuner:
I've tested it of course, but you probably want someone else to confirm it.
I'd also change the command to prevent jupyter notebook from starting automatically when the container starts, since jupyter isn't necessary. In that case it wouldn't be necessary to forward the container ports either.

If were to be approved, I could help clean up the docker commands

PR #528, opened 2018-10-27
OriginalHighRes model plugin (rc4)

andenixa:
- Model is now configurable through `faceswap.ini`

- Removed `K.function` support as it caused issues


Teckt:
after running multiple training sessions on a 1080ti for 192x @ bs 4, bs 8, bs 16, it all shoots up to loss levels of around .4 and the training squares become all either red or white.

gessyoo:
I had the same issue when training with OriginalHighRes rc4, changing line 72 of Model.py to disable K.Function fixed it:  "USE_K_FUNCTION = False."  If that doesn't fix it for you, switch to the Plugin_OriginalHighResRC4fix branch.  No K.Function = no red or white "screen of death."

PR #563, opened 2018-12-28
Improvement to blur estimates - normalize by image size & efficiency …

kvrooman:
Improvement to blur estimates - normalize by image size & efficiency improvements

Efficiency
1. Laplacian requires grayscale image, yet code previously loaded as either gray or color and then cast to gray. Reading as grayscale off the bat avoids the conversion and if query
2. OpenCV defaults to using double precision ( 64float ) in the Lapacian calculation. This is overkill for this task and single ( 32float ) is sufficient and speedier

Blur estimate
3. As image size increases and the amount of pixels increase, the gradient between any two pixels will decrease accordingly as there are more pixels to spread the difference over. This dampening effect on the gradient has a corresponding effect on variance and  skews the blur estimate to give different results on differnet size images. Normalizing by pixel number attempts to counteract this effect


torzdf:
Good PR. I will test it out. Sort by Blur has always been problematic, so any improvements are appreciated

PR #568, opened 2018-12-30
Update setup.py

DKingCN:
1.  Update setup.py 
    * Provide temporary solution for CUDA 10.0 (tf-nightly-gpu) which will be merged back to tensorflow-gpu in mid of Jan.2019.
    * Correct chcp code from 866 to 65001.
    * Update to VS2017 guide to compile dlib (which requires important extra library)
    * Remove detection and guide for Manual CMake installtion because it is unnecessary with proper setup of VS2017
    * Remove miserable detetion of VS2017 and its components. Lets just hope user make it right. We actually need to detect `cl.exe`, `cmake.exe` but its basically undetectable and even we find them exist it does not means that Visual studio is properly setup so just forget it.

2. Remove version declaration in requirements.txt
   It conflicts with new packages and easily gets outdated
   As I tested extraction&train&conversion, the project works with latest dependencies.
   (I am working on Windows with VS2017 & CUDA 10.0 & prebuilt tf-nightly-gpu & all latest deps)

### DIFF
--- a/Dockerfile.cpu
+++ b/Dockerfile.cpu
@@ -6,9 +6,10 @@ RUN apt-get update -qq -y \
  && rm -rf /var/lib/apt/lists/*
 
 COPY requirements.txt /opt/
+RUN pip3 install --upgrade pip
 RUN pip3 install cmake
 RUN pip3 install dlib --install-option=--yes --install-option=USE_AVX_INSTRUCTIONS
 RUN pip3 --no-cache-dir install -r /opt/requirements.txt && rm /opt/requirements.txt
 
-WORKDIR "/notebooks"
-CMD ["/run_jupyter.sh", "--allow-root"]
+WORKDIR "/srv"
+CMD ["/bin/bash"]
--- a/Dockerfile.gpu
+++ b/Dockerfile.gpu
@@ -6,6 +6,7 @@ RUN apt-get update -qq -y \
  && rm -rf /var/lib/apt/lists/*
 
 COPY requirements.txt /opt/
+RUN pip3 install --upgrade pip
 RUN pip3 install cmake
 RUN pip3 install dlib --install-option=--yes --install-option=USE_AVX_INSTRUCTIONS
 RUN pip3 --no-cache-dir install -r /opt/requirements.txt && rm /opt/requirements.txt
@@ -16,5 +17,5 @@ RUN cd /usr/local/cuda/lib64 \
  && ln -s libcuda.so libcuda.so.1 \
  && ldconfig
 
-WORKDIR "/notebooks"
-CMD ["/run_jupyter.sh", "--allow-root"]
+WORKDIR "/srv"
+CMD ["/bin/bash"]
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -13,10 +13,9 @@
     - [Notes](#notes)
 - [Windows Install Guide](#windows-install-guide)
     - [Prerequisites](#prerequisites-1)
-        - [Microsoft Visual Studio 2015](#microsoft-visual-studio-2015)
+        - [Microsoft Visual Studio 2017](#microsoft-visual-studio-2017)
         - [Cuda](#cuda)
         - [cuDNN](#cudnn)
-        - [CMake](#cmake)
         - [Anaconda](#anaconda)
         - [Git](#git)
     - [Setup](#setup-1)
@@ -76,7 +75,7 @@ The developers are also not responsible for any damage you might cause to your o
 - [virtualenv](https://github.com/pypa/virtualenv) and [virtualenvwrapper](https://virtualenvwrapper.readthedocs.io) may help when you are not using docker.
 - If you are using an Nvidia graphics card You should install CUDA (https://developer.nvidia.com/cuda-zone) and CUDNN (https://developer.nvidia.com/cudnn). If you do not plan to build Tensorflow yourself, make sure you install no higher than version 9.0 of CUDA and 7.0.x of CUDNN
 - dlib is required for face recognition and is compiled as part of the setup process. You will need the following applications for your os to successfully install dlib (nb: list may be incomplete. Please raise an issue if another prerequisite is required for your OS):
-    - Windows: Visual Studio 2015, CMake v3.8.2
+    - Windows: Visual Studio 2017 with some special components (listed [here](#microsoft-visual-studio-2017))
     - Linux: build-essential, cmake
     - macOS: xquartz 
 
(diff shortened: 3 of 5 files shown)


bryanlyon:
We can't remove all version declarations in requirements.txt.  Even if it works right now (which is not actually given, though I'm glad it works for you), we include the version declarations to prevent future problems as much as anything.  Actually changing those declarations over time as new versions come out and are verified is MUCH easier than handling the troubleshooting when it doesn't.

The rest of these pulls make sense to me, but torzdf will have to make the final call on that.

torzdf:
Ok, some feedback.

Firstly, thanks for taking the time to do this PR. I think it will be useful and it is appreciated.

However.

The main reason that I did the full install instructions for Windows and put so many 'hacky' checks for correct installed versions in setup,py was because I was having to take an inordinate amount of time walking people through installing for Windows.

I still have issues now, albeit no where near as many, from people who fail to follow the instructions and then don't understand why their install doesn't work. 

It is widely reported that Dlib won't compile with VS 2017, so I will need to check that. If we do move to VS 2017 (ideally build tools, rather than full code), then we will need checks in place to make sure it is installed for Windows users. The check may not catch 100% of the installs, but if we can catch 80% then it will make my life a great deal easier.

DKingCN:
Glad you point all these things that I am not so sure about.

Lets sort them out one by one.

### 1. Compile dlib against VS2017
This is the main reason why Im pulling this request. I think I located the real problem that troubled most people so trying to give some real tips here.

VS2017 actually works to compile dlib with AVX&CUDA but it requires packages of VC++ 2017 for Spectre which I have mentioned in commit message. This is the reason that troubled most people. I had an article debugging it [here](https://blog.qaq.link/cmake-with-vs2017-error/). I hope it explained well for you.

The following content is translated by google
```
CMAKE WITH VS2017 COMPILATION FAILURE REASON AND SOLUTION
 Published on: December 31, 2018  Category: Uncategorized Edit
When compiling dlib, I found that cmake reported an error. 

The hint is that the CMAKE_C_COMPILER and CMAKE_C_COMPILER cannot be found. 
This error is very vague, because in fact the VS2017 C++ project supports direct compilation. Setting vcvars manually has no effect.

So the troubleshooting method is as follows:

Try to compile dlib manually, check CMakeError.log for error, and find out the actual reasons are: 
1. VS2017 activates Spectre mitigation option by default 
2. Spectre Mitigation library does not exist, resulting in MSVCRTD.lib not found.

So the solution is to use the VS2017 Installer to manually select 
VC++2017 version 15.X v14.x Libs for Spectre (x86 and X64)

Try to recompile and successfully resolve.
```

### 2. Why Im trying to remove detection of VS
I suppose people to have many hacky style of building environment. Not only Visual Studio 201X but also CMake and MinGW together or whatever else. They all actaully works.\
But you are right and I might have gone too far. People seeking to use the `setup.py` seems to be less flexible on this. They might not expect this much information but simple detection and guides.\
I will start to improve the detection right now.

Tho cmake path is actually unnecessary so I removed them. Will give tips elsewhere.

### Dep version
you are right.
The update is too aggresive to apply. Forget it.

DKingCN:
### About CMake installation
I did some more tests.

The conclusion is:
1. In theory, only one installation of cmake is needed.
2. Yet compile with cmake from VS2017 will fail.
3. Yet the fail of 2 is not the fault of VS2017, its a bug of dlib CMakeList.txt.
4. dlib made an inappropriate cmake version check, which failed cmake from VS2017 who actually should pass the build.

The fast solution is to install another cmake as a python dependency from requirements.txt, So python will use the ideal version of cmake.

The right thing to do is to fix dlib CMakeList.txt, so we will only need a single cmake installation from VS2017.

DKingCN:
dont forget this bro

kvrooman:
thanks for this work as well. We'll try to get here to test in the near future as the refactor is integrated and bug fixes are fixed.

-Two notes:
TF 1.13 is supposed to come in Jan 2019, so should we wait for that stable build to be released

The following are listed as exact versions. Do we need these exact versioning? Can we support >= to give users more flexibility...
cmake>=3.13
dlib==19.16.0
pathlib==1.0.1
ffmpy==0.2.2
scandir==1.7
h5py==2.9.0
matplotlib==3.0.2

torzdf:
I'd happily remove fixed versions, only implementing them if new versions bug our code.

@DKingCN Sorry that it's taking so long to get around this. The refactor has taken longer than expected. It will be looked at!

DKingCN:
easy. maybe I need to remove the temp solution for cuda 10.0 once TF is officially released with cuda 10.0 support. TF 1.12 release note is not mentioning this so i guess we need to wait some more time.

torzdf:
Not really happy updating requirements/setup until it enters full release

torzdf:
This is still on the agenda.

@DKingCN I don't suppose you'd want to write a docker install/useage guide for the wiki? @kvrooman has started adding wiki pages, and a docker guide is likely to be useful.

Thanks

DKingCN:
The guide is pretty much there in setup.py
The needed command line to run the docker container is listed there and the remaining things are all docker stuff. nothing special.

The tf version tho, i need to check the current status of it in pypi and decide whether to update the script.

torzdf:
Ok, I've finally had a chance to look at this... I can build dlib with Visual Studio Build Tools 2017, but only if I download dlib from the github repo and run setup.py. If I try to install from pypi I get the following error:
```"distutils.errors.DistutilsError: Could not find suitable distribution for Requirement.parse('PasteScript==dev,>=1.6.3dev-r7326')"``` 
and I can't solve this with any amount of googling :/

Secondly, we can entirely scrap the CUDA and cuDNN steps by installing tensorflow from conda (```conda install tensorflow-gpu```), this automatically handles the CUDA and cuDNN requirements inside the environment and will make install a LOT easier. This does mean that tensorflow will need to be installed prior to dlib installation so that CUDA can be picked up for dlib.

With this in mind I propose the following:
- updating INSTALL.md for Windows to remove cmake, cuda and cudnn steps. Change VS2015 step to VS Build Tools 2017
- Work on checks for correct version of build tools/vs >= 2015 <= 2017
- Install Tensorflow from Conda for Conda environments
- clone dlib from 'https://github.com/davisking/dlib' at depth=1 or download the zip and unzip it.
- run setup.py from the cloned dlib repo with detected optional argument `--yes USE_AVX_INSTRUCTIONS` detected argument `-G Visual Studio 14/15 2015/2017` `--clean`
- delete cloned repo

I'm happy to do this, but thought you may want to, or maybe make a start? Let me know how you want to proceed.

DKingCN:
I never tried conda, and still not feeling like to do it. Will prefer leave this to you. Thanks.

torzdf:
@DKingCN Thanks again for this. As you may have seen, I decided to go a different way and create an Installer for Windows, packaging in Dlib.... At some point in the future I will go back and review the Windows Code for compiling, but unfortunately my proposed solution up there didn't work.

Would it be possible for you to raise a new PR to amend the new code to fix the recent issues with Docker? You don't need to do anything with Windows, I will look at that at some point in the future, taking your notes on board.

I really appreciate any assistance you can give.

DKingCN:
Fix about docker #625 

Now this old PR only exists because of the trouble of dlib installation. Maybe you can close this when you managed to find a way to get around with dlib.

torzdf:
Thank you. Hugely appreciated. I will close this one off and re-address the Windows/VS2017 issue in future.

deepfakes:
Closing this off as I think setup.py and INSTALL.md are in a relatively good place at the moment.

Thanks again for all your help @DKingCN

PR #574, opened 2019-01-03
Consolidation of converters & refactor

kvrooman:
Changes
- consolidate convert_masked & convert_adjust into one converter with all enhancement options
- allow mask transition blur size to be a fixed integer of pixels and a fraction of the facial mask size
- allow erosion/dilation size to be a fixed integer of pixels and a fraction of the facial mask size
- allow users to adjust the coverage factor of the model's output faces 
- dynamic interpolation algorithm for  upsample / downsample for each image based on scale
- eliminate redundant type conversions to avoid  round-off errors
- addition of 3 more masks type
      - smoothed: Dfaker style mask which uses rect as a base and then smooths the mask transition
      - ellipse: fit the closest enclosing ellipse around all the facial landmark points
      - cnn: currently not working but where the facial segmentation model will be incorporated
- reorganize for clarity & style & speed where available

TODO
- histogram currently non-functional.. working on mask / image interactions


torzdf:
Thanks for this. Going to change base to train_refactor so you can remove GAN. 

If you'd rather it went straight to Staging > Master, then feel free to change it back,

PR #586, opened 2019-01-11
attempted fix of a crash in apply_new_face function

d8888:
1. apply_new_face will crash during seamless clone of openCV if src image cannot be fitted into dest image
2. point structure of openCV should be (maskx, masky) rather than (masky, maskx)

### DIFF
--- a/plugins/convert/Convert_Masked.py
+++ b/plugins/convert/Convert_Masked.py
@@ -117,10 +117,21 @@ class Convert():
                 leny = maxy - miny
                 masky = int(minx + (lenx // 2))
                 maskx = int(miny + (leny // 2))
+                
+                crop, xmin, xmax, ymin, ymax, cx, cy = self.cal_crop(new_image, 
+                                                                     base_image, 
+                                                                     maskx, 
+                                                                     masky)
+                if crop:
+                    new_image = new_image[ymin:ymax+1, xmin:xmax+1]
+                    unitMask = unitMask[ymin:ymax+1, xmin:xmax+1]
+                    maskx = maskx - cx
+                    masky = masky - cy
+                
                 outimage = cv2.seamlessClone(new_image.astype(numpy.uint8),
                                              base_image.astype(numpy.uint8),
                                              unitMask,
-                                             (masky, maskx),
+                                             (maskx, masky),
                                              cv2.NORMAL_CLONE)
                 return outimage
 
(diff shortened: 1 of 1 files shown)


d8888:
Sorry it appears git logs are truncated 

I am getting the following error when using seamlessclone (-S) option When I convert images using trained model: 

OpenCV Error: Assertion failed (0 <= roi.x && 0 <= roi.width && roi.x + 
roi.width <= m.cols && 0 <= roi.y && 0 <= roi.height && roi.y + roi.height 
<= m.rows) in cv::Mat::Mat

The exception is thrown by the apply_new_face function in Convert_Masked.py if new_image cannot "fit" into base_image at the coordinates of (maskx, masky). The PR determine if new_image will exceed the boundary of base_image and crops new_image if necessary.

Also, the point in openCV should follow the format of (X, Y) instead of (Y,X) so I modified the call of cv2.seamlessClone.

Thanks

torzdf:
@kvrooman Does this look ok to you? We can merge it into staging/master with the understanding that any of these changes will be overwritten by your convert refactor

kvrooman:
The crop adjusting for the out of bounds error is a good addition. I did something similar by just padding the source image which we are pasting into to ensure there are no cases for out of bounds.

Regarding the y,x versus the x,y order of the center parameter in the function call... this is a case of poor code labeling.  Note
```lenx = maxx - minx
                leny = maxy - miny
                masky = int(minx + (lenx // 2))
                maskx = int(miny + (leny // 2))
                outimage = cv2.seamlessClone(new_image.astype(numpy.uint8),
                                             base_image.astype(numpy.uint8),
                                             unitMask,
                                             (masky, maskx),
                                             cv2.NORMAL_CLONE)```

The original code

d8888:
@kvrooman: Thanks for your kind reply. It seems maskx, masky are just local variables, I changed their names so that maskx, masky corresponds to X,Y. Is this feasible? Thanks~

kvrooman:
It does work now and I believe your addressing of the out_of_bounds issues are a good item ( although I haven't tested extensively )

Unfortunately, we're in the midst of a code refactor with the training code being updated. As part of that, I had consolidated the Convert_Masked and Convert_Adjust converters into one converter ( PR #574 ) over the last week or so. So the Convert_Masked will hopefully shortly be overwritten in order to provide more features and have proper face alignment.

If you'd like to review #574 and think something should be modified there to address these similar issues, lemme know

torzdf:
In the meantime, I will merge this into staging, as it does address an existing bug.

PR #587, opened 2019-01-14
Code to use ICNR

kvrooman:
Class I've been using to incorporate ICNR into my models

Example Usage
```
 x = Conv2D(filters=filters*4, kernel_size=3, strides=1, padding='valid', activation = self.l_relu, data_format = 'channels_first', kernel_initializer=ICNR_init(initializer=he_normal()), name = '{}_conv'.format(layer_name))(x)


torzdf:
The code for the nn_block should probably come out, as the initializer can be passed in as an argument from the model, so it should be selected from the model and passed through to the nn_block.

Once this change is made, I'll merge.

torzdf:
ICNR Has been implemented, so closing PR.

PR #588, opened 2019-01-14
default initializer = He instead of Glorot

kvrooman:
https://www.groundai.com/project/how-to-start-training-the-effect-of-initialization-and-architecture/

More stable weights / gradients with different initializer when using RELU


torzdf:
I have standardised the nn_blocks, so please could you fix for the conflicts.
`kernel_initializer` can also be passed through as a parameter from the models (e.g. OriginalHiRes), so please could you make it overridable?

Once this is done I'll merge this

kvrooman:
Conflict avoided

kvrooman:
Question on the OriginalHiRes - do we want to standardize to He initializers there as well...

PR #600, opened 2019-02-03
Effmpeg split quality

Lorjuo:
effmpeg currently defaults to bad quality when extracting / splitting jpg pictures.
This commit will extract jpg with maximal quality


kvrooman:
Note quality 1 takes much longer than other more moderate quality levels
for less and less incremental quality gain.  Fine if this is for offline
work, but we may want a preset for "real-time" in the future

PR #601, opened 2019-02-06
Fixed GPU memory allocation issues.

HoltSpalding:
This edit fixes the GPU memory allocation issues. The solution is described here https://www.tensorflow.org/guide/using_gpu. The same thing is done when running train.py with the "--allow-growth" flag. This solution has only been tested with a single GPU. 
RTX 2060, Cuda 10.0, 4.15 driver, cudnn 7.4, tensorflow 1.13rc

### DIFF
--- a/plugins/model/Model_Original/Model.py
+++ b/plugins/model/Model_Original/Model.py
@@ -1,4 +1,11 @@
 # Based on the original https://www.reddit.com/r/deepfakes/ code sample + contribs
+#For GPU memory allocation issues
+import tensorflow as tf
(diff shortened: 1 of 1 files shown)


HoltSpalding:
There is no --ag cli option for training as far as I can tell, you should add that in the next version.

torzdf:
On training? There is.

bryanlyon:
It's only one dash.  -ag (or --allow-growth).

gessyoo:
@HoltSpalding, python faceswap.py train -h

PR #605, opened 2019-02-13
Fix -dt by starting with a 4 channel mask already in place.

bryanlyon:
Existing -dt crashes out without success.  This PR puts the 4 channel alpha in before patching.  To my knowledge, none of the patches are 4 channel aware so we can fake a flat alpha for their purposes.

### DIFF
--- a/plugins/convert/masked.py
+++ b/plugins/convert/masked.py
@@ -43,8 +43,14 @@ class Convert():
         detected_face.load_aligned(image, size=self.training_size, align_eyes=False)
         new_image = self.get_new_image(image, detected_face, coverage, image_size)
         image_mask = self.get_image_mask(detected_face, image_size)
+
         if self.args.draw_transparent:
-            new_image = dfl_full(detected_face.landmarks_as_xy, new_image, channels=4)
+            new_image = dfl_full(detected_face.landmarks_as_xy, new_image, channels=4 #Add mask as 4th channel for saving as alpha on supported output formats
+
+            #This make sure that all the arrays match in size for later actions despite not actually using alpha in any way.
+            image_mask = cv2.cvtColor(image_mask, cv2.COLOR_RGB2RGBA)
+            frame = cv2.cvtColor(frame, cv2.COLOR_RGB2RGBA)
+
         patched_face = self.apply_fixes(image,
                                         new_image,
                                         image_mask,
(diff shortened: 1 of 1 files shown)


plugins/convert/masked.py
```diff
@@ -43,6 +43,8 @@ def patch_image(self, image, detected_face):
         detected_face.load_aligned(image, size=self.training_size, align_eyes=False)
         new_image = self.get_new_image(image, detected_face, coverage, image_size)
         image_mask = self.get_image_mask(detected_face, image_size)
+        if self.args.draw_transparent:
+            new_image = dfl_full(detected_face.landmarks_as_xy, new_image, channels=4)
```
kvrooman: The next logic line in the code is in apply fixes and is a boolean
 ```if self.args.draw_transparent:```
Might as well put line 47 at line 157

plugins/convert/masked.py
```diff
@@ -149,12 +151,12 @@ def set_blur_size(self, mask):
 
     def apply_fixes(self, frame, new_image, image_mask, image_size):
         """ Apply fixes """
+        masked = new_image  # * image_mask
```
kvrooman: old code that was removed in previous commits. you don't need to reinsert

plugins/convert/masked.py
```diff
@@ -149,12 +151,12 @@ def set_blur_size(self, mask):
...
 
         if self.args.draw_transparent:
             alpha = np.full((image_size[1], image_size[0], 1), 255.0, dtype='float32')
```
kvrooman: Delete this line ( 154 ) too as it will be unused

plugins/convert/masked.py
bryanlyon: This is necessary here since alignments aren't sent to apply_fixes().  Moved those lines up instead.

plugins/convert/masked.py
```diff
@@ -43,6 +43,14 @@ def patch_image(self, image, detected_face):
         detected_face.load_aligned(image, size=self.training_size, align_eyes=False)
         new_image = self.get_new_image(image, detected_face, coverage, image_size)
         image_mask = self.get_image_mask(detected_face, image_size)
+
+        if self.args.draw_transparent:
+            new_image = dfl_full(detected_face.landmarks_as_xy, new_image, channels=4 #Add mask as 4th channel for saving as alpha on supported output formats
+
+            #This make sure that all the arrays match in size for later actions despite not actually using alpha in any way.
+            image_mask = cv2.cvtColor(image_mask, cv2.COLOR_RGB2RGBA)
+            frame = cv2.cvtColor(frame, cv2.COLOR_RGB2RGBA)
```
kvrooman: ```frame``` should be ```image```

PR #606, opened 2019-02-13
Multi-pass blur

kvrooman:
Approximate a Gaussian blur with multiple passes of a box filter.
Reduce filter size to keep blurred area approximately the same.
Experiment with pass number [3 or 4]

Other minor editing cleanup


kvrooman:
Added in code attribution to master

PR #612, opened 2019-02-18
import numpy for lines 544 and 545

cclauss:
[flake8](http://flake8.pycqa.org) testing of https://github.com/deepfakes/faceswap on Python 3.7.1

$ __flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics__
```
./lib/model/losses.py:544:15: F821 undefined name 'numpy'
    kernels = numpy.transpose(numpy.asarray(kernels), (1, 2, 0))
              ^
./lib/model/losses.py:544:31: F821 undefined name 'numpy'
    kernels = numpy.transpose(numpy.asarray(kernels), (1, 2, 0))
                              ^
./lib/model/losses.py:545:15: F821 undefined name 'numpy'
    kernels = numpy.expand_dims(kernels, -2) / numpy.sum(numpy.abs(kernels))
              ^
./lib/model/losses.py:545:48: F821 undefined name 'numpy'
    kernels = numpy.expand_dims(kernels, -2) / numpy.sum(numpy.abs(kernels))
                                               ^
./lib/model/losses.py:545:58: F821 undefined name 'numpy'
    kernels = numpy.expand_dims(kernels, -2) / numpy.sum(numpy.abs(kernels))
                                                         ^
5     F821 undefined name 'numpy'
5
```
__E901,E999,F821,F822,F823__ are the "_showstopper_" [flake8](http://flake8.pycqa.org) issues that can halt the runtime with a SyntaxError, NameError, etc. These 5 are different from most other flake8 issues which are merely "style violations" -- useful for readability but they do not effect runtime safety.
* F821: undefined name `name`
* F822: undefined name `name` in `__all__`
* F823: local variable name referenced before assignment
* E901: SyntaxError or IndentationError
* E999: SyntaxError -- failed to compile a file into an Abstract Syntax Tree

PR #613, opened 2019-02-18
Automate the discovery of undefined names

cclauss:
Automate the discovery of issues such as #88 and #612

### DIFF
--- a/.github/main.workflow
+++ b/.github/main.workflow
@@ -1,0 +1,8 @@
+workflow "on push" {
+  on = "push"
+  resolves = ["Python Syntax Checker"]
+}
+
+action "Python Syntax Checker" {
+  uses = "cclauss/Find-Python-syntax-errors-action@master"
+}


sofiahanks:
Netgear delivers the best assortment of choices to meet every type of business and home networking need. In homes more, people use Smart devices, laptops or computers for which they required WiFi routers to connect those devices to the internet. The Netgear router provides strong WiFi signals, reliability and performance. For any query, you can dial [Netgear router support number](http://mywifiext-helpnumber.com/) or[ Netgear router support](http://routernetlogin.com/).

torzdf:
Thanks for this PR. I need to review it, which I will do as soon as I can.

torzdf:
Thanks for the PR. I have had a discussion with some of the other devs, and whilst we all agree that we should, at some point, implement a framework to test commits, it is something we'll need to review in more depth before implementing anything solid.

I'm closing this off for now, but may well revisit it at some point in the future.

PR #663, opened 2019-03-14
modified file to allow converting multiple faces in a frame with mult…

ameenba:
Not functioning yet - but wanted some help with an error

Looks like when I wrote this 'patch_face' was written as 'convert_one_face', and I'm running into some issue - I think it has to do with loading multiple models. Anyway, another pair of eyes on this would be fantastic.

Logs are attached, modified file structure looks like: '_path_to_models/[model1 folder, model2 folder, etc]', each model should also contain a .png with the targets face, running it is the same except you specify the folder (i.e. python faceswap.py convert -i src/ -o out_test/ -m models/) (there's no checking that faces/models are different, so for testing feel free to just copy/paste something you already have)

[crash_report.2019.03.14.175431537621.log](https://github.com/deepfakes/faceswap/files/2968615/crash_report.2019.03.14.175431537621.log)

[faceswap.log](https://github.com/deepfakes/faceswap/files/2968617/faceswap.log)

### DIFF
--- a/scripts/convert.py
+++ b/scripts/convert.py
@@ -19,6 +19,11 @@ from plugins.plugin_loader import PluginLoader
 
 from .extract import Plugins as Extractor
 
+import numpy as np
+import face_recognition
+from PIL import Image
+import glob
+
 logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
 
@@ -40,6 +45,37 @@ class Convert():
         self.post_process = PostProcess(arguments)
         self.verify_output = False
 
+        ##initialize last_used face to whichever face is first
+        self.matched_face = 0
+        ##load images for target face id
+        try:
+            models_dir = self.args.model_dir
+            target_images = [face_recognition.load_image_file(file) for file in glob.glob(models_dir + '**\\*.png', recursive=True)]
+        except:
+            print("no images found")
+            quit()
+
+        ##create target face encodings which will match closest
+        try:
+            self.target_encodings = [face_recognition.face_encodings(image)[0] for image in target_images]
+        except:
+            print("I wasn't able to locate any faces in at least one of the images. Check the image files. Aborting...")
+            quit()
+
+        ##load trained models, should have one model per image
+        try:
+            models = [self.load_model(get_folder(folder)) for folder in glob.glob(models_dir + '*')]
+        except:
+            print("error finding trained models")
+            quit()
+
+        ##create converters for each model
+        try:
+            self.converters = [self.load_converter(model) for model in models]
+        except:
+            print("error loading model converters")
+            quit()
+
         self.opts = OptionalActions(self.args, self.images.input_images, self.alignments)
         logger.debug("Initialized %s", self.__class__.__name__)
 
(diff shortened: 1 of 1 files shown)


scripts/convert.py
```diff
@@ -19,6 +19,11 @@
 
 from .extract import Plugins as Extractor
 
+import numpy as np
+import face_recognition
+from PIL import Image
```
kvrooman: PIL and glob are not current requirements but the tasks they perform are replicated elsewhere in the FS toolchain. consider consolidating to one choice

scripts/convert.py
```diff
@@ -40,6 +45,37 @@ def __init__(self, arguments):
         self.post_process = PostProcess(arguments)
         self.verify_output = False
 
+        ##initialize last_used face to whichever face is first
+        self.matched_face = 0
+        ##load images for target face id
+        try:
+            models_dir = self.args.model_dir
+            target_images = [face_recognition.load_image_file(file) for file in glob.glob(models_dir + '**\\*.png', recursive=True)]
+        except:
```
kvrooman: Consider using cv2.imread and os.walk  with file type constaints
or  
```from scripts.fsmedia import Images``` class to load images

File: scripts/convert.py [hunk 3]
```diff
@@ -40,6 +45,37 @@ def __init__(self, arguments):
...
+
+        ##create target face encodings which will match closest
+        try:
+            self.target_encodings = [face_recognition.face_encodings(image)[0] for image in target_images]
+        except:
+            print("I wasn't able to locate any faces in at least one of the images. Check the image files. Aborting...")
+            quit()
+
+        ##load trained models, should have one model per image
+        try:
```
kvrooman: one model per image ?! a 3.5 minute video clip will have 5,000 models ?!

scripts/convert.py
```diff
@@ -171,22 +212,35 @@ def check_alignments(self, frame):
...
+            ##if not skip:
+            for face in faces:
+                ## pick best model and swap
+                im = Image.fromarray(face.image)
+                try:
+                    unknown_face_encoding = face_recognition.face_encodings(np.array(im))[0]
+                    results = face_recognition.face_distance(self.target_encodings, unknown_face_encoding)
+                    self.matched_face = np.argmin(results)
+                    print(self.matched_face)     
+                except:
```
kvrooman: a face encoding for every image and a model for every image seems like overkill. Face_recognition has a k_means clustering technique to consolidate multiple encodings of the same person / identity into a master encoding.  You could argmin against this master and then would only need one model per identity then?

scripts/convert.py
slahvalyn: no no no sorry, one model per face encoding image. each model is trained the same way as before with however many pictures you like of source and target - this just associates each model with a face so we know which model to use once we identify a face.

scripts/convert.py
slahvalyn: could potentially cluster all of the images in the train set to create an encoding, should be an insignificant amount of time compared to training the nn - although as long as faces are different enough selecting which face you've detected seems to pick the right one pretty accurately.

torzdf:
I haven't had a chance to look at this code yet, I will do. However, Convert is getting refactored very shortly, so I will look then

scripts/convert.py
slahvalyn: removed imports not in requirements

scripts/convert.py
slahvalyn: using os.listdir and Images class - good tip, thanks

torzdf:
I'm closing this off, as realistically the code has moved on a bit too much.

I'm also not entirely sold on the benefits.

PR #665, opened 2019-03-15
replaced PenalizedLoss function with keras alternative

kilroythethird:
This enables plaidml backends to also use penalized mask loss.
Imo the comments around the function could be removed as, except variable naming, nothing else stayed the same.
I switched from a class to a closure because it seemed easier for that trivial piece of code.

*Ops, sorry i planed to send this PR to staging, my bad. Shouldn't matter tho i think.*


andenixa:
Sounds solid

torzdf:
Sorry guys. There are PRs stacking up and I'm behind. Will get to these asap

PR #673, opened 2019-03-19
Mask optimizations

kvrooman:
- Remove repetitive np.array() of items that are already np.arrays
- Remove repetitive reshaping
- Perform dilation on single channel only
- Clarify dfl_full mask parts
- Simplify with for loop
- Add more complex mask based on facial components
- Prep for convert masks

### DIFF
--- a/plugins/train/_config.py
+++ b/plugins/train/_config.py
@@ -8,7 +8,11 @@ from lib.config import FaceswapConfig
 logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
 MASK_TYPES = ["none", "dfaker", "dfl_full", "components"]
-MASK_INFO = "The mask to be used for training. Select none to not use a mask"
+MASK_INFO = ("The mask to be used for training:"
+             "\n\tnone: Doesn't use any mask."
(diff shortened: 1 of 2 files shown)


plugins/train/_config.py
```diff
@@ -7,7 +7,7 @@
 
 logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
-MASK_TYPES = ["none", "dfaker", "dfl_full"]
+MASK_TYPES = ["none", "dfaker", "dfl_full", "components"]
 MASK_INFO = "The mask to be used for training. Select none to not use a mask"
```
torzdf: Please edit info as follows (adding a description for what components does)
```
MASK_INFO = ("The mask to be used for training:"
             "\n\tnone: Don't use a mask."
             "\n\tdfaker: A basic face hull mask."
             "\n\tdfl: An improved face hull mask"
             "\n\tcomponents: <INSERT DEFINITION HERE>")
```

kvrooman: Ahh, started pylinting now (10.0/10.0) but haven't flake8 yet...

PR #675, opened 2019-03-20
fix: specific TensorFlow version to the latest 1.x(1.13.1) since it is incompatible with TensorFlow 2.x API

qiwihui:
Docker tensorflow/tensorflow:latest-gpu-py3 and tensorflow/tensorflow:latest-py3 has linked to TensorFLow 2.0.0a0 which leads to faceswap run into crash.

### DIFF
--- a/Dockerfile.cpu
+++ b/Dockerfile.cpu
@@ -1,4 +1,4 @@
-FROM tensorflow/tensorflow:latest-py3
+FROM tensorflow/tensorflow:1.13.1-py3
 
 RUN apt-get update -qq -y \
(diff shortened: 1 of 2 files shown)


kvrooman:
to maintain consistency, we also should pin the specific version to 1.12 as the main non-Docker code uses that version. ( conda still working on 1.13 packages ... )

torzdf:
As @kvrooman says, please pin this to 1.12 and I will merge

qiwihui:
@torzdf Has pinned to tf 1.12.0

PR #678, opened 2019-03-21
old-style resblocks

andenixa:
added use_bias parameter to res_block (defaults to True)


kvrooman:
Hey @andenixa - as discussed on discord. I'm fine with adding this toggle. Allows some flexibility if you don't want bias. Why'd you close?

torzdf:
My understanding is this can already be done with kwargs

andenixa:
@kvrooman partially because I agree with @torzdf and because it received more criticism than I would anticipate :)

PR #688, opened 2019-03-31
PenalizedLoss for plaidml the second

kilroythethird:
Should prevent "The graph couldn't be sorted in topological order” bug in TF.
Since all seems to point to this being produced by a bug in TF or cuda itself i can't really be sure what leads to it.
Tested with the CPU version of tensorflow.

### DIFF
--- a/lib/model/losses.py
+++ b/lib/model/losses.py
@@ -168,46 +168,17 @@ class DSSIMObjective():
         return patches
 
 # <<< START: from Dfaker >>> #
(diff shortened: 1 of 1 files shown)


kilroythethird:
Damn it i did it again. Please switch to staging. Its based on it but i (once again)missed to switch it in the GH interface. Sorry about that.

torzdf:
This seems to work without generating the previous errors.
Merging

PR #690, opened 2019-04-02
Added cache for nearest landmarks

kilroythethird:
This PR adds caching to increase the training data preparation speed for the "Warp to landmarks" feature.
Currently the 10 nearest landmarks are calculated for each image over and over again.

This speed up might or might not be noticeable as it depends on how long the data preparation process takes vs. how fast the GPU can finish working on a batch.
It should decrease CPU usage tho for all cases and doesn't really cost much RAM as only the nearest 10 hashes are saved.

### DIFF
--- a/lib/training_data.py
+++ b/lib/training_data.py
@@ -196,7 +196,7 @@ class TrainingDataGenerator():
             dst_points_items = list(landmarks.items())
             dst_points = list(x[1] for x in dst_points_items)
             closest = (np.mean(np.square(src_points - dst_points),
-                          axis=(1, 2))).argsort()[:10]
+                               axis=(1, 2))).argsort()[:10]
             closest_hashes = tuple(dst_points_items[i][0] for i in closest)
             self._nearest_landmarks[filename] = closest_hashes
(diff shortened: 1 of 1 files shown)


lib/training_data.py
```diff
@@ -187,12 +190,16 @@ def get_closest_match(self, filename, side, src_points):
...
-                           axis=(1, 2))).argsort()[:10]
-        closest = np.random.choice(closest)
-        dst_points = dst_points[closest]
+        landmarks = self.landmarks["a"] if side == "b" else self.landmarks["b"]
+        closest_hashes = self._nearest_landmarks.get(filename)
+        if not closest_hashes:
+            dst_points_items = list(landmarks.items())
+            dst_points = list(x[1] for x in dst_points_items)
+            closest = (np.mean(np.square(src_points - dst_points),
+                          axis=(1, 2))).argsort()[:10]
```
torzdf: Please make the following change:
```
lib/training_data.py:199:0: C0330: Wrong continued indentation (add 5 spaces).
                          axis=(1, 2))).argsort()[:10]
                          ^    | (bad-continuation)
```

PR #698, opened 2019-04-07
Setup works for python 3.7

wiso:
Now the check is just for python >= 3.3, any new version will work

I haven't checked if there is anything that prevent python 3.7 to work

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -138,9 +138,9 @@ class Environment():
         self.output.info("Installed Python: {0} {1}".format(self.py_version[0],
                                                             self.py_version[1]))
         if not (self.py_version[0].split(".")[0] == "3"
-                and self.py_version[0].split(".")[1] in ("3", "4", "5", "6")
+                and int(self.py_version[0].split('.')[1]) >= 3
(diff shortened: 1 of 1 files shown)


kvrooman:
You mentioned that you havent checked if Python 3.7 actually works...? Perhaps, you should do some testing before making any changes that could break people's installs?

wiso:
is there any specific unit test I can run?

I just tried train.py with default arguments and it works.

By the way: now the Docker image is broken, since it contains python 3.7

torzdf:
Keras does not 'officially' support Python 3.7. Whilst it does work, there is no guarantee that it will in the future until it is updates. For that reason, we pin faceswap at 3.6

wiso:
I see (strange...)

The point is that if you build the Docker image you get python 3.7 inside and so, with the present version of faceswap it doesn't work. So, if you don't want to fix this you have to fix the Dockerfile.

torzdf:
Unfortunately I don't use docker, but would welcome a PR that addresses this,

PR #701, opened 2019-04-08
Pegasus model plug-in rc1

andenixa:
added new `RealFace` training model (codenamed: Pegasus)
Input size: `64-128`
Output image size: `64x64` to `256x256`

### DIFF
--- a/plugins/train/_config.py
+++ b/plugins/train/_config.py
@@ -188,45 +188,48 @@ class Config(FaceswapConfig):
         self.add_item(
             section=section, title="coverage", datatype=float, default=62.5, rounding=1,
             min_max=(62.5, 100.0), info=COVERAGE_INFO)
-        
+
         # << PEGASUS MODEL OPTIONS >> #
         section = "model.realface"
         self.add_section(title=section,
-                         info="an extra detailed variant of Original utilizing the majority of suggestions by Bryanlyon,\n"
+                         info="an extra detailed variant of Original utilizing the majority of "
+                              "suggestions by Bryanlyon,\n"
                               "and is greatly inspired by VillainGuy's model.\n"
                               "This model doesn't provide a low_mem variant!\n"
-                              "Requires about 6GB-8GB of VRAM (batchsize 8-16)." + ADDITIONAL_INFO )
+                              "Requires about 6GB-8GB of VRAM (batchsize 8-16)." + ADDITIONAL_INFO)
         self.add_item(
             section=section, title="lowmem", datatype=bool, default=False,
-            info="The option is here for compatibility reasons. Has no real effect.")                   
+            info="The option is here for compatibility reasons. Has no real effect.")
         self.add_item(
             section=section, title="mask_type", datatype=str, default="dfl_full",
             choices=["none", "dfaker", "dfl_full"],
             info="The mask to be used for training. Select none to not use a mask")
         self.add_item(
             section=section, title="coverage", datatype=float, default=62.5, rounding=1,
-            info="The model is essentially created for 60-80% coverage as it follows Original paradigm.\n"
-                  "You may try higher values but good results are not guaranteed.",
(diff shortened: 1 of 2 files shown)


clouders1111:
I'm just interested to know how this training model compares to the others? how is it different?

andenixa:
> I'm just interested to know how this training model compares to the others? how is it different?
There is a link of a swap made with it at 128x128 resolution.
[RealFace "Pegasus" at 128x128](https://www.youtube.com/watch?v=PJwdtPy1c64&t=3s)
I'd say it somewhere in between `Unbalanced` and `Villian` detail's wise meaning it's pretty detailed.
It would be betteer if you'd just try it. It also supports variable resolutions from 128x to 256x (in steps of 64)

Kirin-kun:
> 
> 
> > I'm just interested to know how this training model compares to the others? how is it different?
> 
> There is a link of a swap made with it at 128x128 resolution.
> [RealFace "Pegasus" at 128x128](https://www.youtube.com/watch?v=PJwdtPy1c64&t=3s)
> I'd say it somewhere in between `Unbalanced` and `Villain` details-wise meaning it's pretty detailed.
> It would be better if you'd just try it. It also supports variable resolutions from `128x128` to `256x256` (in steps of 64)

What GPU did you use to get that result? What were the final losses? Either with unbalanced or villain, I never get that level of detail on the teeth and eyes.

Kirin-kun:
And villain has a nasty habit of changing the colors on the converted faces.

Kirin-kun:
Tried it with a GTX 1060 6Gb with "Memory Saving Gradient" and "Allow Growth", it doesn't even start. It crashes with OOM even with a batch size of 2. Either it doesn't support MSG, or it's really, really, memory hungry.

It actually started with ping pong and a batch size of 8, but I think the training will take a looong time.

gessyoo:
@Kirin-kun, the model is very memory hungry.  This was made with Pegasus rc1, https://youtu.be/C0nr0429Q2k, but I obviously don't have AnDenixa's level of skill.

andenixa:
@Kirin-kun Depending on the resolution I might require more memory. I suppose 64->128 should start with 6G which is the resolution I made my swap with. It has a slightly high `dense` size which could perhaps be lowered to 1024 in `config.ini`

Kirin-kun:
> @Kirin-kun Depending on the resolution I might require more memory. I suppose 64->128 should start with 6G which is the resolution I made my swap with. It has a slightly high `dense` size which could perhaps be lowered to 1024 in `config.ini`

With the defaults, this model requires more memory than villain and I couldn't start it with 6Gb. Only with ping-pong did it start, The model files are obviously bigger too.

Villain starts with a bs of 8 with 6Gb, no problem. 

The result is good, but as you said, it's less detailed than villain. It's difficult to judge, because the amount of iterations between the two and the losses, are not the same, so I can't definitely say that one is better than the other.

andenixa:
@Kirin-kun I confess it came out a little more power hungry than I anticipated. I was aiming at preserving fine details versus high memory optimizations. Saying that it's still much less demanding than `DFL` SAE and whatnot. If `villain` works for you that is probably what you should be using. 
I  plan on releasing `RealFace`'s close cousin FaceLight (codenamed: `Buttercup`) which has much tighter memory optimizations with arguably comparable level of details.

Enyakk:
Can the model still be commented on? I set up the model like this:
Input 96, Output 160, everything else default.

I noticed this results in Decoder B.h5 being 767MB, while encoder being 115MB. In other trainer models the encoder is typically much more complex than the decoder. Can you comment on why Realface is reversed in this fashion?

andenixa:
@Enyakk it's like you said `reversed` in a way. The above size are to be expected and it doesn't mean the model is lacking but simply of a different layout.
PS: I am not sure there is such thing as `typical` when i comes to `ae` we having models like `villian` which are anything but that.

PR #706, opened 2019-04-13
Force CPU training on Tensorflow-GPU

kvrooman:
Additional flexibility for Tensorflow-GPU by allowing a " force training on CPU option"

Simply set GPUs = 0 in training cli/GUI before training start or re-load. Can dynamically switch back and forth between GPU and CPU based on the flag in different training sessions.

Benefits
- easier debugging for CPU users
- Absolute last option to train very big models ( assuming RAM is bigger than VRAM )


torzdf:
I think my preference for this would be to make cpu an option for all workflows (extract, train, convert) and to set `os.environ["CUDA_VISIBLE_DEVICES`] =""`. I am happy to implement this if you don't want to,

kvrooman:
we can do both. hide the CUDA device in the environment and set the session parameters...ill have to call the same configure_session function in extract and convert 

there are some advantages to fine-tuning the CPU process thread variables to the Intel suggested optimized values so it would be good to keep the session config

kvrooman:
@torzdf Where were we with this CPU enabler? I recall some dicussion on preferences for where we configure TF sessions? Anyways, I've resolved the conflicts and ready to re-test this again

kvrooman:
left out to long and session architectures changes setup

torzdf:
I still want to implement this, or a version of.

kvrooman:
yes, me too. Easiest to incorporate into the session.py structure using config_proto now. The code moved on so much it made sense to shut this down. I'll swing back around to using the current struture in the to-do list

torzdf:
Yep, Currently only in Extract, but the class needs expanding and shipping across the code imho.

Things to bear in mind (for my reference as much as anything), There can only ever be one TF Session per process, so it is not possible to set up a config_proto for an aligner + detector in the same session. However, we should be able to leverage with K.tf.device() for this kind of use case

PR #732, opened 2019-05-18
Fix for Effmpeg not working when using Faceswap with Docker

timothydelter:
1. Added the installation of ffmpeg for the Docker container in the Dockerfile
2. Made a small correction on the Docker part in INSTALL.md where tools.py is incorrectly used instead of faceswap.py in launching the GUI.


torzdf:
Thanks for this

Issue #739, opened 2019-05-30
Alignements.json not generated with -si option


JPFrancoia:
Hi,

I'm trying to extract faces from a video. I use this command:

```
python faceswap.py extract -si 20 -i src/input.mp4 -o faces/test
```

I tried it locally, with tensorflow installed without GPU support. The script works fine, does 2 passes, but export the alignements.json only at the very end of the second pass.

I also tried it on a AWS g2.large instance, with tensorflow installed with GPU support (tensorflow-gpu == 1.12.0). The script also works fine, GPU is used, but once again the alignements.json file is only generated at the end.

It's slightly annoying since I want to run this script on a spot instance, so the total lifetime of the instance shouldn't exceed 6h. I would like to be able to stop my instance, and restart it another time. I therefore need the alignements.json file.

Is this result expected?

torzdf:
Thanks for the PR. Merged to staging.

PR #740, opened 2019-05-30
Bug fix: export alignement.json file with save-interval option.

JPFrancoia:
One-liner fix. The modulo operation has priority over the addition. Because parentheses were missing in the condition, the condition was never met. Because of this, even if we run the extract script with the `save-interval` option, the alignements.json file is never generated.

Fixes https://github.com/deepfakes/faceswap/issues/739


torzdf:
Thanks for the fix

PR #749, opened 2019-06-03
small fix for docker cmd. tools.py->faceswap.py

wallopthecat:


wallopthecat:
Should all PRs be made against staging? Will do that in the future

torzdf:
Yeah, we tend to put it there first, just to prevent anything catastrophic occurring in Master ;)

PR #754, opened 2019-06-09
Ubuntu 19.04

martin-pr:
Attempting to make faceswap run on Ubuntu 19.04:

- Python 3.7 explicitly allowed (seems to work)
- nVidia's .deb package installs includes and libs of cudnn into different directories - replaced the original code with a query for gcc's includes (also not ideal, but it seems to be more reliable)

Please test before merging - I believe this code is generic and should not break on other distros, but haven't tested on anything non-debian-based.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -140,8 +140,8 @@ class Environment():
         if not (self.py_version[0].split(".")[0] == "3"
                 and self.py_version[0].split(".")[1] in ("3", "4", "5", "6", "7")
                 and self.py_version[1] == "64bit"):
-            self.output.error("Please run this script with Python version 3.3, 3.4, 3.5, 3.6 or 3.7 "
-                              "64bit and try again.")
+            self.output.error("Please run this script with Python version 3.3, 3.4, 3.5, 3.6 "
+                              "or 3.7 64bit and try again.")
             exit(1)
 
     def output_runtime_info(self):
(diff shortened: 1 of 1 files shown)


kilroythethird:
On my last system where i used py3.7 there where no wheels for tensorflow 1.13.* for python 3.7, only for py <= 3.6.
And while faceswap runs with tensorflow 1.14.* the graphs in the GUI won't work with that version due to changes in tensorboard.
This might have changed, so just a heads up that this could potential lead to problems.
*You could ofc. always compile tensorflow 1.13 for py > 3.6*

torzdf:
Yes, as @kilroythethird says we will not lift the python 3.6 requirement until we can get Keras Tensorboard logging to play nice with Tensorflow 1.13.1. Also, TF 1.12 is faster than TF 1.13.

I will be revisiting the TF 1.13 issue again soon, as it is currently a barrier to RTX users.

I'm happy to add any further checks for Cuda/cuDNN (it is a pain in the ass to detect. Especially cuDNN), but please either remove the check for Python 3.7 or add in an additional check that TF 1.13 is being run in conjunction with Python 3.7

martin-pr:
Fair enough, I did notice problems with logging, but didn't understand that the cause is Python version. I'll remove the first part of the pull request and try to clean-up the second part to make it as reliable as I can.

Thanks a lot!

deepfakes:
Oh. Things have moved on a bit. Sorry. @torzdf has updated to TF1.13 and bumped the Python versions.

You may need to review what is still required

martin-pr:
Ah, ok :) Let's close this pull request then - if I still encounter problems with the newer version, I'll open a new one.

Thanks a lot!

PR #756, opened 2019-06-11
Added keras version of FAN for plaidml backends

kilroythethird:
This PR adds a keras version of the FAN aligner to support plaidml backends.
I currently import keras in the global scope. If this should lead to multiprocessing problems we can move the import behind a backend check, but i'd prefer not to if we don't need to.

This is not tested with multithreading/processing nor with an nvidia card nor on windows.
Please test if it break the "normal" behaviour.

### DIFF
--- a/lib/plaidml_tools.py
+++ b/lib/plaidml_tools.py
@@ -202,5 +202,5 @@ def setup_plaidml(loglevel):
     os.environ["KERAS_BACKEND"] = "plaidml.keras.backend"
     plaid = PlaidMLStats(loglevel)
     plaid.load_active_devices()
-    logger.info("Using GPU: %s", plaid.active_devices)
+    logger.info("Using GPU: %s", [plaid.ids[i] for i in plaid.active_devices])
     logger.info("Successfully set up for PlaidML")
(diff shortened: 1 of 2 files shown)


torzdf:
Could you put this in as a separate plugin please? (fan_plaidml or fan_amd maybe). I'd rather have the separation in case of changes/running into issues down the line.

kilroythethird:
Sure. Sounds like the better idea actually.
Gotta love the plugin system.

torzdf:
I have added a `supports_plaidml` flag that defaults to False for extract plugins. You may want to push a change to your initialization script to set it to True. I will then check out this PR and see about getting it integrated asap.

Want to close the book on plaidML support

kilroythethird:
Sorry for the delay.
I also changed the "Using GPU:" log output in plaidml_tools to output the device ids again instead of the index. Hope that was ok.

torzdf:
I've split this off to a new plaidML branch as there are still a few bugs I'm finding around the plaidML implementation (logging on extract etc.), and I want to be able to fix those with this in place.

I will merge the branch directly back to staging to ensure that you maintain your contrib status.

Good work. Works great on my test AMD rig. Still need to do some testing on Nvidia platform.

kilroythethird:
Its basically just the old implementation of FS tbh ;)

PR #785, opened 2019-07-09
Multiple Tools->Alignment speed ups

kilroythethird:
This PR optimizes some code in `tools.lib_alignments.jobs`.
It mainly does this by pulling calculations out of loops.
Most of this is required because the original implementation probably assumed `self.alignments.hashes_to_frame` would be a dictionary, while it is in deed a property generating a dictionary. Doing this in the loops seriously harmed the performance.

I also removed `RemoveAlignments.remove_alignments` as this doesn't seem to be used anywhere.
If i am wrong about that please let me know and i can restore it.

Some numbers (working on ~18k faces/frames):

**get_multi_faces_faces**
~ 40it/s to ~ 190000it/s
**get_leftover_faces**
~40it/s to ~ 1000000it/s
**merge**
~ 38it/s to ~ 4300it/s
**rename_faces**
~32it/s to 85000it/s
**get_missing_frames**
~1300it/s to ~ 1200000it/s
**remove_frames**
~ 1000it/s to ~500000it/s

### DIFF
--- a/tools/lib_alignments/jobs.py
+++ b/tools/lib_alignments/jobs.py
@@ -114,11 +114,12 @@ class Check():
         """ Return Faces when there are multiple faces in a frame """
         self.output_message = "Multiple faces in frame"
         seen_hash_dupes = set()
+        hashes_to_frame = self.alignments.hashes_to_frame
         for item in tqdm(self.items, desc=self.output_message):
             filename = item["face_fullname"]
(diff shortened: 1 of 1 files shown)


torzdf:
This looks fine. I think remove_alignment is still hanging around from before filter_hashes was implemented (which does the same job)

PR #791, opened 2019-07-11
Switch to keras version for sort face

kilroythethird:
Changes the cv2 run vggface model to vggface2_resnet50 using keras.

Since i don't have a Nvidia GPU please test carefully before merging ;)


kilroythethird:
*Fixed potential bug with using CPU mode when the backend is tensorflow.*

kilroythethird:
Switched lambda layer to keras custom layer to be compatible with different python version.

PR #799, opened 2019-07-20
Update Copy Alignments to use last frame with face

tt-cube:
This PR changes the copy alignment function from copying the alignment in the last/next frame, to copying the alignment of the last/next frame with a face. Useful when frames were skipped during extraction.


tt-cube:
There is a bug when using copy previous on the first frame or copy next on the last frame, will fix

PR #801, opened 2019-07-20
Contrast and Brightness adjustment for color transfer plugin

andenixa:
- added brightness/contrast adjustments to color transfer plugin


torzdf:
Why have you added this to the color transfer plugin? I think it would be better as standalone

andenixa:
> Why have you added this to the color transfer plugin? I think it would be better as standalone

I think it fits the paradigm. I thought we discussed that at `#dev-chat`
There is no issue in creating a plugin but we don't have visual plugin chains atm to put it in front of color effects.

torzdf:
Color Transfer is pretty much a stock implementation, so i think my preference would be to not add to it....

I will look at how we can create a "multi-plugin" plugin, because I also appreciate that having it on it's own may not be particularly helpful.

PR #809, opened 2019-07-25
Do not `rename` files after `remove-faces` action

mdingena:
After a long sorting job, it's kind of annoying that `remove-faces` _also_ does the `rename` action for you. I might want to keep my files sorted after I clean up my alignments file with `remove-faces`. So I think it's better if we manually trigger `rename` action when we decide it's time to remove the sorting.


torzdf:
As discussed in Discord server, I would much prefer a flag "no-rename" or similar added to cli.py and checked for so that we can support both workflows.

Otherwise, thanks for the PR!

PR #812, opened 2019-07-27
documentation, pep8, style, clarity updates  - Prep for Segmentation

kvrooman:
Breaking up of a multi part PR into chunks. This first section is mostly cosmetic, style changes, documentation adds, and some minor edits

Cli
- group all A inputs together, group all B inputs together,
- move outputs from inputs and move down in the GUI list

Layers
- PEP8 cleanup
- addition of 3 custom layers for ease of use

Losses
- Clarification and simplification of variable names and code logic

Multithreading
- explicit hard coding of variables

Training Data
- line width within PEP8 for clarity

Mask_base
- use of round to int rather than int

Train_config
- documentation of new losses
- consolidation of some model specifics configs to universal configs
 - expose losses in config --- however, model/_base does not yet expose them for training

Model_base
- Default parameter for gpus

convert
- declare array with proper shape/type rather than conversion after creation

train
- Default parameter for gpus

Switch of Train Option Orderings
image


kvrooman:
Let me know but the two PEP8 comment flags by the bot are two things I'd ignore
- the first looks clean to me but can change if youd like
- the hardcoded matrix for edge detection is large, and its clarity is much improved if the matrix rows are not split to two lines

torzdf:
> 
> 
> Let me know but the two PEP8 comment flags by the bot are two things I'd ignore
> 
>     * the first looks clean to me but can change if youd like
> 
>     * the hardcoded matrix for edge detection is large, and its clarity is much improved if the matrix rows are not split to two lines

It's a bit annoying, I know, but the first one should be indented one space, as it should line up with the list, not the function call:
```py
        return pad(x,
                   [[0, 0],
                    [padding_top, padding_bot],
                    [padding_left, padding_right],
                    [0, 0]],
                   'REFLECT')
```

Add I would rather try to remain pep8 compliant (with line length of 100) if at all possible. I'm not really too bothered about extra lines, so this would be fine for me:
```py
    matrix = [[[[0.00070, 0.0007]],
               [[0.00520, 0.0037]],
               [[0.03700, 0.]],
               [[0.00520, -0.0037]],
               [[0.00070, -0.0007]]],
              [[[0.00370, 0.0052]],
               [[0.11870, 0.1187]],
               [[0.25890, 0.]],
               [[0.11870, -0.1187]],
               [[0.00370, -0.0052]]],
              [[[0.00000, 0.0370]],
               [[0.00000, 0.2589]],
               [[0.00000, 0.]],
               [[0.00000, -0.2589]],
               [[0.00000, -0.0370]]],
              [[[-0.0037, 0.0052]],
               [[-0.1187, 0.1187]],
               [[-0.2589, 0.]],
               [[-0.1187, -0.1187]],
               [[-0.0037, -0.0052]]],
              [[[-0.0007, 0.0007]],
               [[-0.0052, 0.0037]],
               [[-0.0370, 0.]],
               [[-0.0052, -0.0037]],
               [[-0.0007, -0.0007]]]]```

**Comment by kvrooman** [comment]
Pepped

File: plugins/train/_config.py [hunk 1]
```diff
@@ -41,21 +41,13 @@ def set_globals(self):
...
-                           "\n\t extended: Based on components mask. Extends the eyebrow points "
-                           "to further up the forehead. May perform badly on difficult angles."
-                           "\n\t facehull: Face cutout based on landmarks")
         self.add_item(
             section=section, title="icnr_init", datatype=bool, default=False,
-            info="Use ICNR Kernel Initializer for upscaling.\nThis can help reduce the "
-                 "'checkerboard effect' when upscaling the image.")
+            info="\nUse ICNR to tile the default initializer in a repeating pattern. \n"
+                 "This strategy is designed for  sub-pixel / pixel shuffler upscaling \n"
+                 "and should only be used on upscaling layers. This can help reduce the \n"
```
torzdf: I'm not sure you need to say that it should only be used on upscaling layers here. This is an end user option, so it will either be meaningless or confusing to most. A better place might be in the docstring for the function

plugins/train/_config.py
```diff
@@ -68,23 +60,249 @@ def set_globals(self):
...
+            info="\nImage loss function is weighted by mask presence. For areas of \n"
+                 "the image without the facial mask, reconstuction errors will be \n"
+                 "ignored while the masked face area is prioritized. May increase \n"
+                 "overall quality by focusing attention on the core face area.\n")
+        self.add_item(
+            section=section, title="image_loss_function", datatype=str,
+            default="Mean_Absolute_Error",
+            choices=["Mean_Absolute_Error", "Mean_Squared_Error", "LogCosh",
+                     "Smooth_L1", "L_inf_norm", "SSIM", "GMSD", "Pixel_Gradient_Difference"],
+            info="\nGeneral Loss Discussion \n"
```
torzdf: I admire the dedication to providing maximum possible information, but any attempt to make this slightly less verbose would be appreciated!

plugins/train/_config.py
```diff
@@ -68,23 +60,249 @@ def set_globals(self):
...
+            section=section, title="learning_rate", datatype=float, default=5e-5,
+            min_max=(1e-6, 1e-4), rounding=6, fixed=False,
+            info="Learning rate - how fast your network will learn (how large are \n"
+                 "the modifications to the model weights after one batch of training).\n"
+                 "Values that are too large might result in model crashes and the \n"
+                 "inability of the model to find the best solution.\n"
+                 "Values that are too small might be unable to escape from dead-ends \n"
+                 "and find the best global minimum.")
+
+        # << ORIGINAL MODEL OPTIONS >> #
```
torzdf: Bad news. All of these model specific options need to go. They are now handled in the relevant _default.py with the model

plugins/train/_config.py
kvrooman: K, I'll move that around to docstring

plugins/train/_config.py
kvrooman: I'll strip some -- Like I said earlier, I was fulfilling a request. Ha

plugins/train/_config.py
kvrooman: Gah - I'll look at how to integrate with the new method

torzdf: Fortunately I have documented them extensively!

lib/model/losses.py
```diff
@@ -327,171 +325,145 @@ def instnorm():
...
-                   (2.0*K.tanh(var_a/2.0)) + 0.5 + tf.floor(var_x/width))
-    loss += 1e-10
-    return K.mean(loss, axis=-1)
+def l_p_norm(y_true, y_pred, p_norm=np.inf):
+    """
+    Calculate the L-p norm as a loss function,
+    valid choics of p are [0,1,no.inf]
+    """
+    diff = y_true - y_pred
+    loss = tf.norm(diff, ord=p_norm, axis=-1)
```
kilroythethird: Note to myself: Write keras fallback norm supporting all `ord`s

torzdf:
> 
> 
> Hello @kvrooman! Thanks for updating this PR. We checked the lines you've touched for [PEP 8](https://www.python.org/dev/peps/pep-0008) issues, and found:
> 
>     * In the file [`plugins/train/_config.py`](https://github.com/deepfakes/faceswap/blob/f1a881dc288fc3992961ac05afa22c1b65032688/plugins/train/_config.py):
> 
> 
> > [Line 39:12](https://github.com/deepfakes/faceswap/blob/f1a881dc288fc3992961ac05afa22c1b65032688/plugins/train/_config.py#L39): [W291](https://duckduckgo.com/?q=pep8%20W291) trailing whitespace
> > [Line 73:100](https://github.com/deepfakes/faceswap/blob/f1a881dc288fc3992961ac05afa22c1b65032688/plugins/train/_config.py#L73): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (100 > 99 characters)
> > [Line 149:17](https://github.com/deepfakes/faceswap/blob/f1a881dc288fc3992961ac05afa22c1b65032688/plugins/train/_config.py#L149): [E131](https://duckduckgo.com/?q=pep8%20E131) continuation line unaligned for hanging indent
> 
> ##### Comment last updated at 2019-08-02 06:32:36 UTC

Sorry, please resolve these and I;ll review

plugins/train/model/_base.py
```diff
@@ -666,36 +666,35 @@ def update_loss_names(self):
...
-                if not predict and side.lower() == "a":
-                    logger.verbose("Penalizing mask for Loss")
+                logger.verbose("Image loss function is weighted by mask presence")
+                loss_funcs.append(PenalizedLoss(mask[0], loss_dict[img_loss_config]))
+                logger.debug("final face loss: %s", img_loss_config)
             else:
-                logger.debug("face loss func: %s", loss_func)
-                loss_funcs.append(loss_func)
-        logger.debug(loss_funcs)
+                logger.verbose("Using %s loss function for image", img_loss_config)
```
torzdf: If doing verbose, then you should keep the check for "predict" and "side" to avoid logging twice, and logging loss on predict.

i.e:
```py
            if not predict and side.lower() == "a":
                logger.verbose(<logging message>)
```

Also applies to line 684

kvrooman: k, I had some of the prior logic on logging from before the Loss class.. i'll look again to incorporate better

kvrooman:
> > Hello @kvrooman! Thanks for updating this PR. We checked the lines you've touched for [PEP 8](https://www.python.org/dev/peps/pep-0008) issues, and found:
> > ```
> > * In the file [`plugins/train/_config.py`](https://github.com/deepfakes/faceswap/blob/f1a881dc288fc3992961ac05afa22c1b65032688/plugins/train/_config.py):
> > ```
> > 
> > 
> > > [Line 39:12](https://github.com/deepfakes/faceswap/blob/f1a881dc288fc3992961ac05afa22c1b65032688/plugins/train/_config.py#L39): [W291](https://duckduckgo.com/?q=pep8%20W291) trailing whitespace
> > > [Line 73:100](https://github.com/deepfakes/faceswap/blob/f1a881dc288fc3992961ac05afa22c1b65032688/plugins/train/_config.py#L73): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (100 > 99 characters)
> > > [Line 149:17](https://github.com/deepfakes/faceswap/blob/f1a881dc288fc3992961ac05afa22c1b65032688/plugins/train/_config.py#L149): [E131](https://duckduckgo.com/?q=pep8%20E131) continuation line unaligned for hanging indent
> > 
> > 
> > ##### Comment last updated at 2019-08-02 06:32:36 UTC
> 
> Sorry, please resolve these and I'll review

pepped

plugins/train/model/_base.py
kvrooman: removed verbose and kept the prior debug level

torzdf:
Ok, sorry. A few outstanding things......

**coverage**
Thanks for moving coverage. The help text should probably be amended though. Something like:
```
How much of the extracted image to train on. Higher coverage will train on more of the face,
but come at a detail cost, as more face is being fit into the same size image. Sensible values
to use are:
...
```
**loss options**
The loss description is much better now... I'm going to make a couple of requests though.

I would suggest changing `image_loss_function` to just `loss_function`.

Please abbreviate those that can be abbreviated (e.g. `Mean_Absolute_Error` > `mae`). The descriptions tell us what the abbreviations mean.

Also, please lowercase the options for these. It helps us handle in the code as we can lower() any of the options to make it case insensitive.

Both of these make it easier for cli users and also ensures less room for error

**dssim**
Now that DSSIM has been removed, then the old values for True/False will have to be handled when loading legacy models, and the associated state file updated accordingly to point to the correct loss. This is the first time that a config option has been explicitly changed, so we currently have no code to handle this. I would imagine that the best place to put it would be in `model._base.py` within `State().load()` with a call for check + update legacy options there.

**missing defaults**
We should probably remove any `_defaults.py` files for models which have no options to get them out of the config file/GUI tabs. I'm not sure if they can just be deleted or will have to insert a bit of code to handle. I can do this after merge if you'd prefer.

PR #813, opened 2019-07-27
DRAFT --- Model and Conversion Changes - Prep for Segmentation

kvrooman:
Only to be merged following the previous Prep for Segmentation PR on clarity, docs

**Masks as an Input to Models**
- The toolchain is significantly complicated by the fact models can either have a mask or not have a mask. Logic is needed through the code base to determine the structure of the model, the length of the inputs, and whether a mask needs to be applied or not.

- This can be simplified by always applying a mask to training and conversion or allowing it still as an option / flag

- If the user strictly wants to avoid using a mask still, a dummy mask of all 1s can be used in place to arrive at the exact same outcome, while streamlining application in code

- This creates a backward compatibility issue however for identifying and modifying previous models without a mask input

- The solution is to flag such models, modify their graphs to use two inputs, and resave legacy models in the updated format

**Use of Masks in Training Models**
- Masks should not be used in the decoder when the model is designed to learn its own mask. Peeking it the ground truth will degrade performance in out of sample application

- Masks can and should be used in the encoder to inform the encoder on which pieces of the input image are important to compress into the latent space and which pieces of the input image ( background ) should be ignored. This information is not present currently and likely causes the encoder to encode useless background data in the space limited latent space


lib/model/losses.py
```diff
@@ -509,46 +481,52 @@ def scharr_edges(image, magnitude):
...
+               [[-0.0007, -0.0007]]]]
+    num_kernels = [2]
+    kernels = K.constant(matrix, dtype='float32')
+    kernels = K.tile(kernels, [1, 1, image_shape[-1], 1])
+
 
     # Use depth-wise convolution to calculate edge maps per channel.
+    # Output tensor has shape [batch_size, h, w, d * num_kernels].
     pad_sizes = [[0, 0], [2, 2], [2, 2], [0, 0]]
     padded = tf.pad(image, pad_sizes, mode='REFLECT')
```
kilroythethird: Could this be changed to use plaidml pad or tf. pad depending on backend please ?
See top of layers.py for this.

lib/model/losses.py
```diff
@@ -509,46 +481,52 @@ def scharr_edges(image, magnitude):
...
-    if magnitude:  # magnitude of edges
-        output = tf.sqrt(tf.reduce_sum(tf.square(output), axis=-1))
-    else:  # direction of edges
-        output = tf.atan(tf.squeeze(tf.div(output[:, :, :, :, 0]/output[:, :, :, :, 1])))
+    if not magnitude:  # direction of edges
+        # Reshape to [batch_size, h, w, d, num_kernels].
+        shape = K.concatenate([image_shape, num_kernels], axis=0)
+        output = K.reshape(output, shape=shape)
+        output.set_shape(static_image_shape.concatenate(num_kernels))
+        output = tf.atan(K.squeeze(output[:, :, :, :, 0] / output[:, :, :, :, 1]))
```
kilroythethird: Not for myself. Add `plaidml.tile.unary_op(K.ones(10), "atan(I)")` as plaidml fallback.

lib/model/losses.py
```diff
@@ -509,46 +481,52 @@ def scharr_edges(image, magnitude):
...
-
-    if magnitude:  # magnitude of edges
-        output = tf.sqrt(tf.reduce_sum(tf.square(output), axis=-1))
-    else:  # direction of edges
-        output = tf.atan(tf.squeeze(tf.div(output[:, :, :, :, 0]/output[:, :, :, :, 1])))
+    if not magnitude:  # direction of edges
+        # Reshape to [batch_size, h, w, d, num_kernels].
+        shape = K.concatenate([image_shape, num_kernels], axis=0)
+        output = K.reshape(output, shape=shape)
+        output.set_shape(static_image_shape.concatenate(num_kernels))
```
kilroythethird: Could this two lines be changed to `output = K.reshape(output, shape=static_image_shape.concatenate(num_kernels))`, or did i misunderstand the set_shape function ?

kvrooman:
Other PR incorporates better method and have implemented a new method for adding masks to encoder. Cleaning this up and closing

PR #823, opened 2019-08-04
S3FD keras detector + FAN-amd updates + AMD goodies

kilroythethird:
Changes/Additions in this PR:

- Added S3FD keras version (using batches and multithreading to increase performance)
- Added batches and multithreading to FAN keras version to increase performance
- Added config support to Align plugins
- Added pad_to parameter to extract `compile_detection_image` to support padding
- Added workaround for convert with amd cards to use bs=1 when the batch is != requested batch_size to prevent plaidml from compiling new kernels with frames where multiple or none faces are found in the frame
- Added minor bugfix to extract `reload_images` where -een did lead to warnings being generated for each skipped frame when using -sp.
- Added `multiprocessing_queue` parameter to queue_manager.get_queue to make getting a non multiprocessing queue more convenient

Both Keras version of FAN and S3FD have a config setting for the batch size.
- Fan default is 16 which is about 1.6GB vram on my system.
- S3FD default is 8 which is about 2GB vram on my system.

As mentioned in the dev-chat i'd like to hide both keras models from the parameter and GUI and auto select them when --amd is provided.
If that is ok i could add these changes to this PR or send it in another one.?


kilroythethird:
- Added rotation support for s3fd-amd.
- Changed detector._base.compile_detection_image method to return the used padding if pad_to parameter is given,
- Removed the -r deprecation message.

From my point of view it should be good to go now.

torzdf:
This PR is deprecated by the S3FD-AMD push isn't it?

kilroythethird:
Yup. Its only still here as template for the changes i did for FAN-keras, but i am going to close it.

PR #837, opened 2019-08-13
Added s3fd-amd + minor fixes

kilroythethird:
S3FD-amd part of #823

Additionally contains:
- Added multiprocessing_queue parameter to queue_manager.get_queue to make getting a non multiprocessing queue more convenient
- Minor bugfix to extract reload_images where -een did lead to warnings being generated for each skipped frame when using -sp.
- Workaround for convert with amd cards to use bs=1 when the batch is != requested batch_size to prevent plaidml from compiling new kernels with frames where multiple or none faces are found in the frame (Sorry for smuggling this in ;))

### DIFF
--- a/plugins/extract/detect/s3fd_amd.py
+++ b/plugins/extract/detect/s3fd_amd.py
@@ -1,0 +1,481 @@
+#!/usr/bin/env python3
+""" S3FD Face detection plugin
+https://arxiv.org/abs/1708.05237
+
+Adapted from S3FD Port in FAN:
+https://github.com/1adrianb/face-alignment
+"""
+
+from scipy.special import logsumexp
+import numpy as np
+from ._base import Detector, logger
+import keras
+import keras.backend as K
+from lib.multithreading import FSThread
+from lib.queue_manager import queue_manager
+import queue
+
+
+class Detect(Detector):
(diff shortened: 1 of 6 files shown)


plugins/extract/detect/s3fd_amd.py
```diff
@@ -0,0 +1,480 @@
...
+        """ Load the keras Model and weights """
+        logger.verbose("Initializing S3FD_amd Network model...")
+        layers = {
+            'O2K_Add': O2K_Add, 'O2K_Slice': O2K_Slice,
+            'O2K_Sum': O2K_Sum, 'O2K_Sqrt': O2K_Sqrt,
+            'O2K_Pow': O2K_Pow, 'O2K_ConstantLayer': O2K_ConstantLayer,
+            'O2K_Div': O2K_Div
+        }
+        model = keras.models.load_model(self.model_path, custom_objects=layers)
+        return model
```
torzdf: This hangs on Tensorflow Backend. Unfortunately Keras does not play nicely with threading with the TF backend.

You can make it play nicely by adding the following (unfortunately protected access) function prior to returning the model:
```py
        model._make_predict_function()  # pylint: disable=protected-access
```

PR #844, opened 2019-08-26
model_trainable_opts

andenixa:
- Added the ability to Enable/Disable of training Encoder and Decoders to global config file

### DIFF
--- a/plugins/train/_config.py
+++ b/plugins/train/_config.py
@@ -169,17 +169,17 @@ class Config(FaceswapConfig):
         self.add_item(
             section=section, title="encoder_trainable", datatype=bool,
             default=True, fixed=False, group="optimizer",
-            info="Specify whther to train Encoder. "
+            info="Specify whether to train Encoder. "
                  "\nHelpful with pre-trained models. ")
         self.add_item(
             section=section, title="decoder_a_trainable", datatype=bool,
             default=True, fixed=False, group="optimizer",
-            info="Specify whther to train Decoder A(old face). "
+            info="Specify whether to train Decoder A(old face). "
                  "\nHelpful with pre-trained models. ")
         self.add_item(
             section=section, title="decoder_b_trainable", datatype=bool,
             default=True, fixed=False, group="optimizer",
-            info="Specify whther to train Decoder B(new face). "
+            info="Specify whether to train Decoder B(new face). "
                  "\nHelpful with pre-trained models. ")
 
     def load_module(self, filename, module_path, plugin_type):
--- a/plugins/train/_config.py
+++ b/plugins/train/_config.py
@@ -175,12 +175,12 @@ class Config(FaceswapConfig):
             section=section, title="decoder_a_trainable", datatype=bool,
             default=True, fixed=False, group="optimizer",
             info="Specify whther to train Decoder A(old face). "
-                 "\nHelpful with pre-trained models. ")      
+                 "\nHelpful with pre-trained models. ")
         self.add_item(
             section=section, title="decoder_b_trainable", datatype=bool,
             default=True, fixed=False, group="optimizer",
             info="Specify whther to train Decoder B(new face). "
-                 "\nHelpful with pre-trained models. ")                                              
+                 "\nHelpful with pre-trained models. ")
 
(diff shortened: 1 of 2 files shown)


plugins/train/model/_base.py
```diff
@@ -356,6 +358,27 @@ def reset_pingpong(self):
         self.compile_predictors(initialize=False)
         logger.debug("Reset models")
 
+    def set_networks_trainable(self):
+        """ Set model's networks trainable state prior to compiling """
+        logger.debug("Setting Networks traiable state")
```
kilroythethird: Nitpick: typo in "traiable" ;)

torzdf:
Thanks for this.....
We *may* need to rework it a little though, Basically, how would it work for IAE (for example).

plugins/train/_config.py
```diff
@@ -166,6 +166,21 @@ def set_globals(self):
                  "inability of the model to find the best solution. "
                  "Values that are too small might be unable to escape from dead-ends "
                  "and find the best global minimum.")
+        self.add_item(
+            section=section, title="encoder_trainable", datatype=bool,
+            default=True, fixed=False, group="optimizer",
+            info="Specify whther to train Encoder. "
```
gjgd: typo whther -> whether

plugins/train/_config.py
```diff
@@ -166,6 +166,21 @@ def set_globals(self):
...
+                 "\nHelpful with pre-trained models. ")
+        self.add_item(
+            section=section, title="decoder_a_trainable", datatype=bool,
+            default=True, fixed=False, group="optimizer",
+            info="Specify whther to train Decoder A(old face). "
+                 "\nHelpful with pre-trained models. ")
+        self.add_item(
+            section=section, title="decoder_b_trainable", datatype=bool,
+            default=True, fixed=False, group="optimizer",
+            info="Specify whther to train Decoder B(new face). "
```
gjgd: typo whther -> whether

plugins/train/_config.py
```diff
@@ -166,6 +166,21 @@ def set_globals(self):
...
                  "and find the best global minimum.")
+        self.add_item(
+            section=section, title="encoder_trainable", datatype=bool,
+            default=True, fixed=False, group="optimizer",
+            info="Specify whther to train Encoder. "
+                 "\nHelpful with pre-trained models. ")
+        self.add_item(
+            section=section, title="decoder_a_trainable", datatype=bool,
+            default=True, fixed=False, group="optimizer",
+            info="Specify whther to train Decoder A(old face). "
```
gjgd: typo whther -> whether

plugins/train/model/_base.py
```diff
@@ -356,6 +358,27 @@ def reset_pingpong(self):
...
+        """ Set model's networks trainable state prior to compiling """
+        logger.debug("Setting Networks trainable state")
+
+        train_encoder = self.config.get("encoder_trainable", True)
+        train_decoder_a = self.config.get("decoder_a_trainable", True)
+        train_decoder_b = self.config.get("decoder_b_trainable", True)
+
+        encoder = self.networks['encoder'].network
+        for layer in encoder.layers:
+            layer.trainable = train_encoder
```
kvrooman: ? This won't set any of the layers in encoder.layers as trainable. Iterating over a iterator won't change the underlying items. You have to index though the iterator to actually affect them.
```
for index, layer in enumerate(encoder.layers):
    encoder.layers[index] = train_encoder
```

andenixa:
@kvrooman I've seen other developers do `isinstance(layer, Dense): layer.trainable=trainable`,
perhaps I should also be more selective instead of "brute-forcing" it.
The setting probably won't work for all models layouts as @torzdf mentioned.
Thanks for reviewing. I shall think of a reasonable docstring.

kvrooman:
Looking at the one odd model ( IAE ) . Think it may be good to have the encoder.trainable flag logic you've added apply to both the "encoder" , "intermediate_a", "intermediate_b" sub-models if they exist.

torzdf:
This also impacts SAE LIAE

torzdf:
Closing due to lack of activity, but if this is something that may be useful, then we can look to revisit in the future.

PR #849, opened 2019-08-29
Added simple travis tests

kilroythethird:
This adds simple test cases using travis.
It does this by running different faceswap job with sub processes.
Namely:
- Extract from short video
- Extract from image
- Sort
- Rename
- Train 5 iterations with lightweight
- Convert images
- Convert video

These tests are only to detect crashes and hangs.
To test for correct results we might want to add some tests using pytest at some later time.

I don't have permissions to enable travis tests for this repo. Please activate it at:
https://travis-ci.org/deepfakes/faceswap

It shouldn't matter if the tests on travis are activated for all branches as it will only run when the .travis file exists.

For an example run see: https://travis-ci.org/kilroythethird/faceswap

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,7 @@
 !plugins/convert/*
 !tools
 !tools/lib*
+!.travis.yml
 *.ini
 *.pyc
 __pycache__/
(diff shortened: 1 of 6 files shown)


kvrooman:
This cycles through all the parameter combos and the config options too ? I had a long debugging process especially on all the config options.

PR #851, opened 2019-08-30
Align eyes deprecation

kvrooman:
Over the last year, the amount of options and toggles has been slowly increasing in FS. In order to manage this growth, I wanted to re-evaluate align_eyes. The face detection and face alignment models have been greatly improved with the additional of superior models and the retirement of older inaccurate ones. This largely removes the primary case for align_eyes which is to vertically align the pupils of the face due to inaccuracies in the alignment model and/or umeyama fitting routine.

In addition, this feature has also been rendered less useful by the fact that the training input face images are randomly rotated anyways prior to training. This rotation augmentation directly makes the align_eyes feature useless.  For this reason and in order to limit the option creep in the ui, I'd like to deprecate align_eyes.

I'm open to hearing any reasons or use cases that I have overlooked that call for this toggle's continuing inclusion. Thank you.


torzdf:
I'm all for this, to be honest.

torzdf:
One small bit of feedback.... It's annoying that we have to insert `pylint:disable=no-member` for cv2, but linter complains otherwise :(

My preference is to keep it on the line which triggers the linter. Putting it at the top of the function means that it will ignore any actual "no-member" errors that occur.

kvrooman:
Not much feedback one way or another from other folks. Leave up for comments for two more days and then merge?

torzdf:
I have no issue with that

linnik:
Instead of completely removing this functionality, what do you think about moving it out to separate script?

kvrooman:
Per feedback, I have added a tool to replicate the functionality. This simplifies the code-base but still allows a means of aligning a image dataset if absolutely desired.

tools/cli.py
```diff
@@ -25,31 +25,36 @@ def get_argument_list(self):
...
         argument_list = list()
         argument_list.append({
             "opts": ("-j", "--job"),
             "action": Radio,
             "type": str,
-            "choices": ("draw", "extract", "extract-large", "manual", "merge",
-                        "missing-alignments", "missing-frames", "legacy", "leftover-faces",
-                        "multi-faces", "no-faces", "reformat", "remove-faces", "remove-frames",
-                        "rename", "sort-x", "sort-y", "spatial", "update-hashes"),
+            "choices": ("draw", "extract", "extract-large", "extract-aligneyes", "manual",
```
torzdf: Could you move this back to an extract option please?

2 reasons:
1) The number of job options is already inordinately large, so would rather avoid adding more options here (I may change extract-large to be an extract option too)
2) We just convert it to a bool anyway, which is what the option will come in as, so there is no significant additional coding to be honest.

kvrooman: I'm confused? Making it an extraction option defeats the whole purpose of deprecating it from the extract options and extract toolchain ?

User feedback was to add it to tools if were removing it from extract options, so I added it here.

torzdf: I mean in the align tools menu (it currently resides in there as an option)

torzdf: image

kvrooman: yup, realized.. will do

torzdf:
And sorry. I conflicted your commit again :(

kvrooman:
Merged conflicts. Re-ran some testing and looks good.

PR #858, opened 2019-09-03
Replaced multiprocessing in training_data with threading

kilroythethird:
This is little bit faster but, more important, removes a lot of complexity.
Need to be tested on windows and macOs if possible.

### DIFF
--- a/lib/multithreading.py
+++ b/lib/multithreading.py
@@ -178,7 +178,7 @@ class MultiThread():
     @property
     def errors(self):
         """ Return a list of thread errors """
-        return [thread.err for thread in self._threads]
+        return [thread.err for thread in self._threads if thread.err]
 
     def check_and_raise_error(self):
         """ Checks for errors in thread and raises them in caller """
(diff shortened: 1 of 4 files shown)


lib/multithreading.py
```diff
@@ -462,31 +217,41 @@ def join(self):
...
-class BackgroundGenerator(threading.Thread):
+class BackgroundGenerator(FSThread):
     """ Run a queue in the background. From:
         https://stackoverflow.com/questions/7323664/ """
     # See below why prefetch count is flawed
-    def __init__(self, generator, prefetch=1):
-        threading.Thread.__init__(self)
-        self.queue = Queue.Queue(maxsize=prefetch)
+    def __init__(self, generator, prefetch=1, queue=None):
+        FSThread.__init__(self, target=self._run)
```
torzdf:
```py
super().__init__(target=self._run)
```
Just for consistency

lib/multithreading.py
```diff
@@ -462,31 +217,41 @@ def join(self):
...
+        self.queue = queue or Queue.Queue(prefetch)
         self.generator = generator
         self.daemon = True
         self.start()
 
-    def run(self):
+    def _run(self):
```
torzdf: Is this right? Changing this surely stops it from overriding its parent?

plugins/train/trainer/_base.py
```diff
@@ -25,8 +25,8 @@
 import os
 import time
 
-import cv2
 import numpy as np
+import cv2
```
torzdf: [*super anal mode*] I prefer imports to be in alphabetical order

kilroythethird: I did it this way because numpy set environment variables to ensure openblas doesnt set the cpu affinity and i wasn't sure if opencv might use openblas too. Doesn't seem like it tho.

torzdf: I did some testing with importing numpy in faceswap.py. I didn't observe any difference.

I will test on Windows before posting my full results.

kilroythethird:
I made the BackgroundGenerator use multiple threads.
Trainingdata currently uses 2 threads per side.
This brought up my performance from ~80 img/s for both sides to ~ 113.

Also fixed an error in `MultiThread` where `check_and_raise_error` would have failed if anything else except the first thread raised an exception.

PR #859, opened 2019-09-03
Minor style update for Nvidia spelling

therumbler:
minor nVidia -> Nvidia update


torzdf:
Yeah, this is either or. We will keep the styling as is.

Thanks for taking an interest though

PR #871, opened 2019-09-14
De-Multiprocess Extract

torzdf:
This PR removes multiprocessing from the extraction pipeline, and fully removes multiprocessing from the entire project.

The result is a more maintainable code base, and better extraction speeds. 

MTCNN has been ported to be supported by AMD cards, and batching has been added for both the S3FD detector and FAN aligner (in config options).

This PR also fixes the S3FD multi process bug and the Tensorflow 1.14 logging bug


torzdf:
> 
> 
> Hello @torzdf! Thanks for updating this PR. We checked the lines you've touched for [PEP 8](https://www.python.org/dev/peps/pep-0008) issues, and found:
> 
>     * In the file [`docs/conf.py`](https://github.com/deepfakes/faceswap/blob/e70623b3c495e02c3a1aa07b7e42df0e9bfadbfd/docs/conf.py):
> 
> 
> > [Line 34:1](https://github.com/deepfakes/faceswap/blob/e70623b3c495e02c3a1aa07b7e42df0e9bfadbfd/docs/conf.py#L34): [E124](https://duckduckgo.com/?q=pep8%20E124) closing bracket does not match visual indentation

This is what happens when you use auto-generated scripts!

PR #874, opened 2019-09-18
Optimized mtcnn a bit + added batching

kilroythethird:
- Reduced image size for mtcnn from `1440**2` to `640**2`. This increases speed dramatically and didn't have any negative impact in my tests.
- Added batching to mtcnn.
- Set queue_size for all intermediate queues to 1. Ie: if the next step in the pipeline is blocking 2 full batches will be waiting (1 in queue, 1 waiting to be added to the queue). I didn't saw any speed difference between size 1 and higher sizes.
- Added optimized prediction for amd backends which tries to minimize the kernels to be compiled while keeping the BS as high as possible. Only active when `batch_size` parameter is provided in the predict call.

Batching in mtcnn is only done partial (altho for the most relevant part). There is still room for performance optimization.
`mtcnn.vram_per_batch` still needs to be set to the appropiate value (no nvidia card here to test)

Example speed difference with tf-CPU:
- 01:07 minutes for 300 images with current staging
- 00:19 minutes for 300 images with this PR

Changes with the AMD backend are also noticeable, altho not as big due to the kernel compile time.

### DIFF
--- a/lib/model/session.py
+++ b/lib/model/session.py
@@ -5,6 +5,7 @@ import logging
 
 import tensorflow as tf
 from keras.models import load_model as k_load_model, Model
+import numpy as np
 
 from lib.utils import get_backend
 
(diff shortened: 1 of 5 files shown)


torzdf:
Good work

PR #891, opened 2019-10-02
Feat: add wandb logging

borisdayma:
Training runs are now logged through Weights & Biases to plot training losses & accuracy graphs as well as sample predictions.
See example: https://app.wandb.ai/borisd13/faceswap/runs/vz384aby?workspace=user-borisd13


kvrooman:
@torzdf seems to be similiar if not exact capabilities as Tensorboard logging? Thoughts?

torzdf:
@borisdayma Thanks for the PR. Could you explain what this gives us that Tensorboard doesn't?

borisdayma:
@torzfd W&B is a superset of Tensorboard.
To try to summarize quickly, Tensorboard is for a single user training a specific model, W&B is for collaboration on a training paradigm: many model variants and experiments by many users, persisting in the cloud (or optionally in local storage) with easy reproducibility over the lifetime of a project, and it's framework agnostic, and it tracks your system/hardware metrics, and lets you log way more formats.

You can see this as an example: https://app.wandb.ai/borisd13/faceswap/runs/vz384aby?workspace=user-borisd13

Those runs are long but let's say I (or other people) did several with different parameters, we would be able to compare them together. Also it's always convenient for reproducibility to log and keep track of an experiment.

deepfakes:
First up, thanks for the Pull Request.

At an absolute minimum PRs need to meet PEP8 standards, and have to make [pep8speaks happy](https://github.com/deepfakes/faceswap/pull/891#issuecomment-537658611) , so this PR will not be accepted until then.

With that said, I do see some value in being able to share data, however I have some outstanding concerns.

1) If we were to accept this PR then it would have to be Opt-In rather than Opt-Out. This is standard practice for information stored on remote servers and should not come as a surprise.

2) This prompts the user to create an account (from the cli) when running with WandDB:
```
wandb: W&B is a tool that helps track and visualize machine learning experiments
wandb: (1) Private W&B dashboard, no account required
wandb: (2) Create a W&B account
wandb: (3) Use an existing W&B account
wandb: (4) Don't visualize my results
wandb: Enter your choice: 
```
This is, unfortunately, not a workable solution for our users, as many, if not most, use the GUI to run Faceswap and will not be able to enter information in the displayed console window.

3) Monetization. Whilst I see that there is some benefit to this tool, the primary goal will be to monetize for WanDB. Faceswap is entirely open source, with no paywall for our users at any stage. We have not looked to monetize and have no plans to do so in future. We are driven entirely by voluntary donations. We would need assurances that:
    - The T+Cs would not change and users who previously were able to use wandDB for free would not suddenly find that they have to pay for the same service.
    - That any revenue generated by wanDB for faceswap users would be shared with faceswap. The details of this would clear need to be ironed out.

4) I would require clarification about what you intend to do with the data that you collect from our users.

These are all deal breakers, from our perspective.

I have posted this as a comment on this PR as I would like to remain as transparent as possible to our users about our plans for Faceswap. You are welcome to continue this discussion with me at: deepfakesrepo@gmail.com

borisdayma:
Thank you so much for this detailed feedback @deepfakes . I'll try to answer the best I can.

> At an absolute minimum PRs need to meet PEP8 standards, and have to make [pep8speaks happy](https://github.com/deepfakes/faceswap/pull/891#issuecomment-537658611) , so this PR will not be accepted until then.

Noted, I'll make sure it is PEP8 compliant.

> With that said, I do see some value in being able to share data, however I have some outstanding concerns.
> 
> 1. If we were to accept this PR then it would have to be Opt-In rather than Opt-Out. This is standard practice for information stored on remote servers and should not come as a surprise.

That's a fair point. Actually the integration to W&B is more valuable to ML Engineers/Scientists who try to fine-tune the architecture. It would be less valuable to people "just" playing with the GUI. So I can definitely change the CLI arg to be "no W&B logging" by default.

> 2. This prompts the user to create an account (from the cli) when running with WandDB:
> 
> ```
> wandb: W&B is a tool that helps track and visualize machine learning experiments
> wandb: (1) Private W&B dashboard, no account required
> wandb: (2) Create a W&B account
> wandb: (3) Use an existing W&B account
> wandb: (4) Don't visualize my results
> wandb: Enter your choice: 
> ```
> 
> This is, unfortunately, not a workable solution for our users, as many, if not most, use the GUI to run Faceswap and will not be able to enter information in the displayed console window.

No problem, the library actually won’t show this menu if it doesn’t have access to stdin.  When that’s the case it doesn’t send any metrics to the cloud, instead only logging locally.  The first option in the menu allows users to log metrics to W&B without an account so we can also see what possibilities there are there.

@lukas (CEO of Weights & Biases) can you reply to the next points?

Thanks again for those comments which are very useful to ensure W&B can benefit the open-source community. This is definitely a great feedback for them.

I've been personally using it for my own projects and I feel it can help the AI community with more transparency and reproducibility especially when publishing new papers.

lukas:
@deepfakes - we would really like wandb to be useful for open source repos.  We are committed to making the tool always free to open source repositories.  I would love to talk to you about how we could make that commitment feel binding and what uses of data feel good/bad.  Would you be up for jumping on a hangout?

borisdayma:
@deepfakes I implemented the following changes per your comments:

- I made pep8speaks happy
- logging through W&B is now opt-in by using `-wandb` or `--wandb-logs`
- there is no more prompt from the cli (logging is anonymous if wandb has not been configured) so Travis CI check passed

Let me know if anything else is needed.

borisdayma:
@deepfakes I just wanted to check if you needed anything else from me.

torzdf:
Sorry for sitting on this PR for so long. I completely forgot about it (and thought I had got back to you), I am cleaning up PRs as things have moved on.

We discussed this amongst our Devs, but decided that we wouldn't include this in our repo, as we want to avoid plugging into 3rd party libraries where possible.

Thanks for taking the time though.

PR #909, opened 2019-10-25
BUGFIX: Nn_blocks upscale

kvrooman:
When up-scaling by a factor other than 2x, upscale does not provide the correct logic as it is hard-coded for a 2x upscale factor


linnik:
@torzdf what are all these changes? Should I be looking only for a this small one https://github.com/deepfakes/faceswap/commit/bb26afde3494d2f06a8225f2d908708a845889c1?diff=split#diff-3445dbb2b17101f403f25d977ac60d7f or everything interconnected? For me it looks like the whole smart-mask-alpha branch got merged within this pull request.

torzdf:
This was a push against the wrong branch :/

Have had to force reset it.

PR #910, opened 2019-10-25
BUGFIX: recursive loading of images all sub-directories in sorter

kvrooman:
Sort uses os.walk to find images in a directory. This logic will recursively walk that directory and __every sub-directory__ to find images. This recursive nature is unintended and potentially dangerous.


tools/sort.py
```diff
@@ -620,6 +620,7 @@ def find_images(input_dir):
             for file in files:
                 if os.path.splitext(file)[1].lower() in extensions:
                     result.append(os.path.join(root, file))
+            break
```
torzdf: Very hacky ;)

PR #922, opened 2019-11-01
I want to fix the double 'fsa' suffix issue

50mkw:
I want to fix the double 'fsa' suffix issue

### DIFF
--- a/lib/alignments.py
+++ b/lib/alignments.py
@@ -81,7 +81,7 @@ class Alignments():
         if extension[1:] == self.serializer.file_extension:
             logger.debug("Valid Alignments filename provided: '%s'", filename)
         else:
-            filename = "{}.{}".format(filename, self.serializer.file_extension)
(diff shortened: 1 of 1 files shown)


torzdf:
I don't believe there is an issue here any more.....
https://github.com/deepfakes/faceswap/blob/11ab910c5e830dacc4030ac3b70e4bb49a5879b0/lib/serializer.py#L258

https://github.com/deepfakes/faceswap/blob/11ab910c5e830dacc4030ac3b70e4bb49a5879b0/lib/alignments.py#L81

Test:
```py
>>> import os
>>> filename = "/path/to/alignments.fsa"
>>> extension = os.path.splitext(filename)[1]
>>> serializer_extension = "fsa"
>>> extension[1:] == serializer_extension
True
```

So if an alignments file is passed in with a .fsa extension it will never reach your code.

kilroythethird:
This was in fact still a problem when providing .json files, for example when using a config created pre FSA changes.
I merged  (a changed) version of your PR in #929

PR #928, opened 2019-11-03
Landmarks stored and used as floating point numbers

kvrooman:
Aligners currently will produce sub-pixel locations for the facial landmarks.

These landmarks are then cast to integers before being used in the point registration algorithm (Umeyama) and stored in the alignment folder.  This casting to integers makes the sub-pixel localization pointless. 

This change always keeps landmarks as floating point numbers. Result will be improved affine warping matrix, more stability in converts and training, and slightly better matching with Warp-to-Landmarks.

Note: Landmarks were originally cast to integers as cv2 required integer inputs for several functions used in creating mask and drawing landmarks on preview images.  These function's inputs have been converted to integers immediately prior to the function where appropriate.

While working through the affected files, several pylint messages were also removed


torzdf:
Sorry, conflict resolution required :(

PR #937, opened 2019-11-19
Added masking output support to tools->mask

kilroythethird:
Currently the mask tool outputs a combination of face, mask and masked face.
This PR add options to output only the mask or the masked face too.
Additionally it adds an option to output the whole frame.

This allows using the extraction and masking part of faceswap in an after effects workflow.
I am aware that this is not the goal of faceswap but since it became a pretty nice face extraction / clean and masking tool it would be nice to have.
This will become even more interesting as soon as better obstruction mask models are implemented.

### DIFF
--- a/tools/mask.py
+++ b/tools/mask.py
@@ -38,6 +38,9 @@ class Mask():
         self._input_is_faces = arguments.input_type == "faces"
         self._mask_type = arguments.masker
         self._output_opts = dict(blur_kernel=arguments.blur_kernel, threshold=arguments.threshold)
+        self._output_type = arguments.output_type
+        self._output_full_frame = arguments.full_frame
+        self._output_suffix = self._get_output_suffix()
 
         self._face_count = 0
         self._skip_count = 0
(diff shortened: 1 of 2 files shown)


tools/mask.py
```diff
@@ -38,6 +38,8 @@ def __init__(self, arguments):
         self._input_is_faces = arguments.input_type == "faces"
         self._mask_type = arguments.masker
         self._output_opts = dict(blur_kernel=arguments.blur_kernel, threshold=arguments.threshold)
+        self._output_type = arguments.output_type
+        self._output_full_frame = arguments.full_frame
```
torzdf:
```py
        self._output_suffix = self._get_output_suffix()
```

torzdf: Then create function:
```py
    def _get_output_suffix(self):
        """ The filename suffix, based on selected output options

        Returns
        -------
        str:
            The suffix to be appended to the output filename
        """
        sfx = "{}_mask_preview_".format(self._mask_type)
        sfx += "face_" if not self._output_full_frame or self._input_is_faces else "frame_"
        sfx += "{}.png".format(self._output_type)
        return sfx
```

torzdf: then in function `_save`

change:
```py
        filename = os.path.join(self._saver.location,
                                "{}_{}_{}_mask_preview.png".format(os.path.splitext(frame)[0],
                                                                   idx,
                                                                   self._mask_type))
```
to:
```py
        filename = os.path.join(self._saver.location, "{}_{}_{}".format(os.path.splitext(frame)[0],
                                                                        idx,
                                                                        self._output_suffix))
```

torzdf:
eg:
image

tools/mask.py
torzdf: eg:
image

kilroythethird: Done. Thanks for the suggestion.
Definitely makes sense.

PR #944, opened 2019-12-02
add sys.exit

prometeobot:
In python 3.6.9, I get an error: `NameError: name 'exit' is not defined`. This PR fixes this

### DIFF
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -5,6 +5,7 @@ import json
 import logging
 import os
 import sys
+from sys import exit
 import urllib
 import warnings
 import zipfile


torzdf:
This shouldn't be a bug....

`exit()` isn't an alias for `sys,exit()`. Whilst it effectively does the same thing, it gets imported from the `site` module which should be auto imported for all Python projects. 
https://docs.python.org/3/library/site.html
ref: https://help.semmle.com/wiki/pages/viewpage.action?pageId=29394142

torzdf:
Ok. A better approach would be to replace all of the `exit()` commands with either `sys.exit()` or a `raise SystemExit`

However, this is still not without its issues, because afaik sys.exit() exits the thread rather than the application which is problematic for multithreaded operations.

Either way, exits have been in the code since this repo started, and this is the first time this issue has ever been raised.

PR #949, opened 2019-12-07
Clarification of Phases in Extract

kvrooman:
When multi-processing, the status bar description lists the current phase as only Detect. This is misleading and there likely should be a reference to the multiple simultaneous phases being run.

image


torzdf:
Hmmm. Not too keen on this. I do take the point though. Probably best to say "Extraction" rather than spamming all the individual phase names though. The "Phase 1 of 1" should be enough to be clear it is only running through once.

kvrooman:
Adjusted status message as suggested to 'Extraction' when multi-processing

PR #952, opened 2019-12-09
Speedup mask loading before training start

kvrooman:
Roughly 10x speedup of loading masks into the training dictionary prior to training start.

For large data-sets, this setup time can be large. For a test set of 85k images, the mask loading time was 5 minutes, 40 seconds long on my computer . This PR reduces the mask loading time to 40 secs.

### DIFF
--- a/plugins/train/trainer/_base.py
+++ b/plugins/train/trainer/_base.py
@@ -1042,16 +1042,16 @@ class TrainingAlignments():
     training_opts: dict
         The dictionary of model training options (see module doc-string for information about
         contents)
-    image_list: dict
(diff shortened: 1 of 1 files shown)


torzdf:
Haha. I've also coded a fix for this.

I'll run some tests. Quickest will get implemented ;)

torzdf:
Images: 68547

Old: 132.78
Mine: 0.22
Yours: 29.63

Thanks for this PR, it is appreciated, but we'll go with the quicker solution

kvrooman:
Shakes fist and grumbles . Then says thank you for better speedup.

PR #960, opened 2020-01-09
Fix Bug if input size is different than 128

rragundez:
At the moment if the input size is different than 128 the model structure building fails on this line: `var_x = add([var_x, tmp_x])`https://github.com/deepfakes/faceswap/blob/master/plugins/train/model/villain.py#L41

This is because the number of filters in `var_x` and `tmp_x` are different. On line 37 and 38, `tmp_x` is created using  # of filters `in_conv_filters` but  `var_x` using 128, then when we try to add them, there is dimension mismatch on the 3rd dimension (# of filters).

### DIFF
--- a/plugins/train/model/villain.py
+++ b/plugins/train/model/villain.py
@@ -38,7 +38,7 @@ class Model(OriginalModel):
         tmp_x = var_x
         res_cycles = 8 if self.config.get("lowmem", False) else 16
         for _ in range(res_cycles):
-            nn_x = self.blocks.res_block(var_x, 128, **kwargs)
+            nn_x = self.blocks.res_block(var_x, in_conv_filters, **kwargs)
             var_x = nn_x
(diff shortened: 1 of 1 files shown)


bryanlyon:
While this isn't exactly a "bug" I don't see any reason to not commit this PR.  (The reason it's not really a bug is that Villain is meant to only allow a 128x input.)

Just as a curiosity, what do you see as the need for changing the input size here?

rragundez:
We are using it in a project , where we tried for the default input size but still doesn't render the desired quality, so we are trying a lot of things, one of them is increasing the input size.
So we modified the code a bit with the input size, but then it breaks in this line. In general is not a good idea that input size is hardcoded in my opinion but if it is, I think is best that it is done in one place.

torzdf:
I have no issue with this. Merging.

PR #970, opened 2020-02-11
Matplotlib fails, revert to 3.0.3?

JustinGuese:
Hey, 
Matplotlib fails in docker, ubuntu 16, 18 ,19.
Is it okay to revert to 3.0.3 which works with pip install? In my setup it seems to work.


torzdf:
What's the failure?

torzdf:
Should be fixed by #976

PR #972, opened 2020-02-15
Fix no-augment-color deprecation warning

xirvian:
Fix deprecation warning when using no_augment_color option


torzdf:
Thanks for the catch

PR #976, opened 2020-02-25
Update Tensorflow Docker image

jkramarz:
It is not possible to build project against TF 1.12 container, as it is based on old version of Python.
Also, solves #970


torzdf:
Thanks for this.

None of the current Devs use Docker, so I am going to wait for some feedback from other Docker users (@JustinGuese ?) and push if confirmed.

eduard-sukharev:
This solution actually works.
When attempted to run faceswap from docker I've faced same two issues: with required matplotlib version (#970 ) and with python version being outdated in 1.12 (required 3.6, but only 3.5 provided in image).

PR #994, opened 2020-03-22
Fix small mistake in USAGE.cmd

millerf:
Fix small error in USAGE in the "convert" section:

```
 You can see the full list of arguments for training by hovering
```
```
You can see the full list of arguments for extracting by hovering 
```


bryanlyon:
This needed to be convert actually, not training.

PR #1005, opened 2020-04-02
Change temporal-spatio smoothing to spatial smoothing

pfakanator:
Increase accuracy using 1 less frame for smoothing.  Subsequent runs can be done more smoothing rather than 1 shot garnering results closer to temporal if desired.


pfakanator:
cleaned up in 1008

PR #1006, opened 2020-04-02
Staging : Spatial Smoothing

pfakanator:
Rename existing spatial smoothing to spatio-temporal to better reflect it's function.  Introduce spacial smoothing witch performs 2 frames of interpolation rather than the 4 that temporal performs.


torzdf:
Thanks for this....

I am actually trying to reduce the number of jobs listed in the alignments tool, so I think I would prefer an approach which adds an interpolation parameter (from 0 (no spatial) to a number you feel is sensible)

pfakanator:
Where in the gui do you see that?

PR #1026, opened 2020-05-24
Update setup.py

kindect:
Grammar fixed


torzdf:
Errrrr. No.

kindect:
WTF???

torzdf:
It is already grammatically correct. Your PR would make it nonsensical.

And you raised an identical PR about a month ago which was closed for the same reason.

Issue #1037, opened 2020-07-21
Dockerfile setup for _requirements_base.txt is screwed


typoworx-de:
My Docker complains about wrong usage for COPY in Dockerfile as it can only copy ressources from a folder and not single files.

I think there is something required like folder 'docker-build' and a build or make file doing this:
```bash
#!/bin/bash
cp ../_requirements_base.txt .
sudo docker build -t faceswap-gpu - < ../Dockerfile.gpu
```

typoworx-de:
Something still goes wrong for me always getting this Error trying to build:
> COPY failed: stat /var/lib/docker/tmp/docker-builder790036227/_requirements_base.txt: no such file or directory

typoworx-de:
This approach worked for me. Notice the '.' which seems to give docker the current workdir so that it knows where to catch the files given by COPY command. Did not test (yet) if COPY without build-directory also works.

> ```shell
> #!/bin/bash
> cp ../_requirements_base.txt .
> sudo docker build -t faceswap-gpu . -f ../Dockerfile.gpu
> ```

torzdf:
Ultimately none of the current devs use Docker, so it's kind of unsupported and is really still hanging around for legacy reasons and because (to date) it hasn't broken yet....

We very much welcome PRs to fix any issues with the docker version, otherwise our only real choice is to drop support altogether,

consciousgaze:
I am still having this issue. My build command is 
`docker build -t faceswap:latest -f Dockerfile.gpu .`

Even specifying the building folder does not work. Does anyone has the similar issue?

torzdf:
I have just merged #1041 which may or may not resolve this issue

consciousgaze:
Yes. Changing the docker ignore file works. I really hope the docker function can be formally supported. That simplifies everything.

torzdf:
As long as it still works, it won't be going anywhere.

Glad this is resolved.

PR #1041, opened 2020-08-02
Update .dockerignore

FMS-Cat:
I think `_requirements_base.txt` should not be included in `.Dockerignore`.
I could not build the docker image without this change.

I'm not that good at using Docker, so ignore this if it does not make sense.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,2 +1,3 @@
 *
 !requirements*
+!_requirements*


torzdf:
Thanks for this, I assume this resolves #1037 ?

None of the current Devs use Docker, so I will merge this.

PR #1086, opened 2020-11-12
Update tensorflow to 2.4.0rc4

dmiszkiewicz:
I know this is only a draft, many things don't work like Analysis and learning graph. But I think this is a good point for starting supporting RTX 3080.  I think you should create a new branch to allow people to work on that change. Best! PS. I will try to work on this but I'm not a Python person.

### DIFF
--- a/lib/cli/launcher.py
+++ b/lib/cli/launcher.py
@@ -53,10 +53,10 @@ class ScriptExecutor():  # pylint:disable=too-few-public-methods
         Raises
         ------
         FaceswapError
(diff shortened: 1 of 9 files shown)


torzdf:
Thanks for this. I am currently working on implementing some other work, but will review when I can....

I won't be implementing to main repo until 2.4 is released (due to potential API changes), and whilst we will add support for tf 2.4, we won't add it to requirements until Conda update their packages.

Something to bear in mind, at least whilst we transition, we will need support for 2.2 - 2.4, which will require some branching logic. This is due to Conda generally being behind Pip.

PR #1096, opened 2020-12-08
Add Tensorflow 2.4 support

deepfakes:
This is the Dev branch for supporting Tensorflow 2.4 and supporting RTX 30xx cards.

It is currently WIP with many things not working. Use at your own risk.

See #1086 for original PR/Issues


dmiszkiewicz:
On Ampere, mixed_precision is turned on automatically, even if you set it to false in faceswap app.  I will collect all issues I have faced on Ampere and paste them here.

deepfakes:
> 
> 
> On Ampere, mixed_precision is turned on automatically, even if you set it to false in faceswap app. I think seting it to false should disable it on Ampere also. I will collect all issues I have faced on Ampere and paste them here.

Thanks. I do not currently have an Ampere GPU so will just be working through issues on TF2.4rc + Cuda 11.1 on an RTX 2080Ti

torzdf:
Ok, the main issue I see with this PR is that even if it is pinned to tf2.4_rc4, then it still will not work fully on RTX 3xxx cards, as ALL Tensorflow pip 2.4 packages are compiled with Cuda 11.0 and the 3xxx cards require Cuda 11.1.

Any work that Faceswap do to support a currently unreleased TF version is for nothing as end users will still need to compile their own Tensorflow version. And if they need to do that, they might as well compile a supported version.

PR #1107, opened 2020-12-29
Update alignments.py

CaptainStabs:
Added better handling of an already existing alignment file being used in the -al argument
Not really sure why it marked the entire file as changed, changes are on line 4, 695-704


CaptainStabs:
Noticed a few errors in my forked version.

Issue #1111, opened 2021-01-04
dockerfile.gpu python is 3.6


amitsh1:
*Note: For general usage questions and help, please use either our [FaceSwap Forum](https://faceswap.dev/forum) 
or [FaceSwap Discord server](https://discord.gg/FC54sYg). General usage questions are liable to be closed without
response.*

**Crash reports MUST be included when reporting bugs.**

**Describe the bug**
using dockerfile.gpu results in errors

**To Reproduce**
1. build using Dockerfile.gpu
2. run docker image
3. run: python faceswap.py -h

root@73fe496dfbc7:/app# python faceswap.py -h
Setting Faceswap backend to NVIDIA
Traceback (most recent call last):
  File "faceswap.py", line 11, in <module>
    raise Exception("This program requires at least python3.7")
Exception: This program requires at least python3.7

**Expected behavior**
it should run

**Screenshots**

**Desktop (please complete the following information):**
 - OS: ubuntu 16.04

 - 
**Additional context**

**Crash Report**
The crash report generated in the root of your Faceswap folder

PR #1112, opened 2021-01-04
python3.8 with cuda 10.1 cudnn 7 and tensorflow2.3

amitsh1:
#1111 

completely different from original version. 
hope this helps. please ignore if it doesn't


torzdf:
Many thanks for this. None of the current Devs use Docker, but this looks good on the face of it.

I'll leave this for a couple of days in case there is any feedback from other Docker users.

PR #1125, opened 2021-02-17
importlib fix for plugins/convert

andenixa:
fixes the infamous importlib error under Windows


torzdf:
Thanks for the heads up on this. I have pushed a fix which should stop this occurring in future (ref: https://github.com/deepfakes/faceswap/commit/48ca4d1b0e52ad940067896a5955553fbd785db3), so closing

PR #1127, opened 2021-02-19
Faceswap.spanish

tokafondo:
Ok. Those are my translations into spanish of the already supplied .pot files. I've tried them to be useful for spanish users, while maintaining most of the original names in english for the options, to keep consistency between different languages.


deepfakes:
Thanks for this. I will review over the weekend.

Yeah, unfortunately "option names" have to remain the same across languages, otherwise it becomes hard to maintain. But language specific help text/group headers should help some

torzdf (changes requested):
Good job on this. Seems fine from my initial checks. A couple of changes please:

`tools.sort.cli.pot` does not appear to have been translated. Please could you add this?

I have also updated the English for `cli.args.pot` for the `Mask` option to include the `Predicted` mask, so you will need to update the `msgid` in your es `cli.args.po` file to the following. You will have to do this in the source `.po` file directly, as you cannot edit source text in PoEdit:
```
"R|Masker to use. NB: The mask you require must exist within the alignments file. You can add additional masks with the Mask Tool.\n"
"L|none: Don't use a mask.\n"
"L|components: Mask designed to provide facial segmentation based on the positioning of landmark locations. A convex hull is constructed around the exterior of the landmarks to create a mask.\n"
"L|extended: Mask designed to provide facial segmentation based on the positioning of landmark locations. A convex hull is constructed around the exterior of the landmarks and the mask is extended upwards onto the forehead.\n"
"L|vgg-clear: Mask designed to provide smart segmentation of mostly frontal faces clear of obstructions. Profile faces and obstructions may result in sub-par performance.\n"
"L|vgg-obstructed: Mask designed to provide smart segmentation of mostly frontal faces. The mask model has been specifically trained to recognize some facial obstructions (hands and eyeglasses). Profile faces may result in sub-par performance.\n"
"L|unet-dfl: Mask designed to provide smart segmentation of mostly frontal faces. The mask model has been trained by community members and will need testing for further description. Profile faces may result in sub-par performance.\n"
"L|predicted: If the 'Learn Mask' option was enabled during training, this will use the mask that was created by the trained model."
```
Once you've added the English text, it is best to open the edited `.po` file in PoEdit and do the translation there and then save back out, as it will update the `.mo` file

torzdf:
FYI. You can ignore the failing check. It has nothing to do with this PR

PR #1132, opened 2021-03-16
apt -> apt-get in Dockerfile

DenDen047:
apt should be used on GUI systems. On CLI, apt-get is more suitable.


torzdf:
If you can cite the reason for this, it would be appreciated, as imho apt is fine, and is.

DenDen047:
apt has no issue. I just suggested for consistency for commands on the Dockerfile.

DenDen047:
And the below warning is showing:
`WARNING: apt does not have a stable CLI interface. Use with caution in scripts.`

PR #1133, opened 2021-03-16
Adding some defensive code to better identify which image is causing …

mark-gargan:
Adding some defensive code to better identify which image is causing an issue when permission error is encountered.

A previous functional training set somehow had two images that became corrupted, there was nothing in logs or stdout to indicate which image(s) were causing an issue.


torzdf (approved):
Useful add. Many thanks

PR #1134, opened 2021-03-16
Spanish translations for GUI tooltips.

tokafondo:
Translations for strings related to the GUI tooltips.


torzdf:
Thanks for this.

Please could you also update the following files...
- `locales/es/LC_MESSAGES/tools.manual.po`
- `locales/es/LC_MESSAGES/tools.preview.po`

You can just open the `.po` files directly into PoEdit (I have added the English source files). Make the changes, then save the `.po` file back, and the `.mo` file should then auto-update

torzdf (changes requested):
There are a couple of other entries missing. Please see above comment

torzdf (approved):

PR #1156, opened 2021-06-05
Grammar

82600417:
There were a few grammatical errors, so I fixed them.


bryanlyon:
In general we do not accept grammar/spelling PRs.  We appreciate the notification but many of these fixes are actually invalid.

82600417:
@bryanlyon 

>In general we do not accept grammar/spelling PRs.

Fair enough.

>many of these fixes are actually invalid.

Actually, none of them are.

PR #1166, opened 2021-06-23
Update detected_face.py

albernsrya:
Using the literal syntax can give minor performance bumps compared to using function calls to create dict, list and tuple. 
This is because here, the name dict must be looked up in the global scope in case it has been rebound. Same goes for the other two types list() and tuple().


bryanlyon:
These fixes are not acceptable.  They fail Pep8, provide no real improvement, and actually make the code harder to read.  dict() is completely acceptable as nothing we're doing will benefit from a 0.2 microsecond speed increase yet it is significantly easier to read.

PR #1167, opened 2021-06-23
Update layers_test.py

albernsrya:
Do not use a mutable like list or dictionary as a default value to an argument. Python’s default arguments are evaluated once when the function is defined. Using a mutable default argument and mutating it will mutate that object for all future calls to the function as well.

### DIFF
--- a/tests/lib/model/layers_test.py
+++ b/tests/lib/model/layers_test.py
@@ -18,13 +18,14 @@ from tests.utils import has_arg
 CONV_SHAPE = (3, 3, 256, 2048)
 CONV_ID = get_backend().upper()
 
-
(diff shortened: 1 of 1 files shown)


bryanlyon:
Mutables as a default arg can cause problems, however, this is a test function which those kwargs are defined in every case and never mutated in the function (the only place that could be mutated).

We do not generally accept PRs on minor changes like this, especially when they fail basic pep8 format issues.  If you wish to participate in FaceSwap development there are a lot of areas that some substantial improvements could be made and we'd be more than happy to mentor or advise on what you could do to get a successful PR.  But minor changes like your last two PRs will not be accepted (unless they're fixing an actual live bug).

PR #1169, opened 2021-07-11
Add a new sorting/grouping feature based on black pixels in images

procule:
Context:
- Faces are quite often moving and can get near borders of the image.
- Because of that, the face image has the out of bound part replaced by black pixels.
- Some are still good for training but others not. Also, one might want to keep an alignment file with the near borders faces but excluded (some of) them for training.
- Having to manually move/delete those images in a dataset of thousands of them is painful and prone to mistakes.
- Thus, having a way to sort and group those images would be helpful. At least, it is for me :-)

Added features:
- A new option in Tools/Sort for both Sort and Group By called Black-Pixels.
- Sort feature: calculates the percentage of black pixels in each images and sort them from 0 to 100%.
- Group By feature: Uses the Bins slider number selected and creates 100/Bins folders with the images percentage of black pixels. ie. a Bins of 5 will create 20 folders with the first one containing images with 0-5% of black pixels, the second one with 6-10%, etc. A static method as been added for the rounding error.

### DIFF
--- a/tools/sort/cli.py
+++ b/tools/sort/cli.py
@@ -78,7 +78,8 @@ class SortArgs(FaceSwapArgs):
                    "faces further from the camera and from lower resolution sources will be "
                    "sorted last."
                    "\nL|'black-pixels': Sort images by their number of black pixels. Useful when "
-                   "faces are near borders and a large part of the image is black.\nDefault: face")))
+                   "faces are near borders and a large part of the image is black."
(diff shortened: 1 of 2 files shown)


torzdf:
Thanks for the PR. Appreciated. I have a lot on at the moment but will review as soon as I can.

torzdf (approved):

PR #1215, opened 2022-05-04
Update faceswap.py

Cybergenik:
Use standard Python semantics for entering program


bryanlyon:
Sys exit is not standard and it's completely unnecessary.

PR #1238, opened 2022-06-16
Some Apple Silicion setup fixes

geewiz94:
Ran some tests and fixed these issues:
- Tries and fails to install pynvx, moved to Nvidia requirement
- tensorflow-deps was missing
- Tries and fails to install tensorflow-deps / libblas using pip, now using Conda
- Wrong tensorflow-metal version
- The command for creating the Conda env wasn't working for me for some reason

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -191,7 +191,7 @@ $ source ~/miniforge3/bin/activate
 ## Setup
 ### Create and Activate the Environment
 ```sh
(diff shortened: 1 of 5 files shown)


torzdf:
Thanks, testing and updates are appreciated, I had to implement this blind, so I'm fairly happy that most of the errors seem to be fairly trivial.

I'm hoping by standardising the setup path it will make it easier to maintain this going forward.

PR #1271, opened 2022-09-12
add max scale option

aopsr:
This PR adds the option to set a max scale parameter in conversion. Max scale is defined at the maximum ratio of the longest edge of any face's original_roi in the source frame to the model output size.

Example 1: Let max scale be 2 and model have a 256px output size. Image contains a face that is 768x400 px. Max ratio = 768/256 = 3. Therefore, the image will be scaled by 2/3 before pasting the predicted face.

Example 2: Let max scale be 1.5 and model have a 512px output size. Image contains a face that is 400x300 px. Max ratio = 400/512 < 1. Therefore, the image will not be scaled.

Example 3: Let max scale be 0 (default). The image will not be scaled.

Setting max_scale to a number less than 2 produces results that are far more realistic visually when the model resolution is small relative to dest.

This feature checks if the input is a folder of images, not a video, before applying the appropriate transformations. Video frames will need (trivial) further work to be compatible with variable resolution frames. However, it does not make sense to generate a video from different resolution frames. Therefore, this feature was not included.

### DIFF
--- a/lib/align/aligned_face.py
+++ b/lib/align/aligned_face.py
@@ -421,6 +421,21 @@ class AlignedFace():
         return self._cache.original_roi[0]
 
     @property
+    def original_roi_2(self) -> np.ndarray:
+        """ :class:`numpy.ndarray`: The location of the extracted face box within the original
+        frame. """
+        with self._cache.lock("original_roi"):
+            if self._cache.original_roi is None:
+                roi = np.array([[0, 0],
+                                [0, self._size - 1],
+                                [self._size - 1, self._size - 1],
+                                [self._size - 1, 0]])
+                roi = np.rint(self.transform_points(roi, invert=True)).astype("int32")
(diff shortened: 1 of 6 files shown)


torzdf:
Sorry I have sat on this for so long. I am unsure of the value add here.

If I am understanding correctly, the final output image is resized based on the size of the output of the model. If this is the case, then I have the following reservations:

- Faceswap is mostly used for videos, which would not be covered by this PR, so would be adding an option for a very limited use case
- The resizing of images can be done fairly easily on either the original images fed to Faceswap or on the final output images generated from Faceswap. It could be argued (in the latter case) that it is preferable to see the swapped output before making a judgement call on whether the image should be resized.
- The only potential benefit is if you are swapping onto lots of images of lots of different sizes, as otherwise 'output scaling' can achieve much the same effect

PR #1322, opened 2023-06-07
create even more RU translations

andentze:


andentze:
I've discovered something rather too weird...

Linux encodings will ALWAYS be UTF-8(and I bet you can change it). But Windows is special. It can come up with whatever code page it wants. But UTF-8 _probably_ won't be one of them. But we can fix that.

The reason not so many people know about it is because Microsoft is hiding such unknown or "beta" settings deep in the legacy menus(from Windows XP era, or maybe even older)

"So, I'm getting UnicodeDecodeError's when using your locale. What's wrong?"

Because Faceswap fails to generate config files. And the reason behind that is the wrong encoding.
Sure, the recent update forced UTF-8 encoding, but Windows doesn't seem to comply... How do we fix this?

- Run "Run" by pressing "Win + R" and then type "intl.cpl" in there, which will throw us in the legacy Region settings.

- Then we need to find "Administrative" and press "Change system locale" there.

- On the next window, you'll see a checkmark saying "Beta: Use Unicode ITF-8 for worldwide language support" and you'll need to click that and then apply the settings.

- Reboot the system afterwards.

Or...
We can force the Command Prompt's code page to change.

- Open the Command Prompt

- type "chcp 65001"

- Run Faceswap.

Why is the encoding **not** UTF-8 in Windows? No one really knows... But at least we found the fix.

(_god i've spent so much time working on a fix..._)

torzdf:
So, is this something a user needs to do for locales to work in the .ini files? If so, then I think we'll need to scrap locales for ini files. It's annoying, but it becomes a support headache.

andentze:
No need. It's possible to add it to the batch file that starts Faceswap, so it'll apply only to the command prompt. I haven't really tested that though, but it may work. That may mean changing the way the installer is creating the batch file a tiny bit...

torzdf:
Merging as codepage issue is not recreatable and config is now explicitly saved and loaded in utf-8

PR #1379, opened 2024-03-22
fixed regular expression to add support of unicode

igeni:
fixed regular expression to add support of unicode chars and simplify expression


torzdf:
Fixed in commit https://github.com/deepfakes/faceswap/commit/0f947791f57db7e31d372ce2d098e5d7737cf1fe

Thanks for the PR.

PR #1402, opened 2024-09-20
Update simple_tests.py

JVALVARO:
Fix: Consider using the requests library, which has better error handling and timeout capabilities


torzdf:
This is now obsolete

PR #1483, opened 2025-02-10
fix simple_tests retry

balckduck:
fix simple_tests retry


torzdf:
This is now obsolete. Test media is stored within the project

PR #1542, opened 2026-04-29
bugfix: loss-function correctness and GUI/event-reader robustness

kolzx155:
## Summary

Two unrelated clusters of bugs uncovered while reviewing the recent torch-loss-functions-in-Keras work and trying to use the GUI training preview/analysis tabs.

### Loss-function correctness (`lib/model/losses/`)

- **`GradientLoss._diff_xy`** — variable-name collision. The two halves of the mixed second-order term (` # x_out1 ` and ` # X_out2 `) both wrote into `xy_left` / `xy_mid` / `xy_right`, so by the time `xy_out1` and `xy_out2` were assembled they pointed at the *same* tensors and the final `(xy_out1 - xy_out2) * 0.25` was identically zero. The cross-derivative term in `GradientLoss` has therefore been silently dead for as long as this code has existed. Renamed to `xy1_*` / `xy2_*`.
- **`MSSIMLoss._fspecial_gauss`** — operator-precedence bug. `coords -= size - 1 / 2.` parses as `coords -= size - 0.5`, so for `size=11` the kernel coordinates land at `[-10.5 … -0.5]` instead of the centred `[-5 … 5]`. The companion `DSSIMObjective._get_kernel` already had the correct `(self._filter_size - 1) / 2.`. Added the missing parens.
- **`FocalFrequencyLoss._get_patches`** — was still slicing with NHWC indices (`inputs[:, row_from:row_to, col_from:col_to, :]`) after the `permute(0, 3, 1, 2)` to NCHW. Worked by accident for the default `patch_factor=1` on square images; for `patch_factor > 1` the FFT was being run over the wrong axes. Now slices the H/W axes directly.
- **`LDRFLIPLoss._color_pipeline`** — the hunt-adjusted RGB reference colours and the resulting `c_max` are constants, but were rebuilt on every forward pass. Compute `c_max` once in `__init__` and register it as a buffer so it follows the module across `.to(device)`.
- **`GMSDLoss._map_scharr_edges`** — `torch.atan(gx / gy)` divides by zero wherever `gy == 0`. Switched to `torch.atan2(gx, gy)`. The branch is only reachable when `magnitude=False` (not currently exercised by `forward`), so this is a latent fix.

### GUI / event-reader robustness

- **`_Taskbar.destroy_widgets` (`lib/training/preview_tk.py`)** — `widget.winfo_ismapped()` raised `TclError: bad window path name` when the parent (e.g. the Notebook tab) had already torn the widget down before `destroy_widgets` ran. Wrapped in a `winfo_exists()` guard and a `try/except tk.TclError`. Also `self._gui_mapped.clear()` after the loop — the previous `del widget` only dropped the loop variable, leaving stale references in the list, so a second invocation would walk over already-destroyed widgets.
- **`RecordIterator.__next__` (`lib/training/tensorboard.py`)** — when the GUI tailed an event file the trainer was still writing, it could catch a partial record: 8 bytes of length header flushed but the body not yet there. The garbage length (often `~10**18`) was then handed to `file.read()`, raising `OverflowError: cannot fit 'int' into an index-sized integer` or `MemoryError`, which killed the analysis thread. Now records the start offset, reads the full record (header + len-CRC + data + data-CRC) atomically, and on any short read or implausible length (cap of 1 GiB) seeks back to the start so the next poll resumes once the writer has flushed.

## Behaviour-change notes for reviewers

- Models trained with `GradientLoss` were effectively running with the second-order term reduced to `_diff_xx + _diff_yy` (the mixed term contributed exactly zero). After this fix the loss surface changes; existing checkpoints still load fine but the gradients during continued training will differ.
- `MSSIMLoss` was running on a badly off-centre Gaussian. After this fix the loss values change significantly — same caveat for in-flight training.
- `FocalFrequencyLoss` users who only ever ran with `patch_factor=1` will see no behavioural change on square images.

## Test plan

- [ ] Unit test for `GradientLoss._diff_xy`: verify it now returns non-zero values for non-trivial inputs and matches the analytical mixed-derivative on a smooth synthetic image.
- [ ] Unit test for `MSSIMLoss._fspecial_gauss`: kernel sums to ~1 and is symmetric around the centre.
- [ ] Unit test for `FocalFrequencyLoss` with `patch_factor=2,4` on a non-square-friendly batch shape.
- [ ] Confirm `LDRFLIPLoss` still produces the same `c_max` value before/after (regression check; not a behaviour change).
- [ ] Manual: run training with the GUI, trigger a preview-tab close while training is live (taskbar destroy path) and watch the analysis tab refresh while the writer is still flushing (event reader path) — both should no longer surface tracebacks in the console.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/model/losses/loss.py
+++ b/lib/model/losses/loss.py
@@ -86,7 +86,7 @@ class FocalFrequencyLoss(nn.Module):
                 row_to = (i + 1) * patch_rows
                 col_from = j * patch_cols
                 col_to = (j + 1) * patch_cols
-                patch_list.append(inputs[:, row_from: row_to, col_from:col_to, :])
+                patch_list.append(inputs[:, :, row_from:row_to, col_from:col_to])
 
         retval = torch.stack(patch_list, dim=1)
         return retval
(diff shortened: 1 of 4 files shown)


torzdf:
I'm closing this PR for 2 reasons.

1) Loss porting work is continuing so this PR has conflicts with the current development code
2) We do not accept AI authored code. You are welcome to use AI tools in order to identify issues, but the code itself must be authored (and tested) by a human.

I will review the fixes proposed in this PR and roll those into the current development code where appropriate.

Please do not be dissuaded from contributing in future, as identifying bugs is certainly helpful (as there are invariably many in the codebase). If you identify bugs, but are not prepared to author the fixes yourself, then feel free to raise an issue.

Thanks

PR #1557, opened 2026-08-01
fix(gui): propagate --logfile to processes launched from the GUI

### LINKED ISSUE #1377: -LF option doesn't get carried to the internal tools
Hi. I'm trying to use two faceswap instances in parallel. One of them is using GPU and I want to use the another one with only CPU for secondary tasks.

I use python faceswap.py gui -LF c:\temp\faceswap.log.2 -X0 to start the secondary instance.

When I try to run a sort task in this secondary instance, I get this:

```
Setting Faceswap backend to NVIDIA
File "C:\Users\Yo\faceswap\tools.py", line 48, in <module>
ARGUMENTS.func(ARGUMENTS)
File "C:\Users\Yo\faceswap\lib\cli\launcher.py", line 217, in execute_script
log_setup(arguments.loglevel, arguments.logfile, self._command, is_gui)
File "C:\Users\Yo\faceswap\lib\logger.py", line 375, in log_setup
f_handler = _file_handler(numeric_loglevel, log_file, log_format, command)
File "C:\Users\Yo\faceswap\lib\logger.py", line 419, in _file_handler
handler.doRollover()
File "C:\Users\Yo\miniconda3\envs\faceswap\lib\logging\handlers.py", line 179, in doRollover
self.rotate(self.baseFilename, dfn)
File "C:\Users\Yo\miniconda3\envs\faceswap\lib\logging\handlers.py", line 115, in rotate
os.rename(source, dest)
PermissionError: [WinError 32] El proceso no tiene acceso al archivo porque está siendo utilizado por otro proceso: 'C:\\Users\\Yo\\faceswap\\faceswap.log' -> 'C:\\Users\\Yo\\faceswap\\faceswap.log.1'
Process exited.
```

It seems that internally, it's still using the main log, because the -LF option doesn't get carried to the `sort` command.

If I run `python.exe faceswap\tools.py sort -i C:/workspace.blade/Ford/Ford1 -s none -g face -t -1.0 -b 5 -lf sort_log.json -L INFO  -LF c:\temp\faceswap.log.3` directly in the console, it works in parallel with the main instance.

mvanhorn:
Running the GUI with `-F/--logfile` only redirected the GUI's own log; the extract/train/convert processes it launched still wrote to their default logfile, so the flag looked like it did nothing. Closes #1377.

## Updates

- `scripts/gui.py`: `FaceswapGui` and `Gui` now take `arguments.logfile` and pass it through to `ProcessWrapper`.
- `lib/gui/wrapper.py`: `ProcessWrapper` accepts the logfile and appends `-F <logfile>` to each generated child command, but only when the user has not already set `-F/--logfile` for that command in the GUI, so an explicit per-command setting still wins.
- `lib/logger.py`: when an explicit `log_file` is given and the command is `gui`, the GUI's own handler writes to `<stem>_gui<ext>`. The parent and its children would otherwise share one file, which is the Windows sub-process sharing problem the default path already avoids with a separate GUI log.

Net effect: `python faceswap.py gui -F /path/run.log` puts the GUI in `/path/run_gui.log` and the jobs it launches in `/path/run.log`.

Tested by compiling the three touched modules and by reading the arg-generation path in `gen_cli_arguments`; I did not have a working GUI session here, so a quick sanity run on your side would be worth it. The `_gui` suffix is the one debatable bit: it keeps parity with the existing default behaviour, but if you would rather both go to the same file, that hunk drops cleanly on its own.

### DIFF
--- a/lib/gui/wrapper.py
+++ b/lib/gui/wrapper.py
@@ -28,10 +28,17 @@ logger = logging.getLogger(__name__)
 
 class ProcessWrapper():
     """ Builds command, launches and terminates the underlying
-        faceswap process. Updates GUI display depending on state """
+        faceswap process. Updates GUI display depending on state
 
-    def __init__(self) -> None:
+    Parameters
+    ----------
+    logfile : str | None
+        The GUI startup logfile to pass to child commands. ``None`` to use their default logfile
+    """
+
+    def __init__(self, logfile: str | None = None) -> None:
         logger.debug("Initializing %s", self.__class__.__name__)
+        self._logfile = logfile
         self._tk_vars = get_config().tk_vars
         self._set_callbacks()
         self._command: str | None = None
@@ -164,11 +171,15 @@ class ProcessWrapper():
         args.extend([pathexecscript, command])
 
         cli_opts = get_config().cli_opts
+        has_logfile = False
         for cliopt in cli_opts.gen_cli_arguments(command):
             args.extend(cliopt)
+            has_logfile |= cliopt[0] in ("-F", "--logfile")
             if command == "train" and not generate:
                 self._get_training_session_info(cliopt)
 
+        if self._logfile and not has_logfile:
+            args.extend(("-F", self._logfile))
         if not generate:
             args.append("-G")  # Indicate to Faceswap that we are running the GUI
         if generate:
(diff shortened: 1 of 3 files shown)


torzdf:
Yeah, we can't do this unfortunately. Originally we had a single logfile, but it sometimes causes locking issues on Windows so the GUI and the main app need separate logfiles.

Specific log files can still be specified from within the GUI for the running task.

